PR #23, opened 2025-11-14
Support custom headers and API URL in scan viewer.

rasmusfaber:
Reuse `requests.ts` from Inspect AI to support injecting a `HeaderProvider` and to override the location of the API.

Also allow for setting `resultsDir` and passing that to `/scans`.


dragonstyle (commented):
Overall seems good. Just to confirm did you bring request.ts over from inspect_ai unmodified? (I didn't review it carefully but will be ok if it is what we're already using).

PR #35, opened 2025-11-24
Stream dataframe results to client

rasmusfaber:
Currently calling `/scanner_df/{scan:path}?scanner={query_scanner}` involves reading all the parquet files into memory, selecting the one involving the scanner, converting it to arrow format, and then returning the serialized data.

For large scans, this can require a considerable amount of memory.

This PR changes this into a streaming implementation.


jjallaire:
cc @dragonstyle could you review this?

jjallaire:
Thank you!!! This will be a huge difference maker. We have another variation of this here which makes Arrow IPC a more formal part of the results database contract: https://github.com/meridianlabs-ai/inspect_scout/pull/38

PR #56, opened 2025-12-04
Fix  to ensure`parallel_tool_calls=False` when using structured answer `llm_scanner` and fix loss of `ScanJob`'s  `GenerateConfig`

epatey:


src/inspect_scout/_scan.py
```diff
@@ -260,16 +264,23 @@ async def scan_async(
...
-    scanjob._generate_config = (
-        model_config or scanjob._generate_config or GenerateConfig()
+    merged_config = (
+        (
+            scanjob._generate_config.merge(model_config)
+            if scanjob._generate_config
+            else model_config
+        )
+        if model_config
+        else (scanjob._generate_config or GenerateConfig())
```
epatey: @jjallaire, one thing I want to verify is that my new code no longer mutates `scanjob._generate_config`. I suspect that's correct, but I want to make sure there's not something subtle that I'm missing.

jjallaire: I am uncertain about the virtue of mutating that (AFK so can't dig into it). In any case, the reason to mutate is so it works with resume. If you can satisfy yourself that the same config gets re-applied on resume then we're good. Remember that resume has NO cli args so now that I'm verbalizing it out maybe it does need to mutate?

PR #79, opened 2025-12-12
Fetch input on demand

dragonstyle:
- Don't fetch inputs when loading the initial results dataframe
- fetch input when drilling into a single result

### DIFF
--- a/src/inspect_scout/_view/server.py
+++ b/src/inspect_scout/_view/server.py
@@ -255,16 +255,6 @@ def view_server_app(
         # validate
         await _validate_read(request, scan_path)
 
-        # Read the input for the scanner and uuid
-        result = await scan_results_arrow_async(str(scan_path))
-
-        # ensure we have the scanner data (404 if not)
-        if query_scanner not in result.scanners:
-            raise HTTPException(
-                status_code=HTTP_404_NOT_FOUND,
-                detail=f"Scanner '{query_scanner}' not found in scan results",
-            )
-
         # get the result
         result = await scan_results_arrow_async(str(scan_path))
 


src/inspect_scout/_view/server.py
```diff
@@ -271,13 +352,26 @@ def stream_as_arrow_ipc() -> Iterable[bytes]:
             with result.reader(
                 query_scanner, streaming_batch_size=streaming_batch_size
             ) as reader:
+                # Filter input from the datastream as it can be very
```
rasmusfaber: If I read the pyarrow implementation correctly, this still reads the input column from the Parquet files. It just doesn't write it to the output stream.

You can avoid reading it at all by providing the columns to include to `iter_batches()` like [this](https://github.com/meridianlabs-ai/inspect_scout/pull/83/files#diff-eeb5935a560fae5f099bcd2fa4abf0f3157fc7f5cb6a43f34a62bec436122ff3R253-R255).

dragonstyle: Thank you! I've updated to match your approach from #83 (using the recorder interface). Appreciate your review and time on this.

src/inspect_scout/_view/server.py
```diff
@@ -224,6 +225,86 @@ async def scans(
...
+            raise HTTPException(
+                status_code=HTTP_404_NOT_FOUND,
+                detail=f"Scanner '{query_scanner}' not found in scan results",
+            )
+
+        # Search for the row with matching uuid and extract input and inputType columns
+        input_value: str | None = None
+        input_type: str | None = None
+        with result.reader(query_scanner) as reader:
+            for batch in reader:
```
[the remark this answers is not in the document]
dragonstyle: Thank you! I've updated to match your approach from #83 (using the recorder interface). Appreciate your review and time on this.

PR #134, opened 2026-01-02
Add `TranscriptsView` ABC with eval log support; use in API instead of `TranscriptsReader`

epatey:
## Summary
- Factor out `TranscriptsView` ABC for read-only database access
- Rename `EvalLogTranscriptsDB` → `EvalLogTranscriptsView` and make it conform to `TranscriptsView`
- Switch POST `/transcripts/{dir}` and GET `/transcripts/{dir}/{id}` endpoints from `TranscriptsReader` to `TranscriptsView`

`TranscriptsReader` was not designed for API endpoint use cases. `TranscriptsView` directly supports conditions, limits, and order_by parameters.

### DIFF
--- a/src/inspect_scout/_transcript/database/database.py
+++ b/src/inspect_scout/_transcript/database/database.py
@@ -3,6 +3,7 @@ from types import TracebackType
 from typing import AsyncIterable, AsyncIterator, Iterable, Literal, Type
 
 import pyarrow as pa
+from typing_extensions import Self
 
 from inspect_scout._transcript.transcripts import Transcripts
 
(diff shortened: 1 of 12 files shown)


src/inspect_scout/_transcript/database/database.py
```diff
@@ -15,11 +16,11 @@
 from .source import TranscriptsSource
 
 
-class TranscriptsDB(abc.ABC):
-    """Database of transcripts."""
+class TranscriptsView(abc.ABC):
+    """Read-only view of transcripts database."""
 
     def __init__(self, location: str, where: list[Condition] | None = None) -> None:
```
jjallaire: Move `__init__()` to TranscriptsDB and then remove all the business about calling super with location from EvalLogTranscriptsDB

PR #228, opened 2026-01-24
Feature/skip scored

anthonyduong9:


tests/scan/test_skip_scored.py
```diff
@@ -0,0 +1,302 @@
...
+@solver
+def add_score_solver(
+    scorer_name: str, scored_sample_ids: set[str] | None = None
+) -> Solver:
+    async def solve(state: TaskState, _generate: Generate) -> TaskState:
+        if scored_sample_ids is None or state.sample_id in scored_sample_ids:
+            state.scores = {scorer_name: Score(value=1)}
+        return state
+
+    return solve  # type: ignore[return-value]
```
andrei-apollo: nit: Can we fix the argument name and remove `type: ignore`?

PR #291, opened 2026-02-18
fix: return 404 instead of 500 for stale scanner UUIDs

revmischa:
## Summary

- Wraps `get_field()` calls in the `scanner_input` endpoint with `try/except KeyError` to return a proper 404 instead of an unhandled 500
- Root cause: when a scan is re-run, UUIDs are regenerated via `shortuuid.uuid()`. Users with stale browser tabs still reference old UUIDs, causing `KeyError` in parquet lookup

## Changes

- `src/inspect_scout/_view/_api_v2_scans.py`: Added `try/except KeyError` around `get_field()` calls, returning HTTP 404 with a descriptive message
- `tests/view/test_api_v2_scanner_input.py`: Added tests for unknown UUID (404), unknown scanner (404), and valid UUID (200 with input + header)

## Test plan

- [x] `ruff check` passes
- [x] `ruff format` passes  
- [x] `mypy src` passes
- [ ] `pytest tests/view/test_api_v2_scanner_input.py` — blocked by `inspect_ai` dependency mismatch (`CompactionEvent` missing); affects all tests in repo, not specific to this change

### DIFF
--- a/src/inspect_scout/_view/_api_v2_scans.py
+++ b/src/inspect_scout/_view/_api_v2_scans.py
@@ -8,12 +8,15 @@ import sys
 import tempfile
 import threading
 import time
+import zipfile
 from typing import Any, Iterable
 
+import anyio
 import pyarrow.ipc as pa_ipc
 from duckdb import InvalidInputException
-from fastapi import APIRouter, HTTPException, Path, Response
+from fastapi import APIRouter, HTTPException, Path, Request, Response
 from fastapi.responses import StreamingResponse
+from inspect_ai._util.file import file
 from send2trash import send2trash
 from starlette.status import (
     HTTP_404_NOT_FOUND,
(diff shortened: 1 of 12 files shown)


revmischa:
Good question! `get_field` in `file.py` explicitly raises `KeyError` when the UUID filter returns 0 rows (line 301-302):

```python
if len(table) == 0:
    raise KeyError(f"{id_value!r} not found in {id_column}")
```

So it's not a dict key lookup — it's `get_field`'s own contract for "ID not found in the parquet data."

The stale scenario: user has a scan results page open with UUIDs from a previous run, re-runs the scan, data changes, and those UUIDs no longer exist. The `get_field` lookup returns 0 rows → `KeyError` → we catch it and return 404 instead of 500.

epatey:
Thanks @revmischa.

It may be the right call to map `KeyError`'s to `404`'s, but I want to make sure I understand the scenario before moving forward.

I'm trying to connect the stale UUID description to the code and struggling a bit. From what I know, `scan_id` is generated via `default_factory=uuid` on `ScanSpec` and never passed explicitly — so every scan gets a new `scan_id`, new directory, and new UUIDs. The old scan's data remains intact at its old URL. So I'm not seeing how a new run would invalidate UUIDs in a browser tab still pointing at the old scan.

I'm likely missing something — could you walk me through concrete repro steps? What browser state + action leads to the 500?

I also probably make sense to merge this branch up to resolve the conflicts before too long.

epatey:
The Sentry [issue](https://metr-sh.sentry.io/share/issue/35958c62ce6849c39c46edec9ca2ee06/) makes it clear that this is in the deprecated V1 API.

PR #319, opened 2026-03-03
fix: guard against empty eval logs producing invalid SQL

sjawhar:
## Summary

When `scan()` is called with an eval log that has zero samples, `samples_df_with_caching()` returns a zero-column DataFrame. Passing this to pandas `to_sql()` generates `CREATE TABLE transcripts ()` — invalid SQLite — causing an `OperationalError`.

## Changes

- **`EvalLogTranscriptsView`** (`eval_log.py`): Add `_is_empty` flag. In `connect()`, detect zero-column DataFrames before `to_sql()`, set the flag, and open a no-op `:memory:` connection for idempotency. Guard `select()`, `count()`, `distinct()`, and `transcript_ids()` to return empty results immediately when the flag is set.
- **Regression test** (`test_transcript.py`): Verify `EvalLogTranscriptsView(pd.DataFrame())` connects and all query methods return empty results without crashing.

Closes trajectory-labs-pbc/agent-c#2047

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Unreleased
+
+- Bugfix: Guard against empty eval logs producing invalid SQL
+
 ## 0.4.43 (05 July 2026)
 
 - Store transcript and scan-result event/input columns as compact JSON (`indent=None`), substantially reducing on-disk Parquet size and the bytes streamed to Scout View (e.g. ~700 MiB → ~200 MiB for a large events column).
--- a/src/inspect_scout/_transcript/eval_log.py
+++ b/src/inspect_scout/_transcript/eval_log.py
@@ -214,12 +214,16 @@ def _logs_df_from_snapshot(snapshot: ScanTranscripts) -> "pd.DataFrame":
         # determine unique logs, re-read, then filter on sample_id
         snapshot_logs = snapshot_df["log"].unique().tolist()
         df = _index_logs(snapshot_logs)
+        if df.columns.empty:
+            return df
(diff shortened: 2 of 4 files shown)


epatey:
@sjawhar, thanks for the fix. I assume the `timeline.py` changes aren't necessary for the fix. If so, can you merge up and drop the changes to that file please.

epatey (approved):

sjawhar:
Thanks @epatey. I dropped the `timeline.py` changes and merged up, so this PR is now just the empty-eval-log fix. (Confirmed the `timeline.py` shim wasn't needed: `eval_log.py` doesn't import that module, and both import cleanly against the pinned `inspect_ai`.)

Merge is currently blocked by `check-schema-and-types`, which is failing repo-wide rather than because of this PR. CI installs `inspect_ai` from git main, which recently added `CompactionEvent.role`, so the committed `openapi.json` and `generated.ts` are stale against it. This PR doesn't touch the schema, so it fails the same way main and the other open PRs do.

#507 is the resync for this, and it needs one small fix of its own. Its failing step is that the ts-mono `generated.ts` is out of sync with its own `openapi.json`: the `openapi.json` has the new `role` field but `generated.ts` doesn't. Running `pnpm --filter scout types:generate` against #507's `openapi.json` re-adds `role`, so regenerating and committing `generated.ts` in the submodule should clear that check. Once #507 lands, this PR should go green after a rebase.

Claude (on Sami's behalf)

epatey:
@sjawhar, yeah. The python repos were out ahead of `ts-mono`. I'm working on getting that dealt with. I'll get this merged asap

PR #367, opened 2026-03-26
perf: lazy-load scan_events and use PyArrow range requests for S3 parquet reads

rasmusfaber:
## Summary

Large scan files in S3 can be very slow to load. #79 improved this greatly by lazy-loading the `input` field, but the `scan_events` field can also be huge.

This PR generalizes the lazy-loading of the `input` field, and adds a `/fields` endpoint for lazy loading multiple fields. The `/input` endpoint gets deprecated.

It also replaces the full-file-download with using PyArrow S3FileSystem (and GcsFileSystem), which supports much more efficient retrieval using range requests.

Companion PR: https://github.com/meridianlabs-ai/ts-mono/pull/13

## Changes

- `file.py`: Added `_resolve_parquet_source()`, `_point_lookup()`. Updated `reader()` to use `pre_buffer=True` for remote files. Updated `get_fields()` to do inline row-group scanning with a single `ParquetFile` open.
- `_api_v2_scans.py`: New `/fields` endpoint with allowlist validation. Deprecated old `/input` endpoint. Excluded `scan_events` from bulk streaming.
- `test_v2_api.py`: Tests for the new `/fields` endpoint (happy path, validation, error cases).
- Frontend submodule + dist updated with `/fields` endpoint integration.

## Test plan

- [X] Manual tested on our staging environment. Large scans that previously loaded in 2-3 minutes now loads in ~10 seconds.

### DIFF
--- a/scripts/export_openapi_schema.py
+++ b/scripts/export_openapi_schema.py
@@ -16,7 +16,7 @@ import subprocess
 import sys
 from pathlib import Path
 
-from pydantic import RootModel
+from pydantic import BaseModel, RootModel
 
 
 def main() -> None:
@@ -25,7 +25,9 @@ def main() -> None:
     sys.path.insert(0, str(repo_root / "src"))
 
     from inspect_ai._view._openapi import build_openapi_schema
+    from inspect_ai.log import EventsData
     from inspect_scout._llm_scanner.params import LlmScannerParams
+    from inspect_scout._scanner.types import ScannerInput, ScannerInputNames
     from inspect_scout._transcript.types import Transcript
     from inspect_scout._validation.types import ValidationCase
     from inspect_scout._view._api_v2 import v2_api_app
@@ -42,6 +44,13 @@ def main() -> None:
 
     class RawEncoding(RootModel[_RawEncoding]):
         pass
+
+    class ScannerInputResponse(BaseModel):
+        """Schema-only type for scanner input column values."""
+
+        input_type: ScannerInputNames
+        input: ScannerInput
+        input_data: EventsData | None = None
 
     # Create the real app, then add stub endpoints for scout-specific types.
     app = v2_api_app()
@@ -64,6 +73,10 @@ def main() -> None:
 
     @app.get("/schema/raw-encoding")
     def _raw_encoding() -> RawEncoding:
+        raise NotImplementedError
+
+    @app.get("/schema/scanner-input")
+    def _scanner_input() -> ScannerInputResponse:
         raise NotImplementedError
 
     schema = build_openapi_schema(app)
--- a/src/inspect_scout/_recorder/file.py
+++ b/src/inspect_scout/_recorder/file.py
@@ -7,6 +7,7 @@ import pandas as pd
 import pyarrow as pa
 import pyarrow.compute as pc
 import pyarrow.dataset as ds
+import pyarrow.fs as pafs
 import pyarrow.parquet as pq
 from inspect_ai._util.asyncfiles import AsyncFilesystem
 from inspect_ai._util.file import file, filesystem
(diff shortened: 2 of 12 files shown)


epatey:
@rasmusfaber, thanks for working on this. Here are some initial reactions.

- Can you help me understand how/when `scan_events` would be huge. Are we talking about when using an llm scanner? In that case, it's pretty much a one-shot, so that doesn't feel huge. I'm asking because I think it's important to understand the scenario in order to decide how much effort is justified.

- I may be misunderstand the code, but as currently coded I don't think you're achieving the savings you expect since `/scans/{dir}/{scan}/{scanner}` is still just passing `exclude_columns=["input"]`. 

- More generally, it seems that there has to be some sort of coordination between `/scans/{dir}/{scan}/{scanner}` and the new `../fields` endpoint. It has to be incredibly clear who is in charge. The client or the server? In other words, who decides which fields get omitted from the default dataframe endpoint? That same code needs to then make sure that `/fields` includes those fields.

I'm happy to chat about this if you'd like. I'll be in the office Friday by 9:15 ET or so.

rasmusfaber:
First: here are some loading times for loading a real scan without this fix:
Overview page:
Screenshot From 2026-04-10 10-56-29
Detail page:
Screenshot From 2026-04-10 10-57-53

(this is also without the fix to avoid it continuously reloading the page, so you will see some repetitions)

Here are the loading times for the same scan with the fix:

Overview page:
Screenshot From 2026-04-10 14-18-08
Detail page:
Screenshot From 2026-04-10 14-18-27

rasmusfaber:
> * Can you help me understand how/when `scan_events` would be huge. Are we talking about when using an llm scanner? In that case, it's pretty much a one-shot, so that doesn't feel huge. I'm asking because I think it's important to understand the scenario in order to decide how much effort is justified.

The scan I showed timings for above contains 495 transcripts. I examined one of the scan_events entries. That one contained 8 events. 1 span_begin, 6 model calls, and 1 span_end. Each of the model calls contains a different part of an actual transcript. Each model call is ~330KB, so approximately 2MB per scan_events uncompressed.

I don't know if METR are atypical, but these are very typical scans for us.

> * I may be misunderstand the code, but as currently coded I don't think you're achieving the savings you expect since `/scans/{dir}/{scan}/{scanner}` is still just passing `exclude_columns=["input"]`.

[No?](https://github.com/meridianlabs-ai/inspect_scout/pull/367/changes#diff-9b3ce714c0100c16b9f234ca5944c6c4c80988e47165020a66bbcbcec2111092R289)

> * More generally, it seems that there has to be some sort of coordination between `/scans/{dir}/{scan}/{scanner}` and the new `../fields` endpoint. It has to be incredibly clear who is in charge. The client or the server? In other words, who decides which fields get omitted from the default dataframe endpoint? That same code needs to then make sure that `/fields` includes those fields.

Yes, I think you are right, and also that this isn't currently the case. Which component would you prefer to own this? I think the API is clearer if it is the client that owns it (otherwise the API endpoint should rightly be called `scan_without_expensive_fields` or something similar). On the other hand the decision about which fields are expensive and which are not seems like a backend concern.

epatey:
> > * I may be misunderstand the code, but as currently coded I don't think you're achieving the savings you expect since `/scans/{dir}/{scan}/{scanner}` is still just passing `exclude_columns=["input"]`.
> 
> [No?](https://github.com/meridianlabs-ai/inspect_scout/pull/367/changes#diff-9b3ce714c0100c16b9f234ca5944c6c4c80988e47165020a66bbcbcec2111092R289)
> 
> >

I must have been looking at the wrong branch. Sorry about that. I have too many windows/worktrees. lol

PR #399, opened 2026-04-20
Add transcript search API

brandly:
This PR adds a new set of Search endpoints to support upcoming features in Scout's viewer. The central piece of data is a `SavedSearch` which is a union of `GrepSavedSearch` and `LlmSavedSearch`.

The API allows you to pick between an `llm` and `grep` search and provide options for that type of search. POSTing to `/search` for a given transcript will run a scan against that transcript and store it in Inspect's kvstore.

The other endpoints allow you to get a list of `SavedSearch`es for a given transcript, get a single `SavedSearch` based on its id, and delete a search based on its id.

In the case of an LLM scan, since we're calling out to upstream LLM providers, these calls could fail for a variety of reasons outside of our control, and we'll report the error to the client as a 502 Bad Gateway.


src/inspect_scout/_view/_api_v2_search.py
```diff
@@ -0,0 +1,256 @@
...
+If nothing in the transcript matches the query, say so briefly.
+
+{{ answer_format }}
+"""
+
+MAX_ENTRIES = 500
+SAVED_SEARCH_ADAPTER: TypeAdapter[SavedSearch] = TypeAdapter(SavedSearch)
+
+
+def _normalize_filter(value: object) -> object:
```
epatey, line 55: This is a little bit of a nit, but I think the type hint for this could be narrower avoiding the need for the assert. Seems like it could be `value: str | None | Sequence[object]) -> str | None | Sequence[str]`.

src/inspect_scout/_view/_api_v2_types.py
```diff
@@ -290,3 +291,79 @@ class StreamMetadata:
...
     compression_method: ZipCompressionMethod | None
     uncompressed_size: int | None
+
+
+class SearchRequestBase(BaseModel):
+    """Shared fields for search request variants."""
+
+    query: str
+
+    messages: MessageFilter = None
```
epatey, line 301: This is likely just fine, but if you search around, you'll see us set default values using something like ` = Field(default=None)`. IIRC, there's some really subtle reason why using Pydantic's `Field` is more correct in some cases - though I think for `None` it doesn't matter - other than consistency.

PR #406, opened 2026-04-27
Preserve explanation/metadata when scanner Result.value is None

kaifronsdal:
## Summary

`as_scorer()` previously returned `None` (no score) for any `Result` with `value=None`, silently dropping any `answer`, `explanation`, or `metadata` the scanner had attached.

This makes it impossible for a scanner to fail gracefully and still tell the user *why*. For example, when a judge model hits a content filter, the scanner can catch `RefusalError` and return `Result(value=None, explanation=str(ex), metadata={"refusal": True})` — but today that context is discarded and the sample just shows up with no score and no explanation.

With this change, `Result(value=None)` with an `answer`/`explanation`/`metadata` is recorded as a `Score(value=NOANSWER, ...)` so the context is preserved in the eval log. A bare `Result(value=None)` still returns no score, so existing behavior is unchanged for scanners that genuinely have nothing to report.

## Test plan
- [x] `pytest tests/scanner/test_as_scorer.py` (32 passed, including 2 new tests covering both `None` cases)
- [x] `ruff check` / `ruff format` / `mypy` on touched files


jjallaire:
@dragonstyle Would we be better off here with `value=None` rather than `value=NOANSWER` -- IIRC the NOANSWER constant is used for string based scorers that return "C" or "I". I'm wondering what will happen if we mix "NOANSWER" with values that are dicts.

jjallaire:
The right answer ended up being to use `value=float("nan")` (that's what Inspect supports already for scalars, needed to make a change to Inspect to also support NaN for dicts and lists)

jjallaire (approved):

PR #411, opened 2026-04-29
fix: support scans dir in logs dir by making location type determination look for *.eval files

ransomr:


jjallaire:
Unfortunately people can bring their own Parquet database using whatever filename convention they want so we can't use this filter: https://meridianlabs-ai.github.io/inspect_scout/db_importing.html#parquet-data-lake

I have been vexed by what to do about this! I wonder if we need to invert it and say "if there are .eval logs then it's inspect logs"

ransomr:
> Unfortunately people can bring their own Parquet database using whatever filename convention they want so we can't use this filter: https://meridianlabs-ai.github.io/inspect_scout/db_importing.html#parquet-data-lake
> 
> I have been vexed by what to do about this! I wonder if we need to invert it and say "if there are .eval logs then it's inspect logs"

@jjallaire switched to looking for *.eval

epatey (approved):

PR #420, opened 2026-05-04
Always run search for POST requests

brandly:
This is an internal change to the Search API (#407). We have a GET endpoint to get cached search results for a given transcript. Anytime we make a POST request, we should truly run the search instead of ever returning cached results.


dragonstyle (approved):
LGTM once tests are passing!

PR #424, opened 2026-05-08
Expose recorder primitives for inspect_ai eval_set scanner integration

ransomr:
## Summary

Adds the scout-side support `inspect_ai.eval_set` needs to dispatch scanners per-sample as transcripts are produced (rather than running scout against a finished log directory). Public-facing scout commands (`scout scan`, `scout scan-resume`, `scout scan-complete`) continue to behave the same; this PR adds new opt-in primitives + recorder modes that the inspect_ai integration consumes.

inspect_ai already uses these on its `eval-set-scan` branch — see `inspect_ai/design/eval-set-scanners.md` for the integration design.

## Changes

**Exposed primitives** (commit `eba73933`)

- `Scanners` type alias — extracts the inline shape of `scan` / `scan_async`'s `scanners` arg so external callers can annotate the same parameter without re-spelling the union. Exported from the package init.
- `_scan_one(job, *, validation, fail_on_error)` — lifted from `_scan_function`'s closure inside `_scan_async_inner` to a module-level function in `_scan.py`. Lets external callers run a single scanner against a single in-memory transcript without spinning up `scan_async`'s recorder/display/multi-process machinery. The internal closure now delegates.
- `concurrent_writers: bool = False` opt-in kwarg on `RecorderBuffer.__init__`, `FileRecorder.init`, `FileRecorder.resume`, and `FileRecorder.attach`. When `True`, summary read-modify-write in `record()` / `record_metrics()` is wrapped in a per-buffer-dir `anyio.Lock` and the on-disk summary is re-read inside the lock so concurrent ephemeral `RecorderBuffer` instances don't lose updates. Default `False` preserves existing single-writer behavior; scope is in-process only.

**`FileRecorder.attach`** (commit `8c7f6e50`)

Distinguishes "add records for new transcripts" from `resume()` ("retry errored transcripts"). Three modes for opening an existing scan dir, each with different setup:

| Method | Buffer summary | `_errors.jsonl` | Use case |
| --- | --- | --- | --- |
| `init` | fresh | fresh | New scan |
| `resume` | seeded from scan_dir | **truncated** | Retry errored transcripts |
| `attach` | seeded from scan_dir | **seeded from scan_dir** | Add records for new transcripts; preserve prior errors |

inspect_ai's per-sample dispatch always uses `attach` (the per-sample resume-scan check filters out already-recorded tids before dispatch).

**`_compact_with_prior`** (commit `2d8afb2a`)

Each `sync()` call may need to merge its buffer rows into a previously-written compacted parquet — e.g. on a multi-call eval_set lifecycle where the first call wrote the compacted output and a subsequent call adds new transcripts. `_compact_with_prior` reads the prior compacted parquet and passes it to `scanner_table` as `extra_inputs`. When the prior is on a remote filesystem, it's downloaded to a local temp file before passing in (pyarrow's `ds.dataset(list)` requires uniform local paths).

**Resume after completed scan** (commit `3f23b2be`)

`_seed_buffer_summary_from_scan_dir` / `_seed_buffer_errors_from_scan_dir` bootstrap the buffer's `_summary.json` / `_errors.jsonl` from the scan dir on `resume` / `attach`. After `sync(complete=True)` cleans the buffer, the cumulative summary and errors only live on the scan dir; this seeds them back so the new `RecorderBuffer` picks up prior counts instead of starting fresh.

**`scanner_table` dedup against `buffer_tids`** (commit `5452a739`)

The buffer's per-transcript file `<tid>.parquet` is authoritative for that transcript_id. Any rows for the same transcript_id in `extra_inputs` (a previously-compacted output that was built from an earlier copy of these same buffer files) are duplicates and must be filtered out — otherwise rows double-count when `sync` runs more than once without clearing the buffer between calls (e.g. an eval_set resume cycle where `complete=False`).

Includes new tests in `tests/scanner/test_recorder_buffer.py` covering the dedup invariant.

## Backward compatibility

- All new behavior is opt-in:
  - `concurrent_writers` defaults to `False`.
  - `attach` is a new method; `init` / `resume` keep their existing semantics.
  - `_scan_one` extraction is a refactor — `_scan_function` continues to delegate to it.
  - `_compact_with_prior` only kicks in when a prior compacted parquet exists.
- No on-disk format changes. Scans written by inspect_ai's eval_set integration are byte-equivalent to those `scout scan` produces and remain readable by all existing scout tooling.

## Test plan

- [x] Existing scout tests pass.
- [x] New tests in `tests/scanner/test_recorder_buffer.py` cover the dedup invariant.
- [x] inspect_ai's `tests/test_eval_set_scanner.py` (~95 tests, including end-to-end S3-mocked runs) exercises these primitives via the integration layer.

### DIFF
--- a/src/inspect_scout/__init__.py
+++ b/src/inspect_scout/__init__.py
@@ -28,7 +28,7 @@ from ._scan import (
     scan_complete,
     scan_resume,
 )
-from ._scanjob import ScanJob, scanjob
+from ._scanjob import ScanJob, Scanners, scanjob
 from ._scanjob_config import ScanJobConfig
 from ._scanlist import scan_list
 from ._scanner.extract import (
@@ -121,6 +121,7 @@ __all__ = [
     "scanjob",
     "ScanJob",
     "ScanJobConfig",
+    "Scanners",
     "ProjectConfig",
     "scan_list",
     "scan_status",
--- a/src/inspect_scout/_recorder/buffer.py
+++ b/src/inspect_scout/_recorder/buffer.py
@@ -75,33 +75,34 @@ class RecorderBuffer:
         spec: ScanSpec,
         *,
         pool_dedup: bool = True,
-        reset: bool = False,
         concurrent_writers: bool = False,
     ):
+        """Initialize a buffer attached to `scan_location`.
+
+        This is purely a passive constructor: it reads existing buffer state
+        from disk or creates fresh files only if they don't exist. Mode-
+        specific setup (clearing for a fresh scan, truncating errors for a
+        retry-resume, preserving errors for a continuation) is the caller's
+        responsibility — see `FileRecorder.init` / `resume` / `attach`.
+        """
         self._buffer_dir = RecorderBuffer.buffer_dir(scan_location)
         self._buffer_dir.mkdir(parents=True, exist_ok=True)
         self._spec = spec
         self._pool_dedup = pool_dedup
         self._concurrent_writers = concurrent_writers
 
-        # establish scan summary
+        # read summary if present, else create fresh
         scan_summary_file = self._buffer_dir.joinpath(SCAN_SUMMARY)
-        if reset or not scan_summary_file.exists():
+        if scan_summary_file.exists():
+            self._scan_summary = read_scan_summary(self._buffer_dir, spec)
+        else:
             self._scan_summary = Summary(
                 complete=False, scanners=list(spec.scanners.keys())
             )
             with open(scan_summary_file.as_posix(), "w") as f:
                 f.write(self._scan_summary.model_dump_json(indent=2))
-        else:
-            self._scan_summary = read_scan_summary(self._buffer_dir, spec)
 
-        # Always truncate errors. On resume, previously-errored transcripts are
-        # re-processed (is_recorded returns False for errored parquets), so stale
-        # error entries must not persist -- they would incorrectly mark the scan
-        # as incomplete even if the re-run succeeds.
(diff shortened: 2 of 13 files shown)


src/inspect_scout/_recorder/buffer.py
```diff
@@ -56,31 +75,34 @@ def __init__(
         spec: ScanSpec,
         *,
         pool_dedup: bool = True,
-        reset: bool = False,
+        concurrent_writers: bool = False,
```
jjallaire: Would it be simpler and w/o performance cost to just always enforce the `concurrent_writers` constraint? (that way others using this plumbing in the future get the protection by default).

src/inspect_scout/__init__.py
```diff
@@ -122,6 +122,7 @@
     "scanjob",
     "ScanJob",
     "ScanJobConfig",
+    "Scanners",
```
jjallaire, line 125: Should also add this to the `reference/scanner.qmd` file.

src/inspect_scout/_recorder/buffer.py
ransomr: https://github.com/meridianlabs-ai/inspect_scout/pull/424/commits/bc3924c705967cc2b288100e379e70da6414359e

src/inspect_scout/__init__.py
ransomr: https://github.com/meridianlabs-ai/inspect_scout/pull/424/commits/02ceb73cb5fe3bf75aa825a6736df83028940506

PR #433, opened 2026-05-14
Fix scan_results_df crash on numeric resultsets with empty rows

rasmusfaber:
## Summary

`scan_results_df` (and any downstream consumer) crashes with `TypeError: int() argument must be ... not 'NAType'` when a scan combines a resultset whose values are numeric with at least one row whose pyarrow-backed `value` column is all-NA (e.g. a scanner that timed out, or a non-resultset row mixed in). The dtype-alignment loop picks `numpy.int64` as the target dtype from the numeric result, then tries to `astype()` the all-NA pyarrow array onto it, which routes through `to_numpy(na_value=pd.NA)` and explodes.

Fall back to `object` dtype when the target dtype can't represent NA. That matches the existing fallback used when a column is missing entirely from one of the dataframes, and lets the downstream `pd.concat` handle the values without further coercion.


rasmusfaber:
(The schema error is pre-existing on main)

PR #434, opened 2026-05-14
Add context_tools to generate_answer / structured_generate

kaifronsdal:
## Summary

Adds `context_tools: Sequence[ToolInfo] = ()` to `generate_answer()` and `structured_generate()`. The tools are declared alongside the answer tool in the request (`tools=[*context_tools, answer_tooldef]`) but never invoked, since `tool_choice` already forces the answer tool.

## Motivation

When asking a structured follow-up question about an existing transcript — e.g. appending `ChatMessageUser("Was the conversation above an evaluation?")` to a message history that contains native `tool_use` blocks — the API rejects the request unless every tool referenced in those blocks is declared in `tools=`. Currently `structured_generate` hardcodes `tools=[answer_tooldef]`, so callers must either render the transcript to text (losing the native tool-call signal) or reimplement the retry/validation loop themselves.

With this change:
```python
result = await generate_answer(
    [*transcript_messages, ChatMessageUser(content=probe_question)],
    AnswerStructured(ProbeAnswer),
    context_tools=transcript_tool_infos,
)
```

## Test plan

- [x] `make check` (ruff + mypy)
- [x] `tests/llm_scanner/test_generate.py` — new `test_generate_structured_context_tools` verifies the tools list is `[*context_tools, answer]` and `tool_choice` still forces `answer`


jjallaire:
I think this change definitely works as intended, but I'm wondering whether the client shouldn't be passing `messages_as_str()` rather than `list[ChatMessage]` (as we have observed that sometimes models aren't able to distinguish messages played back to them as "another conversation". This is e.g. how we implement llm_scanner and the petri judge.

kaifronsdal:
The usecase here is exactly that: the transcript was from the model itself, we're just asking a followup question (plus it takes advantage of prompt caching as a side effect)

PR #449, opened 2026-05-22
fix: serialize metadata flds with pydantic_core.to_json

ekin-aisi:
## Summary

- Replace `json.dumps` with `pydantic_core.to_json(v, fallback=str)` in the two parquet writers that touch user-supplied metadata, so a `pathlib.Path` (or any non-JSON-native object) in metadata produces a clean JSON string instead of Python repr text in the parquet column.
- Add tests.

## Original issue

When a scan runs through inspect_ai's `eval_set(scanner=...)` integration (the online path added in #424 ), `eval_sample.metadata` flows directly into `TranscriptInfo.metadata` without prior stringification. If an eval put a `Path` into sample metadata, the `_normalize_scalar` helper in `_recorder/buffer.py` hit:

```python
try:
    return json.dumps(v, ensure_ascii=False, separators=(",", ":"))
except Exception:
    return str(v)
```

`json.dumps` raises `TypeError: Object of type PosixPath is not JSON serializable`, the `except` branch fires, and `str(dict)` yields:

```
"{'eval_name': 'foo', 'eval_file_path': PosixPath('/home/foo/eval.yaml')}"
```

That Python-repr text lands in the parquet `transcript_metadata` column, the viewer can't parse it, and the user sees **"No scan results are available."** in the browser & **"The editor could not be opened due to an unexpected error"** in VS Code.

## Fix
- `pydantic_core.to_json(v, fallback=str)`: natively handles `Path`, `UUID`, `Decimal`, `datetime`, `set`, pydantic models and falls back to `str()` per-value for unknowns — `str(Path)` returns the path string, not the `PosixPath(...)` constructor literal.
- Inspired by inspect_ai's `to_json_str_safe`, which is already used in `_scanner/result.py`.

Two call sites had the same shape of bug:

- `src/inspect_scout/_recorder/buffer.py` — `_normalize_scalar` (scan-results writer, the user-reported path).
- `src/inspect_scout/_transcript/database/parquet/transcripts.py` — `_transcript_to_row` (transcripts DB writer; analogous risk, would have crashed instead of producing repr text since there's no surrounding try/except there).

| Input | Old output | New output |
|---|---|---|
| `{"eval_file_path": Path("/home/foo/eval.yaml")}` | `"{'eval_file_path': PosixPath('/home/foo/eval.yaml')}"` | `'{"eval_file_path":"/home/foo/eval.yaml"}'` |

## Alternative FIx

`src/inspect_scout/_transcript/eval_log.py:748` could also fix at the entry-level here `metadata=dict(eval_sample.metadata),`, did not want to touch the in-memory metadata unless we want this kind of defense-in-depth

### Testing

- [x] New: `tests/scanner/test_recorder_buffer.py::test_record_serializes_path_in_metadata_as_json` — records a transcript with a top-level `Path` plus a nested dict-of-Path, reads the parquet back, asserts the JSON parses and contains no `PosixPath` literal.
- [x] New: `tests/transcript/test_parquet_db.py::test_path_in_nested_metadata_stringified` — round-trips a transcript with `Path` values nested in dicts and lists through `ParquetTranscriptsDB.insert` / `select`, asserts the values come back as path strings.
- [x] `pytest tests/recorder/ tests/scanner/test_recorder_buffer.py tests/transcript/test_parquet_db.py` 
- [x] `ruff check` + `ruff format`
- [x] `mypy src examples tests`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,10 +4,11 @@
 - Structured generate: Warn model for orphan tool_use in structured_generate retry loop.
 - Enumerate all CSV, YAML, and JSON files when looking for validation sets (don't skip gitignored files).
 - Support custom role labels for transcript message rendering.
+- Scout View: Refine scanner result header and All Scores dialog
 - Bugfix: Stop doing blocking S3 file I/O on hot paths.
 - Bugfix: Resolve nested schema references for `AnswerStructured`.
 - Bugfix: `llm_scanner` now reserves tokens for the rendered scanner template when sizing segments, so long templates no longer push the prompt past `context_window`.
-- Scout View: Refine scanner result header and All Scores dialog
+- Bugfix: Serialize metadata flds with pydantic_core.to_json
 
 ## 0.4.35 (16 May 2026)
 
--- a/src/inspect_scout/_recorder/buffer.py
+++ b/src/inspect_scout/_recorder/buffer.py
@@ -1,4 +1,3 @@
-import json
 import os
 import shutil
 from dataclasses import dataclass
@@ -15,6 +14,7 @@ from inspect_ai._util.file import file
 from inspect_ai._util.hash import mm3_hash
 from inspect_ai.scorer import value_to_float
 from pydantic import JsonValue
+from pydantic_core import to_json
 from upath import UPath
 
 from inspect_scout._recorder.summary import Summary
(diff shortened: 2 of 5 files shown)


dragonstyle:
This looks good to me! @jjallaire you ok with this?

PR #457, opened 2026-06-04
Static Scout bundles

brandly:
This PR introduces `scout view bundle` which produces a read-only snapshot of a Scout project that can be
served from ordinary static hosting -- no FastAPI view server needed. The provided `design/static-bundles.md` describes these bundles in detail.

The changes here are additive and introduce the new CLI command as well as the logic for generating the necessary static files.

View an example here: https://brandly.github.io/static-scout-example/


brandly:
Leaving this in a draft for now -- having second thoughts about the scans approach and want to conduct some experiments

PR #477, opened 2026-06-19
Cap oversized tool/text output to prevent resize layerization stalls

dragonstyle:
## Problem

Viewing a transcript with a very large tool result (a single ~15 MB command output) made the scout viewer laggy on resize in Chrome and **spinlock in Safari**.

A Chrome performance trace of one resize showed the cost was **~14s in `Layerize`** (49 calls, up to 1.4s each) plus ~2.7s Layout / ~1.6s Paint. The output renders as one **~1,000,000px-tall DOM node**; the compositor re-layerizes that entire node on every resize. It's a compositor cost (invisible to JS profiling), and a fixed-height scroller does **not** help — the off-screen content is still layerized.

## Fix

Frontend changes land in meridianlabs-ai/ts-mono#341; this PR bumps the submodule pointer and rebuilds the shipped viewer bundle.

- Cap rendered text at **250k chars** (`cappedText` helper) in both `ToolTextOutput` and `RenderedText`, with a static "Output truncated — N hidden" notice. No expand affordance — revealing the full content would reintroduce the cliff.
- `contain: layout paint` on the collapsed `ExpandablePanel` wrapper as defense-in-depth.

## Verification

On the reproducing transcript, the largest rendered node dropped from **~1,045,498px → ~55,478px** tall. `pnpm check` and `pnpm test` pass.

> Note: `Layerize` cannot be measured headlessly, so the final smoothness check is a manual resize in Chrome/Safari on the built bundle.

### DIFF
--- a/src/inspect_scout/_view/dist/assets/index-Bd5cILag.css
+++ b/src/inspect_scout/_view/dist/assets/index-Bd5cILag.css
@@ -1,3 +1,0 @@
-version https://git-lfs.github.com/spec/v1
-oid sha256:616c44f82b4082022299e2aa1b2fe665fff7565e5817657096270ccc6aac47be
-size 529594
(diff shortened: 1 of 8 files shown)


jjallaire:
some conflicts to resolve....

PR #479, opened 2026-06-25
fix: stop storing custom validation predicates as executable dill payloads in scan specs

ericwinsor-aisi:
## PR stack

**Primary implementation PR:** this PR

Prerequisites / companions:
- Inspect registry prerequisite: https://github.com/UKGovernmentBEIS/inspect_ai/pull/4359
- ts-mono generated types companion: https://github.com/meridianlabs-ai/ts-mono/pull/368

## This PR contains:

- [x] New features
- [x] Docs
- [x] Bug fixes
- [x] Code refactor

Companion changes:

- Inspect prerequisite: add `validation_predicate` to the shared
  `RegistryType`.
- ts-mono companion: regenerate Scout API types for the data-only predicate
  specifications.

### What is the current behavior?

`ValidationSet.predicate` currently accepts either a built-in predicate name or
an arbitrary Python callable.

Built-in predicates such as `eq`, `gte`, and `contains` are persisted in
`_scan.json` by name. Custom callables are serialized with dill, base64-encoded,
and stored in the same field:

```json
{
  "predicate": "gASV..."
}
```

When Scout reads a `ValidationSet`, every string that is not a built-in
predicate name is treated as a possible serialized callable and passed to
`dill.loads()`.

This read occurs during ordinary scan operations including:

- listing or indexing scans;
- opening a completed scan;
- reading scan status; and
- resuming an interrupted scan.

Pickle and dill deserialization can execute Python before Scout can validate
that the resulting object is an async predicate. A scan artifact can therefore
execute code merely by being listed or opened.

The callable round-trip was intentional. Validation originally stored custom
predicates with dill, `ScanSpec` later persisted the validation set, and
`resume_scan()` restored `spec.validation`. Removing custom predicates
altogether would close the vulnerability, but would also remove an intended
extension and resume workflow.

### What is the new behavior?

This PR separates the live validation configuration from its portable artifact
representation.

- `ValidationSet` remains the runtime type and continues to accept trusted
  async callables.
- `ScanContext.validation` retains the live callable while the scan is running.
- `ScanSpec.validation` uses a data-only `ValidationSetSpec`.
- `_scan.json` never contains callable bytes, source, closure state, or another
  executable object.
- Listing, opening, and indexing scans never import or resolve predicate code.

Portable predicates have three forms:

| Predicate | Persisted form | Resume behavior |
|---|---|---|
| Built-in | Existing built-in name | Resumes without loading custom code |
| Registered custom | Registered name plus Inspect registry-compatible creation parameters and source/package metadata | Recreated at the explicit trusted resume boundary |
| Anonymous or legacy custom | Inert unavailable marker | Requires the original trusted configuration or an explicit override |

This follows the existing Inspect extension pattern used by scorers, solvers,
metrics, tasks, and Scout scanners: persist a recreatable specification, not
the Python object itself.

This PR depends on a small Inspect prerequisite that adds
`validation_predicate` to the shared `RegistryType`. Predicate factories use
the standard `RegistryInfo`, `registry_add`, `registry_tag`, `registry_lookup`,
and `registry_params` path; Scout does not maintain a second registry.

The persisted `args` object is therefore the standard Inspect registry parameter
representation, not a lossless serialization of arbitrary Python values or an
exact JSON round-trip. Predicate factories should follow the same conventions as
other Inspect extensions: ordinary serializable values and registered Inspect
objects or models are portable, while arbitrary Python objects are not. Non-finite
floating-point values are rejected rather than silently written as `null`.

### Defining a portable custom predicate

The new `@validation_predicate` decorator registers a predicate factory in the
same style as `@scorer`, `@solver`, and `@scanner`.

```python
from pydantic import JsonValue

from inspect_scout import (
    Result,
    ValidationPredicate,
    validation_predicate,
    validation_set,
)


@validation_predicate
def within_range(distance: float = 10) -> ValidationPredicate:
    async def check(result: Result, target: JsonValue) -> bool:
        if not isinstance(result.value, (int, float)):
            return False
        if not isinstance(target, (int, float)):
            return False
        return abs(result.value - target) <= distance

    return check


validation = validation_set(
    "validation.csv",
    predicate=within_range(distance=5),
)
```

Scout records the registered identity and Inspect registry-compatible creation
parameters as data:

```json
{
  "kind": "registered",
  "name": "my_package/within_range",
  "args": {
    "distance": 5
  },
  "package_version": "1.2.3"
}
```

The predicate is not imported or constructed when `_scan.json` is parsed. It
is resolved only when the caller explicitly resumes the scan through a trusted
registry or source decision.

### Existing built-in predicates are unchanged

Built-in predicates keep their current API and behavior:

```python
validation = validation_set(
    "validation.csv",
    predicate="gte",
)
```

CSV, YAML, JSON, and UI-authored per-case predicate names are also unchanged.

### Anonymous predicates still work during the current run

This remains valid:

```python
async def normalized_equal(result: Result, target: JsonValue) -> bool:
    return str(result.value).strip().lower() == str(target).strip().lower()


validation = validation_set(
    "validation.csv",
    predicate=normalized_equal,
)
```

The callable continues to work in the original single-process or
multiprocessing scan. Scout records an inert marker rather than the callable:

```json
{
  "kind": "unavailable",
  "display_name": "normalized_equal",
  "reason": "anonymous"
}
```

Scout emits a bounded warning that artifact-only resume will require the
original trusted configuration or a predicate override.

### Resuming with an explicit predicate override

Python callers can supply the trusted callable again:

```python
scan_resume(
    scan_location,
    predicate_overrides={
        "my_scanner": normalized_equal,
    },
)
```

The persisted validation cases and split remain authoritative. The override
only supplies the missing implementation.

Scout rejects missing or extra overrides, validates that each override is
async, and never silently replaces a missing predicate with `eq`.

### Legacy scans

Existing scans containing dill payloads are not deserialized.

When Scout encounters an old non-built-in predicate string, it:

- does not test whether the value is base64;
- does not decode, hash, import, return, or log the value;
- replaces it in memory with an unavailable legacy marker; and
- retains the validation cases and recorded results.

Completed legacy scans therefore remain listable and reviewable. Resuming one
requires the original trusted predicate or an explicit override.

There is intentionally no `--allow-pickle`, environment variable, or viewer
button that restores legacy deserialization.

### Does this PR introduce a breaking change?

Yes. The change is intentionally incompatible with treating a portable scan
artifact as an executable snapshot.

#### Unchanged

- Built-in predicate names and semantics.
- Per-case predicates loaded from CSV, YAML, JSON, JSONL, DataFrames, or Scout
  View.
- Custom callable behavior during the original scan.
- Custom callable transport to trusted worker processes.
- Completed validation results and scan viewing.
- Automatic resume for built-in predicates.

#### Changed

- Custom callables are no longer serialized into `_scan.json`.
- `ValidationSet.model_dump()`/JSON output no longer contains a base64 dill
  representation of a callable.
- Deserializing a `ValidationSet` no longer reconstructs arbitrary callable
  objects from strings.
- An anonymous local function or closure can no longer resume from the scan
  artifact alone.
- `scout scan resume LOCATION` can no longer resume an interrupted anonymous
  custom-predicate scan by location alone. The predicate must be registered,
  supplied by the original trusted scan job, or passed through the Python
  override API.
- Existing dill-bearing scans require a trusted override to resume.
- Moving an anonymous custom-predicate scan to another machine no longer moves
  the predicate implementation with it.
- The existing `test_custom_predicate_serialization` dill round-trip contract
  is intentionally removed and replaced with registered-spec and unavailable
  marker coverage.

#### Migration

Users who need portable resume should decorate the predicate with
`@validation_predicate` and use Inspect registry-compatible creation
parameters. Ordinary serializable values and registered Inspect objects or
models are portable; arbitrary Python objects are not.

Users who intentionally need an anonymous closure can continue to use it for
the current run and provide it through `predicate_overrides` when resuming.

There is no safe migration that extracts an arbitrary callable from an
existing dill payload without executing that payload. Legacy scan artifacts
are therefore never auto-converted.

### Why not make custom predicates built-in-only?

Restricting `ValidationSet.predicate` to built-in names is the smallest
security fix, but it removes an intentional public extension point.

The registered-spec approach keeps the common portable/resumable workflow and
aligns validation with the existing extension model used throughout Inspect
and Scout.

### Why not deserialize dill only during `scan resume`?

Moving `dill.loads()` from scan listing to resume would remove passive code
execution, but the artifact would still contain executable state and would
still select code during an ordinary first-party command.

It would also leave scan bundles and copied scan directories as executable
artifacts. This PR removes executable predicate state from the portable format
rather than moving the same trust problem to another call site.

### Why not serialize the function's module and qualified name automatically?

Not every callable is safely or reliably recreatable by import:

- nested functions and lambdas have no stable import path;
- a module attribute may change between scan and resume;
- importing a path selected by an artifact is itself an active trust decision;
  and
- converting every callable into an import instruction would make accidental
  persistence easy.

Registration is the predicate author's explicit declaration that the factory
is stable and recreatable. Package version, source, and revision metadata can
then detect drift without treating a hash as authorization.

### Security properties

This PR establishes the following invariants:

1. Parsing `_scan.json` is data-only.
2. Passive scan operations perform no predicate import, construction, or call.
3. `ScanSpec` never contains a callable or serialized callable.
4. Registered code is resolved only at an explicit active trust boundary.
5. Legacy payloads are never decoded or echoed.
6. Missing predicate implementations fail closed without changing validation
   semantics.

### Testing

The security and compatibility coverage includes:

- a captured legacy dill payload causes no marker, decode, import, or registry
  access through `ValidationSetSpec`, `ScanSpec`, `FileRecorder.status()`, and
  the scan-list API;
- a malicious and valid scan can be listed together, with only the legacy
  predicate marked unavailable;
- every built-in predicate and `None` round-trips and resumes unchanged;
- a registered custom predicate with standard Inspect registry parameters
  round-trips and resumes with the same behavior;
- non-finite registered parameters fail before `_scan.json` is written;
- predicate and scanner definitions sharing one source file are loaded once
  during resume;
- listing a registered predicate performs no import or construction;
- a nested custom callable still works in single-process and multiprocessing
  scans;
- `_scan.json` contains no callable bytes, source, closure cells, or automatic
  import instruction;
- anonymous and legacy predicates require an exact trusted override;
- missing or extra overrides fail before model, scanner, network, import, or
  result-writing activity;
- no custom predicate form silently becomes `eq`; and
- existing validation and scan-listing tests remain green.

Validation performed for this branch:

- Focused validation, recorder, scan-index, and E2E tests passed; the follow-up
  hardening was additionally verified with 108 validation/E2E tests and 14
  scanner resume/worklist tests.
- Full mypy passed.
- Ruff 0.15.12 lint and format checks passed.
- OpenAPI schema regenerated.
- The complete Scout suite for the initial implementation produced 3,084 passes
  and 111 skips. Seven unrelated failures (five Trio/asyncio limiter tests and two Rich
  progress-capture tests) reproduce identically on upstream `main` in this
  environment.


dragonstyle:
Pushed a maintainer commit to this branch (c872371c) so it can land without another round trip:

- Merged current `main` (conflicts were `CHANGELOG.md`, which Release Please now owns, so the hand-written entry is dropped and will come from the squash title; and `uv.lock`, taken from main minus the two `dill` lines).
- Docstrings on the three spec types with the OpenAPI descriptions regenerated; an actionable error when resuming a scan whose predicate isn't registered; a docs paragraph on source-file loading at resume; two negative tests showing the serialized-callable path is gone.

Verified: a crafted `_scan.json` payload executes on current `main` during parsing and is inert here; full suite 3319 passed, 0 failed; ruff, mypy, and schema drift clean. Reassigning to @epatey for the final look and landing. Thanks for the design here — the registered-predicate approach is the one we want.

dragonstyle:
Status of the red checks: the PR title is fixed for the title-lint workflow, and the suppressions ledger is updated on the branch. The remaining red `check-schema-and-types` is expected until the companion viewer-types PR (meridianlabs-ai/ts-mono#368, refreshed against this branch's schema) lands on ts-mono `main`; after that the submodule here gets re-pinned to that main commit and both checks go green.

epatey (approved):
Approving. Reproduced the crafted `_scan.json` payload executing on `main` during parse and confirmed it is inert here; full suite, ruff, and mypy clean locally. Two non-blocking follow-ups:

- `_validation/registry.py:25`: `PredicateFactory = Callable[P, ValidationPredicate]` lets a factory return a built-in name string that the wrapper then rejects at runtime. Narrowing to `Callable[P, PredicateFn]` moves that to mypy (the docs example would annotate `-> PredicateFn`).
- `ScanSpec` is public and now references `ValidationSetSpec` / `RegisteredPredicateSpec` / `UnavailablePredicateSpec`, but those are only exported from `_validation` and not on the reference page. Worth exporting the three at top level and listing them in `docs/reference/results.qmd`.

PR #488, opened 2026-06-29
Honour `--log-level` consistently across CLI commands

craig-meridian:
## Summary

The `--log-level` common CLI option was accepted by many commands but only actually honored by `scout scan` and `scout view`. Specifically:

- **`scout import`** never initialized logging at all — the flag was completely inert.
- **`scout scan resume` / `scout scan complete`** accepted `--log-level` but never forwarded it into the runtime, so it was silently dropped.
- **`scout scan list` / `scout scan status`** accepted it but never applied it.
- **`scout scan`** worked correctly (via its own scanjob-aware resolution).
- **`scout view`** worked, but its project-config fallback was effectively dead code.

This PR makes every command that offers `--log-level` resolve and apply it identically.

## Approach

Resolution and logging initialization are centralized in `process_common_options` (the shared `@common_options` entry point that every such command already calls first):

- A new `resolve_log_level(ctx, options)` applies the precedence **CLI arg > project config > env var > default**.
- The resolved value is written back into `options["log_level"]`, so commands that forward it into the scan/view runtime pass an already-resolved value.
- Logging is initialized once, at the CLI entry point, for every command — including the ones (like `import`) that never enter the async runtime.

## Design decisions & rationale

**Precedence is `CLI arg > project > env > default` (project beats env).** This matches the existing, deliberate `scout scan` behavior. **This might be contentious** and I'd be open to going a different direction if we're willing to change how `scout scan` currently applies precedence.

**Write the resolved value back into `options["log_level"]`.** Commands that enter the scan/view runtime forward `options["log_level"]`, and the runtime evaluates `log_level or project.log_level`. Because the written-back value is always truthy (worst case the `"warning"` default), that `or` short-circuits to our resolved value — so the single resolution in `process_common_options` stays authoritative and the runtime never re-resolves and clobbers it (e.g. overriding an explicit `--log-level` with a project value). Without the write-back, a command with no flag but a project `log_level` would forward the bare default and the runtime's project fallback would never fire — which is exactly why `view`'s fallback was previously dead.

**Keep the runtime's `log_level or project.log_level`.** `scan`, `scan_resume`, `scan_complete`, and `view` are part of the public programmatic API (exported in `__all__`) and can be called directly from Python without going through the CLI. For those callers, `resolve_log_level` never runs, so the runtime fallback remains load-bearing — it's the only thing that supplies a project default. It is redundant only for the CLI path (where it short-circuits).

**`scan` keeps its own resolution and defers logging init.** `scan` resolves `log_level` against the merged scanjob value (which can come from a scanjob config *file*, not just the project), so it retains its existing `resolve_scan_option` handling. Because the logging handler is installed once per process (first caller wins), `scan` passes `init_logging=False` to `process_common_options` and initializes logging itself *after* the scanjob value is known — otherwise the eager init would lock the handler at the less-specific (project/default) level and silently drop the scanjob file's `log_level`. A subprocess regression test asserts the scanjob-file level is actually applied to the console logger.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Scout View: Fix broken commit links for ssh-style GitHub origins
 - Scout View: Improve MathJax Sanitization
 - Bugfix: Per-key validation metrics for dict/multi-label targets now bucket each key against its own target value. Previously the whole-dict positivity (always positive for a non-empty dict) was applied to every key, so a key whose target was `False`/negative was scored TP/FN instead of TN/FP, inflating per-key precision and specificity.
+- Bugfix: Respect `--log-level` CLI arg and `SCOUT_LOG_LEVEL` env var across all `scout` commands.
 
 ## 0.4.41 (12 June 2026)
 
--- a/src/inspect_scout/_cli/common.py
+++ b/src/inspect_scout/_cli/common.py
@@ -3,11 +3,15 @@ import os
 from typing import Any, Callable, Literal, TypeVar, cast
 
 import click
+from click.core import ParameterSource
 from inspect_ai._util.constants import ALL_LOG_LEVELS, DEFAULT_LOG_LEVEL
 from typing_extensions import TypedDict
 
+from inspect_scout._concurrency._mp_common import set_log_level
 from inspect_scout._display._display import DisplayType, display, init_display_type
+from inspect_scout._project._project import read_project
 from inspect_scout._util.constants import DEFAULT_DISPLAY, DEFAULT_SERVER_HOST
+from inspect_scout._util.log import init_log
 
 F = TypeVar("F", bound=Callable[..., Any])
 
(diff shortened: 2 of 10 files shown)


epatey (commented):
- Please add a `CHANGELOG.md` entry
- Here's one piece of feedback from Codex that seems legit.

**High**: `scout view <project_dir>` still ignores that project’s `log_level`.
`src/inspect_scout/_cli/view.py:57` resolves common options before `view()` changes into `project_dir; src/inspect_scout/_cli/common.py:117` therefore reads the caller cwd project, not the target project. Then `view.py:68` passes an always-truthy `common["log_level"]`, so `src/inspect_scout/_view/view.py:31-34` never falls back to the target project config. I reproduced this on the PR code: target `scout.yaml` had `log_level: debug`, cwd had no project, and `scout view <project> --no-browser` initialized at warning.

PR #489, opened 2026-06-30
Store transcript/scan-result columns as compact JSON

craig-meridian:
## Summary

The transcript and scan-result write paths serialized condensed events and pooled inputs with the default `indent=2`. The pretty-print whitespace bloats both the on-disk parquet columns and the bytes streamed to the viewer. These columns are machine-read (`json.loads` / `expand_events`) and as far as I'm aware, never human-consumed.

This switches them to compact JSON (`indent=None`) via a shared `to_json_compact` helper.

## Motivation

A large transcript which I've been working with has an `events` column which is ~700 MiB. This is largely because the model events' `input_refs` are is made up of many non-contiguous event index ranges. This occupies more than 95% of the file. Granted, when written to disk as a Parquet file, this will be heavily compressed. But when streamed to the viewer (`scout view`), it exceeds the [existing 350 MiB limit](https://github.com/meridianlabs-ai/inspect_scout/blob/c6ac88d612bfdb1d196170beabb6afe332a2c074/src/inspect_scout/_view/_api_v2_transcripts.py#L35). With `indent=None` the 700 MiB becomes 200 MiB.

## Downsides

If the files or parquet columns are human-read (e.g. for debugging), it will be harder to follow. But a simple `jq . events.json > events-out.json` will restore the whitespace.

### DIFF
--- a/src/inspect_scout/_util/_json.py
+++ b/src/inspect_scout/_util/_json.py
@@ -1,0 +1,16 @@
+from typing import Any
+
+from inspect_ai._util.json import to_json_safe
+
+
+def to_json_compact(x: Any) -> str:
+    """Serialize to JSON without pretty-print whitespace.
+
+    Wraps ``to_json_safe`` (surrogate-safe) with ``indent=None``. Use for
+    machine-read payloads such as stored parquet columns and bytes streamed to
+    the viewer, where the default ``indent=2`` only bloats the representation
+    without benefit since the data is never human-consumed.
+
+    In extreme cases, this reduces serialized JSON from 700 MiB -> 200 MiB.
+    """
+    return to_json_safe(x, indent=None).decode("utf-8")
--- a/src/inspect_scout/_util/_json.py
+++ b/src/inspect_scout/_util/_json.py
@@ -3,13 +3,13 @@ from typing import Any
 from inspect_ai._util.json import to_json_safe
 
 
-def to_json_compact(x: Any) -> str:
+def to_json_str_compact(x: Any) -> str:
     """Serialize to JSON without pretty-print whitespace.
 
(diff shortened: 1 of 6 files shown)


src/inspect_scout/_util/_json.py
```diff
@@ -0,0 +1,16 @@
...
+    """Serialize to JSON without pretty-print whitespace.
+
+    Wraps ``to_json_safe`` (surrogate-safe) with ``indent=None``. Use for
+    machine-read payloads such as stored parquet columns and bytes streamed to
+    the viewer, where the default ``indent=2`` only bloats the representation
+    without benefit since the data is never human-consumed.
+
+    In extreme cases, this reduces serialized JSON from 700 MiB -> 200 MiB.
+    """
+    return to_json_safe(x, indent=None).decode("utf-8")
```
epatey: This is a bit of a nit, but since `to_json_safe` already has an `indent=` option, I'm not sure it's worth having a separate function.

It's also a little bit confusing to have another `to_json_xxx` function that returns a very different currency that `to_json_safe`. `to_json_safe` returns `bytes`, but this new one returns a `str`. It feels inconsistent.

I might either drop the new function altogether or rename it to something like `to_compact json_string`.

epatey: Actually, since there are so many call sites, I'd lean toward the rename and call it a day

epatey (approved):

src/inspect_scout/_util/_json.py
craig-meridian: Completely agree. Settled on renaming to `to_json_str_compact()` to mirror Inspect's `to_json_str_safe()` function.

PR #490, opened 2026-07-01
feat(OpenClaw): Add example transcript importer for telemetry generated by "telemetry-hal" plugin

craig-meridian:
A Scout transcript importer for OpenClaw transcripts which were recorded by the [telemetry-hal](https://github.com/sage-princeton/openclaw-telemetry-hal) plugin e.g. [CRUX1](https://cruxevals.com/crux-1/).

We opted to put this in the `examples/` dir rather than directly in the `inspect_scout` package as this is pretty much a one-off import from telemetry produced by a little-used plugin.

`inspect_ai>=0.3.245` to make use of the improved performance of `stable_message_ids()`. CRUX1 import would take hours without it.


examples/sources/openclaw_telemetry_hal/transcripts.py
```diff
@@ -0,0 +1,187 @@
...
+def _process_telemetry_file(telemetry_file: Path) -> "Transcript" | None:
+    """Process a single OpenClaw telemetry file into a transcript.
+
+    Args:
+        telemetry_file: Path to the telemetry file.
+
+    Returns:
+        A Transcript object, or None if no transcript could be created.
+    """
+    return _create_transcript(read_telemetry_events(telemetry_file), telemetry_file)
```
epatey, line 87: I know this is the same pattern used by the Claude Code importer, but this will block the event loop for the duration of the read. I suspect that will prevent things like Ctrl-C from working. It'd be worth confirming. Wrapping the call site in `anyio.to_thread.run_sync` might be worth it. It's up to you.

examples/sources/openclaw_telemetry_hal/events.py
```diff
@@ -0,0 +1,549 @@
...
+    ChatMessage,
+    ChatMessageAssistant,
+    ChatMessageTool,
+    ChatMessageUser,
+    Content,
+    ModelOutput,
+    ModelUsage,
+    StopReason,
+)
+from inspect_ai.model._generate_config import GenerateConfig
```
epatey: You could import this publicly by dropping the `._generate_config`. I wouldn't normally comment on something like this, but the README goes out of its way to talk about only using public stuff from inspect.

examples/sources/openclaw_telemetry_hal/tests/fixtures/sample-telemetry.jsonl
```diff
@@ -0,0 +1,57 @@
+{"type":"message.in","channel":"telegram","from":"telegram:5912046256","content":"hey can you run echo whoami in the terminal for me and give the result","contentLength":70,"timestamp":1782238951000,"metadata":{"to":"telegram:5912046256","provider":"telegram","surface":"telegram","originatingChannel":"telegram","originatingTo":"telegram:5912046256","messageId":"23","senderId":"5912046256","senderName":"Andrew Schwartz","senderUsername":"atschwartz"}}
```
epatey: It feels a little weird to have Andrew Schwartz's PII in the repo.

epatey (approved):
Approved with a few minor comments.

examples/sources/openclaw_telemetry_hal/transcripts.py
craig-meridian: Agreed. I did pin the event loop. Now offloading the read/parse to a worker thread via `anyio.to_thread.run_sync()` with `abandon_on_cancel=True`. Caveat: this makes cancellation respond promptly, but the in-flight file still runs to completion in the worker thread, so Ctrl-C on a large file is handled promptly, but the process still continues to run for some time. Documented this limitation at the call site.

examples/sources/openclaw_telemetry_hal/events.py
craig-meridian: Good point and completely agree. Now importing GenerateConfig from the public `inspect_ai.model`.

examples/sources/openclaw_telemetry_hal/tests/fixtures/sample-telemetry.jsonl
craig-meridian: Damn. I had made sure to scrub it from the crux1-sample-telemetry.json but forgot to do so in this file which he shared with me. Removed.

PR #491, opened 2026-07-03
perf: bounded-memory streaming transcript reads

rasmusfaber:
## Summary

Scanning or LLM-searching a large transcript materializes the whole filtered transcript as Pydantic objects before scanning starts: `TranscriptsView.read()` peaks at ~3× the content size, `llm_scanner` builds every segment and scans them with unbounded concurrency, and `ResultReport.input` re-embeds the full transcript into the results DB. Peak memory therefore scales with transcript size, and large logs can exhaust it.

This PR adds a streaming read path (`TranscriptHandle`, obtained via `open()`) that bounds peak memory to the in-flight working set regardless of transcript size. Samples over 64 MB are parsed in a single pass that spools filtered content and attachments to disk; iteration replays the spool one item at a time, resolving attachment/pool refs and validating each item on its own. Smaller samples keep the existing in-memory fast path unchanged.

Streaming engages when every scanner in a job is streaming-safe, and covers both messages-only scans and `llm_scanner` over events content. The latter via a two-pass "stub skeleton" that builds inspect_ai's timeline on lightweight event stubs, selects only the events the timeline actually reads, then re-streams to pull those full events. Segments scan through a bounded window (4 in flight). Scanner input is identical to the materialized path, asserted by equivalence tests across content filters, both backends, and a synthetic multi-agent fixture. In development measurements, peak memory was flat in transcript size: a ~50 MB events-heavy sample read messages at sub-1 MB peak, and a ~33 MB agentic sample scanned events at ~7 MB peak, unchanged as turn count doubled.

Also fixes a latent bug that predates this PR: an attachment ref reachable only through an `events_data` pool entry was never resolved on the materialized path. `attachments` precedes `events_data` in the sample, so the parser's "keep only already-referenced attachments" test could not yet know that a pooled message would need one, so it dropped precisely the attachments a pooled system prompt depends on. Streaming spools every attachment and so resolved these already, which is what surfaced the divergence: the same transcript yielded different scanner input depending only on whether it crossed the spool threshold. It bites hardest on an events-only scan, where messages are not parsed and nothing else registers the ref. On a real 100 MB agentic sample the materialized read left 24,290 unresolved `attachment://` refs in its events; it now leaves none, matching the streamed read.

## Recording the transcript

`ResultReport.input` must hold the full transcript so results stay readable without the original logs, and that value lands in a single parquet cell. It cannot be streamed, since pyarrow needs the complete value to build the array. Recording is therefore where the bound above stops applying, and on real agentic logs it was the dominant cost.

Rather than expanding the spool into `Transcript` objects and re-condensing them, the recorder copies the compact form the spool already holds: pool-condensed events with `attachment://` refs intact, pruned to the entries actually referenced. It assembles the column values without holding the transcript whole at any point:

- Each message and event is serialized as the spool replays it and written to a scratch spool, so no `json.dumps` ever runs over the envelope and the pieces never coexist with the result. The finished value is read back exactly once, which is what the cell requires and the floor this cannot go below.
- Sample metadata — most of the envelope on a metadata-dominated transcript — is copied straight from its own spool, never becoming objects or a `str`. Where it genuinely must become objects (`load()`), it is parsed from that spool's chunks rather than from one contiguous buffer.
- Column values are carried as UTF-8 bytes rather than `str`, since they go directly into a parquet column and decoding would double a non-ASCII value for nothing.
- `ItemSpool.items()` reassembles lines with a cursor instead of growing a buffer per 256 KB chunk, which previously made replay quadratic in item length: a 44 MB event cost ~3.8 GB of copying. This backs `messages()`/`events()` on the scan path too.

Peak RSS for the record path, on the largest sample of each of four real agentic logs, chosen to span the shapes that stress it differently. `before` is `origin/main`; both columns come from one harness on a 247 GB host, so neither path was constrained or swapping. The two starred rows produce a cell parquet cannot store at all (see the 2 GiB ceiling below) -- their numbers are the cost of *producing* the value, not of a scan that can currently succeed:

| sample shape | sample | before | after |
|---|---|---|---|
| metadata-dominated (95% of the envelope) | 3,223 MB | 14,688 MB | 1,750 MB |
| event-heavy (~1M events) \* | 2,239 MB | 15,627 MB | 2,470 MB |
| one 44 MB event among 40 | 132 MB | 789 MB | 505 MB |
| largest sample measured \* | 12,118 MB | 78,279 MB | 6,492 MB |

Recorded output is the same size on the event-heavy and single-large-event samples, and smaller where an attachment is stored once in `input_data` instead of being inlined at every use: 3,212 MB → 1,485 MB on the metadata-dominated sample, 12,096 MB → 6,235 MB on the 12.1 GB one. Post-parse RSS on the metadata-heavy sample drops 3,232 MB → 422 MB, which is the scan-path half of the metadata change.

Recorded output is semantically equivalent to the materialized path throughout — see "What equivalence means" below — and each individual change is byte-identical to the implementation it replaced, verified by hashing both columns on real logs.

## Caveats & non-goals

- **A cell above ~2 GiB cannot be stored at all.** Parquet caps a single value just under 2 GiB, and the failure is not graceful: 2,147,483,646 bytes gives `ArrowInvalid: Negative buffer resize: -2147483646`, 2^31-1 gives `ArrowCapacityError`, and 2^31 gives `ArrowInvalid: Parquet cannot store strings with size 2GB or more`. Two of the four samples above exceed it (2,239 MB and 6,235 MB). This is not introduced here -- `main` produces the same content, slightly larger (2,246 MB and 12,096 MB) -- but it means recording a multi-GB transcript inline is impossible rather than merely expensive, and that storing the transcript by reference is a correctness fix, not an optimisation. The spool knows the finished size before it is materialized (`ByteSpool.__len__`), so the check can be made without building the value.
- Below that ceiling, recording still peaks at roughly the size of the cell (~1.5 GB for the 3.2 GB sample). That is the floor's shape: the value is built on the spool, then read back once because a parquet cell is a single value.
- `StreamParseResult.metadata` is now a method that reparses from the spool on each call, deliberately uncached: caching it would reinstate the retention the spool removes. `handle.load()` calls it, since materializing is the one path that genuinely needs the objects.
- Parquet-backed transcript stores are out of scope: this bound applies to the eval-log path only. A streaming `open()` for parquet was dropped from this PR because DuckDB returns each content cell as one whole Python string before anything can stream it, so the win was a constant factor on a path production does not use. Fixing the parquet fetch properly is meridianlabs-ai/inspect_scout#571, which targets `main` directly.
- Fallback: transcripts whose events predate inspect_ai's event `uuid` field (before July 2025) can't be targeted for substitution and transparently fall back to a materialized scan (logged at info level).
- Legacy transcripts: the streaming path hydrates nested `ToolEvent.events` into typed events, where the materialized path leaves them as raw dicts — so results can differ on such logs (the streaming path is the faithful one).
- Still materialized: timeline-content scans (`timeline=`) and the small-file read path.
- The bound assumes a disk-backed temp dir. With no file cache configured, spools go to `tempfile.gettempdir()`, so a host that mounts `/tmp` as tmpfs holds them in RAM and the bound stops holding. The figures above set `TMPDIR` to disk.
- `TranscriptHandle` is intentionally not exported for now
- The recorded `input` keeps `attachment://` refs unresolved, with the lookup table nested in `input_data`. `scan_results_df()` resolves them on read; `scan_results_db()` and `scan_results_arrow()` return the columns verbatim, so a caller using those must resolve against `input_data["attachments"]` itself. The viewer needed the same fix on its scan-results path: **meridianlabs-ai/ts-mono#507**, carried here by the submodule bump to `faber/scout-results-attachment-refs`. **That PR must merge first**. Without it the viewer renders `attachment://<hash>` as literal text for every streamed scan result. Once it lands, the submodule pointer should move from the feature branch to the resulting commit on ts-mono `main`.


epatey:
@rasmusfaber, let me know if you want me to start reviewing this while it's still a draft.

epatey:
@rasmusfaber, my flow pretty much ignores draft PRs. Sorry if this was ready for me. Can you either merge it up and mark it ready or close it.

rasmusfaber:
> @rasmusfaber, my flow pretty much ignores draft PRs. Sorry if this was ready for me. Can you either merge it up and mark it ready or close it.

Yeah, this have been hanging in draft for a while, while I tried finding the time to pare it down or split it out. For what it is worth, we have been running this in METR Hawk for some weeks now, and it seem to be working as it should. I will move it into review, and then you can let me know whether you want me to split it up.

epatey:
Thanks for this, and for running it in Hawk for weeks before sending it over. The equivalence tests are more than most PRs this size come with.

I'll take you up on your offer to split it. At +7.7k across 45 files this is really a few independent changes riding together (the streaming read path, the record-by-reference recorder change, and the pooled attachment-ref fix), and splitting them lets each get a proper review and land on its own instead of gating everything on the whole. It also makes the sequencing explicit: the viewer side depends on ts-mono#507 landing first and the submodule pointer moving, so the streamed-results piece is cleaner once that ordering is in place.

I'm going to close this one in favor of the split series so the queue tracks the pieces. Nothing here is lost. Please open them as separate PRs in whatever order makes sense and link them back here, and I'll pick them up. Appreciate the care you put into this.

PR #498, opened 2026-07-06
Fix MyPy for `inspect_ai@main`: `ConcurrencySemaphore` protocol

craig-meridian:
## Problem

[UKGovernmentBEIS/inspect_ai#4410](https://github.com/UKGovernmentBEIS/inspect_ai/pull/4410) (merged 2026-07-06 14:11 UTC) changed the `ConcurrencySemaphore` protocol:

- `concurrency` is now a read-only property (was a plain mutable attribute)
- a new required `in_use` property (exact count of current holders)
- `ConcurrencySemaphoreRegistry.get_or_create` gained a `resizable: bool = False` parameter

CI in this repo now [fails](https://github.com/meridianlabs-ai/inspect_scout/actions/runs/28803473871/job/85412748559) for MyPy.

Note that we use `inspect_ai` from `main` in CI (but pyproject.toml installs from PyPI).

## Fix

- `MPConcurrencySemaphore` now satisfies the protocol **structurally** instead of nominally inheriting it. The runtime failure was caused by inheriting the protocol's new property descriptors, which block plain attribute assignment in `__init__`; the upstream protocol docstring explicitly notes a plain attribute satisfies the read-only `concurrency` member. The registries' `ConcurrencySemaphore` return types keep the structural conformance checked by mypy.
- Added the `in_use` property (`concurrency - value`, correct for a fixed-limit semaphore per the upstream docstring).
- Both registries accept (and ignore) `resizable`, matching the existing treatment of `adaptive` — live resizing isn't supported cross-process. A warning is emitted if either of these is set.

## Verification

- mypy passes against **both** inspect_ai 0.3.244 (the declared minimum pin) and inspect_ai git main (verified locally by overlaying the new `_concurrency.py` for type-checking; the 5 CI errors reproduced exactly before the fix)

### DIFF
--- a/src/inspect_scout/_concurrency/_mp_registry.py
+++ b/src/inspect_scout/_concurrency/_mp_registry.py
@@ -9,6 +9,7 @@ from __future__ import annotations
 
 import sys
 from collections.abc import Iterable
+from logging import getLogger
 from multiprocessing.managers import DictProxy, SyncManager
 from multiprocessing.queues import Queue
 from threading import Condition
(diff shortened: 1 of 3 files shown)


craig-meridian:
FYI @epatey, @ransomr - thought this was trivial enough to go ahead and merge after reviews from Claude Code & Codex, but thought I'd tag you both just for awareness (Eric, you've done a lot of work in this file, Ransom, you'll know a lot about the concurrency limit support in Inspect).

ransomr:
> FYI @epatey, @ransomr - thought this was trivial enough to go ahead and merge after reviews from Claude Code & Codex, but thought I'd tag you both just for awareness (Eric, you've done a lot of work in this file, Ransom, you'll know a lot about the concurrency limit support in Inspect).

Thanks - LGTM (and I added the new args, so thanks for fixing).

PR #499, opened 2026-07-08
feat(OpenClaw): Native transcript importer

craig-meridian:
A Scout transcript importer for native OpenClaw session files.

Targets `main` directly and is independent of #490 (the `openclaw-telemetry-hal` importer, now relocated to `examples/`).

In addition to automated tests, manually tested with:
- real OpenClaw session files I generated
- some open-source ones available online.


craig-meridian:
This work was motivated when we added support for importing OpenClaw transcripts which were produced using the "telemetry-hal" plugin. We didn't want to push users toward using this (relatively unknown) plugin if they needed to import OpenClaw transcripts. However, we are not aware of any users wishing to import OpenClaw transcripts other than the CRUX evals. If this changes, we'd be happy to revive this PR.

Issue #496, opened 2026-07-06
Request: periodically sync in-progress scans to final scan location


nix-aisi:
My understanding is there's no equivalent of inspect-ai's `log_buffer` option, where you can write results to the final location every N samples.

I think this is a generally helpful feature for being able to run analysis and sanity check results as the scan is in-flight.

nix-aisi:
Closed by #550 , thanks!

PR #500, opened 2026-07-09
Results buffer periodic sync fixing #496

nix-aisi:
Solution is cluade-coded but looks reasonable (aside from docs should be a bit more concise IMO).

I haven't tested yet, not sure when I'll get the chance to pick this up so someone else is welcome to take over.

### DIFF
--- a/docs/workflow.qmd
+++ b/docs/workflow.qmd
@@ -289,7 +289,7 @@ For some scanning work you won't get any benefit from increasing max processes (
 
 ### Batch Mode
 
-Inspect AI supports calling the batch processing APIs for the [OpenAI](https://platform.openai.com/docs/guides/batch), [Anthropic](https://platform.claude.com/docs/en/build-with-claude/batch-processing), [Google](https://ai.google.dev/gemini-api/docs/batch-api?batch=file), and [Together AI](https://docs.together.ai/docs/batch-inference) providers. Batch processing has lower token costs (typically 50% of normal costs) and higher rate limits, but also substantially longer processing times—batched generations typically complete within an hour but can take much longer (up to 24 hours).
+Inspect AI supports calling the batch processing APIs for the [OpenAI](https://platform.openai.com/docs/guides/batch), [Anthropic](https://platform.claude.com/docs/en/build-with-claude/batch-processing), [Google](https://ai.google.dev/gemini-api/docs/batch-api?batch=file), and [Together AI](https://docs.together.ai/docs/batch-inference) providers. Batch processing has lower token costs (typically 50% of normal costs) and higher rate limits, but also substantially longer processing times—batched generations typically complete within an hour but can take much longer (up to 24 hours).
 
 Use batch processing by passing the `--batch` CLI argument or the `batch` option from `GenerateConfig`. For example:
 
@@ -304,6 +304,18 @@ If you don't require immediate results then batch processing can be an excellent
 -   The optimal processing flow for batch mode is to send *all of your transcripts* in a single batch group so that they all complete together. Therefore, when running in batch mode `--max-transcripts` is automatically set to a very high value (10,000). You may need to lower this if holding that many transcripts in memory is problematic.
 
 See the Inspect AI documentation on [Batch Mode](https://inspect.aisi.org.uk/models-batch.html) for additional details on batching as well as notes on provider specific behavior and configuration.
+
+### Monitoring In-Progress Results
+
+For long-running scans it can be useful to inspect partial results while the scan is still in flight—for example to sanity check that a scanner is behaving as expected before committing to a multi-hour run. By default results are only written to the scan location when the scan completes (or is interrupted).
+
+Use the `--results-buffer` option (or the `results_buffer` scan/scan-job/project option) to periodically sync in-progress results to the scan location every N recorded results:
+
+``` bash
+scout scan cybench_scan.py --results-buffer 100
+```
+
+With this set, the compacted scanner Parquet files at the scan location are refreshed every 100 results, so you can point `scan_results_df()` (or Scout View) at the scan directory and analyze what's been produced so far. Writes are atomic, so readers always observe a consistent snapshot. Periodic syncing adds some I/O overhead, so choose N to balance freshness against cost (larger values sync less often); leaving it unset disables periodic syncing entirely.
 
 ### Error Handling
 
--- a/src/inspect_scout/_cli/scan.py
+++ b/src/inspect_scout/_cli/scan.py
@@ -254,6 +254,13 @@ shuffle_option = click.option(
     envvar=["SCOUT_SCAN_SHUFFLE"],
 )
 
+results_buffer_option = click.option(
+    "--results-buffer",
+    type=int,
+    help="Sync in-progress results to the scan location every N recorded results, so partial results can be inspected while the scan runs (defaults to no periodic sync).",
+    envvar="SCOUT_SCAN_RESULTS_BUFFER",
+)
+
 tags_option = click.option(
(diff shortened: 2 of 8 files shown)


dragonstyle:
I asked claude to test out the perf against S3 and here was my local measurement FWIW.

248 transcripts, non-LLM scanner, transcripts always read from the local corpus; only the scans output location and buffer setting vary. Median wall-clock:

```
┌───────────────────────────┬─────────────┬──────────┐
│         Condition         │ Local scans │ S3 scans │
├───────────────────────────┼─────────────┼──────────┤
│ Baseline (no PR)          │ 3.89s       │ 8.78s    │
├───────────────────────────┼─────────────┼──────────┤
│ PR, buffer off (default)  │ 3.89s       │ 8.69s    │
├───────────────────────────┼─────────────┼──────────┤
│ PR, buffer=25 (~10 syncs) │ 4.32s       │ 28.4s    │
├───────────────────────────┼─────────────┼──────────┤
│ PR, buffer=10 (~25 syncs) │ 5.97s       │ 60.2s    │
└───────────────────────────┴─────────────┴──────────┘
```

S3 is still very impactful it seems (even with async writing). We might need a large default buffer size for S3...

epatey:
I dug into whether "S3 writes are async now" actually holds for the periodic sync path this PR uses. Short version: **it's only partially true, which is why @dragonstyle's benchmark still shows large slowdowns — but the approach is fixable without resorting to a huge default buffer size.**

### What's actually async today

The compacted parquet upload in `FileRecorder.sync()` goes through `AsyncFilesystem.write_file()`, which is genuinely async on S3 (aioboto3 under asyncio). That part of the original problem is solved.

### What still blocks

Everything else in the `sync()` path is synchronous S3 I/O that blocks the event loop, and this PR additionally `await`s the whole sync inline inside `record_results`, so each sync's full latency lands on the scan's critical path (in multi-process mode it stalls the parent's result-drain loop). Per `sync()` call:

1. `_compact_with_prior()` calls `prior.exists()` (sync HEAD) and `prior.read_bytes()` — a **synchronous download of the entire previously-compacted parquet**, on every sync. Since each sync's output becomes the next sync's "prior", the per-sync download grows as the scan progresses (O(N²) total transfer over the run).
2. `sync_fs.mv(tmp, output)` — synchronous fsspec; on S3 that's server-side COPY + DELETE round trips, per scanner, per sync.
3. `_sync_status_files()` — two more synchronous S3 writes (`_summary.json`, `_errors.jsonl`).
4. `scanner_table()` compaction is CPU-bound and runs directly on the event loop.

That's consistent with the measurements above (~2s of overhead per sync on S3, fully serialized into wall-clock).

### Conclusion

The reason the approach was originally abandoned (sync S3 writes stalling the scan) is *not* automatically gone — async writes exist but the periodic-sync path doesn't benefit from them as wired up here. However, the feature is viable if the sync is taken off the critical path and made async end-to-end:

- run the periodic sync as a **background task** (skip if one is already in flight) instead of awaiting it in `record_results`;
- use `AsyncFilesystem` for the remote ops (exists/read/write);
- for remote scan dirs, PUT directly to the final key (object-store PUTs are atomic — readers never see partial objects) instead of write-tmp + `mv`;
- download the prior compacted parquet **once per run** rather than on every sync;
- run the pyarrow compaction in a worker thread.

With that, the steady-state per-sync cost off the critical path is one local compaction (in a thread) + one async upload + two small async writes — wall-clock impact should be ≈0 as long as a sync completes within the cadence, and the skip-if-in-flight guard degrades gracefully when it doesn't.

I'm adapting the PR along these lines on `claude/async-s3-scan-buffering-wi2rll` and will follow up here.

---

epatey:
Follow-up: I've adapted this PR on branch [`claude/async-s3-scan-buffering-wi2rll`](https://github.com/meridianlabs-ai/inspect_scout/tree/claude/async-s3-scan-buffering-wi2rll) with the restructuring described above. Summary of what changed relative to this PR:

**Keeps sync off the scan's critical path**
- Periodic syncs run as **background tasks** hosted in a task group scoped to the strategy run (skip if one is in flight; leaving the scope awaits any in-flight sync so it can't race the final sync). `record_results` no longer awaits the sync.

**Makes `FileRecorder.sync()` event-loop friendly** (benefits the final/interrupt syncs too)
- Prior-parquet `exists()`/download go through `AsyncFilesystem` instead of blocking UPath/fsspec calls.
- The prior compacted parquet is downloaded **at most once per run** (module-level cache keyed by buffer dir, cleared on `init`/`resume`/`attach` and on completing sync). Previously every sync re-downloaded the ever-growing output it wrote itself — O(N²) transfer over a run.
- Remote scan dirs get a **direct atomic PUT** to the final key (object stores never expose partial objects) instead of write-tmp + `mv`, which cost extra COPY + DELETE round trips per scanner per sync. Local dirs keep tmp + rename.
- pyarrow compaction runs in a worker thread.
- Status files (`_summary.json`, `_errors.jsonl`) are written through the async filesystem; buffer summary writes are now tmp + rename so a mid-scan sync never reads a truncated file.

**Tests/docs**
- Fixed a bug in this PR's probe test: it looked for the compacted parquet at the scans root, but syncs write into `scan_id=*/`, so the "partial results visible" assertion could never observe anything. The rewritten tests wait deterministically for the background sync at a known invocation, and cover single/multi-process plus the no-buffer default.
- Verified the S3 code path end-to-end against a moto S3 server, including the eval_set-style `attach` lifecycle (prior merged, no duplicate rows, single prior download, cache cleared on completion).
- Trimmed the docs section per the PR description's own note.

**Perf.** Same shape as the earlier benchmark (200 transcripts, trivial non-LLM scanner, `max_processes=1`, median of 3), against a local moto S3 server:

| Condition | S3 scans |
|---|---|
| buffer off | 7.98s |
| buffer=25 (~8 syncs) | 7.89s |
| buffer=10 (~20 syncs) | 7.42s |

No measurable overhead — syncs now overlap scanning instead of serializing into it (the inline-await version measured +2s wall-clock *per sync* on real S3). moto latencies are lower than real S3, but since syncs are off the critical path, higher latency stretches individual syncs rather than the scan; the skip-if-in-flight guard degrades to fewer, coarser syncs if they can't keep up with the cadence. A worst-case caveat: with a trivial scanner and a *local* scans dir, very frequent syncing shows modest overhead from compaction competing for CPU — invisible for real (LLM) scans.

**One loose end for whoever merges:** `results_buffer` is persisted on `ScanOptions`/`ScanJobConfig` (so `scan resume` honors it), which changes the OpenAPI schema. `openapi.json` is regenerated on the branch, but the regenerated `generated.ts` lives in the ts-mono submodule, which I can't push to — a companion ts-mono commit + submodule bump is needed or the type-sync CI check will fail. The diff is just the optional field in five schemas:

```diff
+            /** Results Buffer */
+            results_buffer?: number | null;
```

(produced by `.venv/bin/python scripts/export_openapi_schema.py` with the submodule checked out)

---

epatey:
Closing this in favor of #550, which carries this PR's `results_buffer` feature forward with the restructuring discussed above (background syncs off the record path, fully-async S3 I/O in the sync path, once-per-run prior download, threaded compaction).

@dragonstyle — when you get a chance, could you rerun your S3 perf measurement from [this comment](https://github.com/meridianlabs-ai/inspect_scout/pull/500#issuecomment-5135874808) against the #550 branch (`claude/async-s3-scan-buffering-wi2rll`)? Against a local moto S3 server I measured no wall-clock overhead with `--results-buffer 10`/`25` vs. buffer off, but a confirmation against real S3 would validate the fix under real latency.

---

PR #501, opened 2026-07-09
fix(transcripts): add SSO provider to DuckDB S3 credential chain for parquet transcript writes

sjawhar:
## Problem

`ParquetTranscriptDatabase._init_s3_auth` creates the DuckDB S3 secret with `PROVIDER credential_chain` and no explicit chain. DuckDB's default credential chain does not include the `sso` provider, so a session authenticated only through AWS SSO (no static env credentials) cannot write parquet transcripts to S3. The current workaround is to export static credentials manually before every run.

## Fix

When an AWS config file is present, bind its profile and add `sso` to the chain:

- `CHAIN 'env;config;sso;instance;process'` keeps the existing static-credential and instance-role paths working and adds SSO.
- `PROFILE '<AWS_PROFILE or default>'` so the SSO session resolves against the intended profile. The clause is required for SSO: DuckDB's `sso` chain provider resolves its profile only from a bound `PROFILE`, not from `AWS_PROFILE` in the environment.
- `sts` is omitted because it errors without an `ASSUME_ROLE_ARN`.

Because DuckDB validates a bound `PROFILE` against the config file eagerly at `CREATE SECRET`, the profile-bound secret is wrapped in try/except: on DuckDB's `Secret Validation Failure: no profile ... found in config file` rejection it falls back to the plain default `credential_chain` (preserving named-profile-only and credentials-file-only setups, per review). Any other DuckDB error re-raises.

When no config file exists (for example, bare env-credential containers), the original `PROVIDER credential_chain` is used directly, so those environments are unaffected.

## Testing

Unit tests cover the five scenarios from review: the named-profile-only regression (falls back, env creds resolve), `AWS_PROFILE` defined only in `~/.aws/credentials` (falls back, default chain honors it), explicit SSO profile binding (profile-bound chain consults the correct token cache), no config file (plain chain), and propagation of unrelated DuckDB errors.

Verified end to end: publishing parquet transcripts to a real S3 bucket succeeds under a pure AWS SSO session with no manual credential export, and continues to work with static env credentials. Full serial suite with Inspect 0.3.249-compatible SDKs: 3159 passed, 111 skipped.

### DIFF
--- a/src/inspect_scout/_transcript/database/parquet/transcripts.py
+++ b/src/inspect_scout/_transcript/database/parquet/transcripts.py
@@ -2178,12 +2178,43 @@ class ParquetTranscriptsDB(TranscriptsDB):
 
     def _init_s3_auth(self) -> None:
         assert self._conn is not None
-        self._conn.execute("""
-            CREATE SECRET (
-                TYPE S3,
-                PROVIDER credential_chain
-            )
-        """)
+        # DuckDB's default credential_chain omits the SSO provider, so an SSO-only
+        # session (no static env creds) can't authenticate to S3. When an AWS config
+        # file is present, add `sso` to the chain and bind the profile explicitly:
+        # the sso provider resolves its profile only from a bound PROFILE clause
+        # (it ignores AWS_PROFILE in the environment), so SSO doesn't work without
+        # one. env/config/instance/process keep static-cred and instance-role paths
+        # working; `sts` is omitted -- it errors without an ASSUME_ROLE_ARN.
+        #
+        # DuckDB validates a bound PROFILE against the config file eagerly at
+        # CREATE SECRET and throws if the profile isn't defined there (e.g. no
+        # [default] section, or the profile lives only in the credentials file) --
+        # even when other chain providers could supply credentials. Fall back only
+        # for that eager profile-validation rejection; all other DuckDB errors
+        # must propagate.
+        config_file = os.environ.get("AWS_CONFIG_FILE") or os.path.expanduser(
+            "~/.aws/config"
+        )
+        if os.path.isfile(config_file):
+            profile = (os.environ.get("AWS_PROFILE") or "default").replace("'", "''")
+            try:
+                self._conn.execute(
+                    "CREATE SECRET (TYPE S3, PROVIDER credential_chain, "
+                    f"CHAIN 'env;config;sso;instance;process', PROFILE '{profile}')"
+                )
+                return
+            except duckdb.Error as ex:
+                error_message = str(ex)
+                if not (
+                    error_message.startswith("Secret Validation Failure: no profile ")
+                    and error_message.endswith(" found in config file")
+                ):
+                    raise
+                logger.debug(
+                    f"S3 secret with profile '{profile}' failed ({ex}); "
+                    "falling back to default credential chain."
+                )
+        self._conn.execute("CREATE SECRET (TYPE S3, PROVIDER credential_chain)")
 
     def _is_hf(self) -> bool:
         return self._location is not None and self._location.startswith("hf://")
--- a/tests/transcript/database/test_s3_auth.py
+++ b/tests/transcript/database/test_s3_auth.py
@@ -1,0 +1,125 @@
+from pathlib import Path
+from typing import cast
+
+import duckdb
+import pytest
+from inspect_scout._transcript.database.parquet import ParquetTranscriptsDB
+
+
+class RecordingConnection:
+    def __init__(self, outcomes: list[duckdb.Error | None]) -> None:
+        self._outcomes = outcomes
+        self.statements: list[str] = []
+
+    def execute(self, statement: str) -> None:
+        self.statements.append(statement)
+        outcome = self._outcomes.pop(0)
+        if outcome is not None:
+            raise outcome
+
+
+def _s3_database(connection: RecordingConnection) -> ParquetTranscriptsDB:
+    database = ParquetTranscriptsDB("s3://example-bucket/transcripts")
+    database._conn = cast(duckdb.DuckDBPyConnection, connection)
+    return database
+
+
+def _write_config(tmp_path: Path, contents: str) -> Path:
+    config_file = tmp_path / "config"
+    config_file.write_text(contents, encoding="utf-8")
+    return config_file
+
+
+def test_s3_auth_falls_back_when_config_has_only_named_profile(
+    tmp_path: Path, monkeypatch: pytest.MonkeyPatch
+) -> None:
+    config_file = _write_config(tmp_path, "[profile named-only]\nregion = us-east-1\n")
+    monkeypatch.setenv("AWS_CONFIG_FILE", str(config_file))
+    monkeypatch.delenv("AWS_PROFILE", raising=False)
+    connection = RecordingConnection(
+        [
+            duckdb.Error(
+                "Secret Validation Failure: no profile 'default' found in config file"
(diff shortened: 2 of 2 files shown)


dragonstyle (commented):
Thanks for this — adding `sso` to the chain is the right call, but testing against DuckDB 1.5.4 (what `duckdb>=1.4.0` currently resolves to) turned up a regression in the unconditional `PROFILE` binding, plus a finding that shapes the fix.

**Regression: eager `PROFILE` validation breaks named-profile-only setups.** DuckDB validates a bound `PROFILE` against the AWS config file eagerly at `CREATE SECRET`, before any chain provider runs. If `~/.aws/config` exists but has no `[default]` section and `AWS_PROFILE` is unset, this PR's secret throws `Secret Validation Failure: no profile 'default' found in config file` — even with valid env credentials (`env` being first in the chain doesn't help), and even if `~/.aws/credentials` has a `[default]` (only the config file is checked). A/B in an identical controlled environment: the pre-PR plain `PROVIDER credential_chain` secret succeeds and resolves the env creds; this PR's secret fails, so S3 transcript reads/writes break for anyone who only uses named profiles. The same failure hits `AWS_PROFILE` pointing at a profile defined only in the credentials file.

**But the `PROFILE` clause can't simply be dropped.** We traced file access with a fabricated SSO token cache: DuckDB's `sso` chain provider resolves its profile *only* from a bound `PROFILE` clause. It ignores `AWS_PROFILE` in the environment — without a clause it looks up the token cache for a SHA-1 of an *empty* `sso_start_url` (`da39a3ee...`), even when the config file has a literal `[default]` SSO profile. With `PROFILE` bound it opens the correct `~/.aws/sso/cache/<sha1(start_url)>.json` and proceeds to `GetRoleCredentials`. So conditionally emitting `PROFILE` only when `AWS_PROFILE` is set would silently lose SSO support for default-profile users — the clause has to stay for SSO to work at all.

**Fix (reworked on [`claude/duckdb-s3-sso-profile-fix-cggzbu`](https://github.com/meridianlabs-ai/inspect_scout/compare/main...claude/duckdb-s3-sso-profile-fix-cggzbu), commit ab10011):** keep this PR's chain and profile binding, but wrap the profile-bound `CREATE SECRET` in try/except and fall back to the plain default `credential_chain` when the eager validation rejects the profile. Verified in controlled environments on 1.5.4:

- named-profile-only config + env creds (the regression): falls back, resolves env creds ✔️
- `AWS_PROFILE` defined only in `~/.aws/credentials`: falls back, and the default chain resolves that profile (the non-sso providers *do* honor env `AWS_PROFILE`) ✔️
- SSO profile (explicit or `default`): profile-bound chain consults the correct token cache ✔️
- no config file: unchanged plain `credential_chain` ✔️

One pre-existing behavior worth knowing (unchanged by either version): DuckDB resolves `credential_chain` secrets eagerly and `CREATE SECRET` throws if the chain resolves no credentials at all, so a fully credential-less environment fails at connect either way — the fallback keeps that at parity with `main`.

---

dragonstyle (changes requested):

sjawhar:
Applied your fix from `ab10011`. The profile-bound `CREATE SECRET` is now wrapped in `try/except`, falling back to the plain default `credential_chain` when DuckDB rejects the eager `PROFILE` validation. That keeps SSO working (it resolves its profile only from the bound `PROFILE` clause) while no longer breaking named-profile-only or credential-file-only setups. Pushed as `c435ad0`; re-requesting review.

(Posted by Claude on Sami's behalf.)

sjawhar:
Claude on Sami's behalf: fixed the failing lint check by changing the PR title to the required conventional-commit form. Local validation against the workflow's allowed types passed, and the rerun is green: https://github.com/meridianlabs-ai/inspect_scout/actions/runs/30024080282/job/89263937347\n\nThe profile-validation fallback requested in review is already present in c435ad0d.

sjawhar:
Claude on Sami's behalf: Thanks again for the A/B analysis and for publishing `claude/duckdb-s3-sso-profile-fix-cggzbu@ab10011`. I adopted that profile-bound SSO chain plus default-chain fallback, then narrowed the fallback to DuckDB's `Secret Validation Failure: no profile ... found in config file` rejection. Other DuckDB errors now re-raise.\n\nAdded coverage for the named-profile-only regression, `AWS_PROFILE` defined only in credentials, explicit SSO profile binding, no config file, and propagation of unrelated DuckDB errors. Focused tests: 5 passed. Full serial suite with Inspect 0.3.249-compatible SDKs: 3159 passed, 111 skipped.

dragonstyle:
Before merging I tried to verify the core premise end to end, and I think we need to re-scope the problem first: **"DuckDB's default credential chain does not include the sso provider" doesn't hold for current versions.** When no `CHAIN` is specified, duckdb-aws constructs `Aws::Auth::DefaultAWSCredentialsProviderChain` directly ([aws_secret.cpp](https://github.com/duckdb/duckdb-aws/blob/main/src/aws_secret.cpp)), and the SDK's default chain includes `SSOCredentialsProvider` (and honors `AWS_PROFILE`) ([AWSCredentialsProviderChain.cpp](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/source/auth/AWSCredentialsProviderChain.cpp)). Confirmed live: on unpatched main (macOS arm64, duckdb 1.4.4, `sso-session`-style profile, no static creds), parquet transcript publish to S3 succeeds under pure SSO, fails after `aws sso logout`, and succeeds again after login — so the default chain is genuinely consuming the SSO token cache.

Your failure was clearly real, so the interesting question is what's different in the failing environment. Leading suspects: an older duckdb/aws-extension build whose bundled aws-sdk-cpp predates `sso-session`-format support (legacy `sso_start_url`-in-profile worked in the SDK long before session-style), or a pinned/stale extension in a container. Could you share from the failing environment:

- [ ] OS/arch, duckdb version, and `SELECT extension_name, extension_version FROM duckdb_extensions() WHERE extension_name IN ('httpfs','aws')`
- [ ] Profile format: legacy (`sso_start_url` in the profile) vs `sso-session` reference, and whether `AWS_PROFILE` was set
- [ ] The exact error text from a failing publish on unpatched main
- [ ] Ideally a bare-duckdb repro: `CREATE SECRET (TYPE S3, PROVIDER credential_chain)` + `read_parquet('s3://…')` under the same session

Once we know which builds/formats fail, the fix shape follows — quite possibly this same chain but gated to environments/profiles that need it, which would also let the eager-validation fallback disappear. Worth knowing: switching to an explicit `CHAIN` steps off the SDK default chain entirely (that's *why* `PROFILE` binding became necessary — the SDK default honors `AWS_PROFILE`, the explicit chain doesn't), and it drops providers the default includes (web-identity, and the `sts`/role paths), so unconditional replacement risks regressing setups that work today.

sjawhar:
Claude here, on Sami Jawhar's behalf. I re-tested on the environment that originally failed, and you're right.

Linux x86_64, duckdb 1.5.4, httpfs `c3f215a`, `[sso-session]`-format profile, `AWS_PROFILE` unset, no static credentials in the environment. A bare-duckdb `CREATE SECRET (TYPE s3, PROVIDER credential_chain)` with the default chain both read and wrote parquet on S3 under pure SSO - no 403. The original July failure's exact traceback wasn't preserved, and it doesn't reproduce on current versions, so I can't justify the unconditional chain replacement. Closing. Sorry for the round trip, and thanks for checking the premise rather than the diff.

PR #517, opened 2026-07-14
fix: support legacy JSON logs in EvalLogTranscriptsView.read()

### LINKED ISSUE #250: `EvalLogTranscriptReader.read()` doesn't support legacy JSON format
**Location:** [eval_log.py:429-430](src/inspect_scout/_transcript/eval_log.py#L429-L430)

**Current behavior:**
```python
if recorder_type_for_location(t.source_uri) is not EvalRecorder:
    raise NotImplementedError("JSON format not yet supported")
```

**Problem:** `EvalLogTranscriptReader.read()` only supports `.eval` zip format. Legacy `.json` log files raise `NotImplementedError`.

**Impact:** Users with older JSON-format transcripts can't read them via this code path.

## Implementation Notes

- Must support message/event filtering via `load_filtered_transcript`
- Source sample from JSON via `recorder_type.read_log_sample()` (see [eval_log.py:478](src/inspect_scout/_transcript/eval_log.py#L478))
- Pass resulting `sample_bytes` to `load_filtered_transcript`

rinaldofesta:
## Summary

Closes #250.

`EvalLogTranscriptsView.read()` raises `NotImplementedError("JSON format not yet supported")` for legacy `.json` logs, while `read_messages_events()` in the same class already handles them by sourcing the sample through `recorder_type.read_log_sample()`.

## Change

- `read()` now handles the JSON case with the same path `read_messages_events()` uses: `read_log_sample()` for the sample, `model_dump_json()` for the bytes, then `load_filtered_transcript()` for message/event filtering.
- `max_bytes` is enforced on the serialized sample size. The semantics differ from the `.eval` path, which checks the ZIP entry's uncompressed size before reading anything: for a JSON log the sample has to be read before it can be measured, so the check guards the parse, not the fetch.
- The post-processing both formats need (timeline fallback build, plus re-filtering events after loading "all" for timeline resolution) moved unchanged into a shared helper, `_resolve_timelines_and_filter_events()`.
- Tests: new `tests/transcript/test_eval_log_read.py` with three cases: a `.json` read returns the same messages/events as the `.eval` read of the same sample; timeline requests build timelines and re-filter events identically across formats; an over-limit `max_bytes` raises `TranscriptTooLargeError`. The JSON fixture is generated in the test by round-tripping the existing `.eval` fixture through `read_eval_log`/`write_eval_log`.

Scope I left out on purpose: the parquet reader carries the same timeline-fallback block (`database/parquet/transcripts.py`). I kept this PR to the eval-log reader; happy to unify the two in a follow-up if you want it shared.

## Test

`pytest tests/transcript/`: 579 passed. Full suite: 3026 passed, 3 failures in `tests/scanner/test_validation.py` that reproduce on current main with this change stashed, so pre-existing and untouched here. `ruff check`, `ruff format --check`, and `mypy src examples tests` clean.

---
*Implemented with the help of a coding agent (Claude Code) and reviewed locally by the author.*


epatey (approved):
Thank you for this. I'm approving!

I have two small asks:
- a one-line code comment noting that `max_bytes` on the JSON path guards the parse rather than the fetch — your description explains it well, the code should too
- double-check the timeline test actually exercises the `timeline_build()` fallback — if the fixture stores timelines, the round-trip keeps them and the fallback never runs; stripping timelines before writing the JSON fixture would cover it.

Worth noting for the record, no change requested: JSON logs re-read the whole file per sample, so a full scan is O(N²) in file reads. That's pre-existing (`read_messages_events()` behaves the same) and staying consistent with it here was the right call since this is the legacy JSON path.

rinaldofesta:
Both addressed in 20e2cfb3.

Checked the fallback first: the fixture sample stores `timelines: None`, so `timeline_build()` already ran on both paths (instrumented it: 1 call each). Stripped timelines in the JSON fixture anyway so the coverage survives a regenerated fixture that stores them.

PR #523, opened 2026-07-14
fix: union type validation on Python 3.14

craig-meridian:
## Summary
- Add Python 3.14 (and keep 3.10 as the floor) to the CI test matrix (motivation: https://github.com/meridianlabs-ai/inspect_scout/pull/486#issuecomment-4846004034 & internal discussion at Meridian for standardising on testing floor + latest py versions).
- This surfaced a latent bug in scanner type validation.

## Scanner type validation bug
- Fix `_is_compatible_with_type`: its origin-equality fallback treated *any* union as compatible with a union target (e.g. `ChatMessageSystem | ChatMessageUser` passed as `ChatMessage`). The pipe spelling only escaped before 3.14 because `types.UnionType != typing.Union`; 3.14 unified the spellings and exposed it. `typing.Union[...]` was mishandled on every version.
- A union now satisfies a union target only when it covers every target member (extracted into a `_union_covers_union` helper)
- Regression tests pin the behaviour across both union spellings


craig-meridian:
Split into two PRs per review discussion: #530 (union type validation fix) and #531 (Python 3.14 CI matrix, depends on #530).

Issue #475, opened 2026-06-18
Scout indexes eval files sequentially, which is slow when those files are in S3


pipmc:
When running large scan jobs in [Hawk](https://github.com/metr/hawk) (over hundreds of eval files and thousands of samples), often the majority of the time is spent indexing those eval files, I believe because loading them from S3 is slow.

I believe this is because Scout indexes those eval files sequentially ([1](https://github.com/meridianlabs-ai/inspect_scout/blob/d06a74e5e0dc23b406fbe450149dfd063c1d4489/src/inspect_scout/_transcript/eval_log.py#L654C25-L654C25), [2](https://github.com/meridianlabs-ai/inspect_scout/blob/main/src/inspect_scout/_transcript/caching.py#L107)), although Inspect should [support loading them in parallel](https://github.com/UKGovernmentBEIS/inspect_ai/blob/dc8d44316d4cf302c3e3cc2ec6a40d7a37067eaf/src/inspect_ai/analysis/_dataframe/samples/table.py#L166). It would be great if we could request parallel indexing instead!

epatey:
I'm working on this in #528, and the proper fix is dependent on https://github.com/UKGovernmentBEIS/inspect_ai/pull/3684 getting merged.

epatey:
Whoops. This should have been closed when #528 went in.

PR #528, opened 2026-07-14
perf: index eval logs concurrently on a thread pool (~6x faster for S3 log directories)

epatey:
## Summary
Indexing previously read each eval log with a separate serial call, so jobs over hundreds of S3-hosted logs spent most of their time in per-file round trips (#475). This PR makes the two phases of indexing concurrent:

- **Log resolution** (`_resolve_logs`): paths are expanded and etags fetched concurrently on a shared warm `AsyncFilesystem` client. This builds directly on the recent inspect_ai change ([inspect_ai#3684](https://github.com/UKGovernmentBEIS/inspect_ai/pull/3684) and follow-on work) that gave `iter_files`/`iter_dirs` a `detail=True` mode returning the `FileInfo` the listing previously threw on the floor — made for exactly this use: a directory input now costs one listing sweep yielding file names *and* etags from `list_objects_v2` metadata, with no per-file stat calls or new fsspec call sites. A file input costs one async `info()`.
- **Sample reads**: cache misses are read concurrently on a thread pool (16 workers by default), one file per `samples_df` call, with each file cached (keyed by etag) as soon as its read completes — so a failure part way through preserves the files already read.

## Why one file per reader call (not batching)

An earlier iteration of this PR batched 32 paths into single multi-log `samples_df` calls. That was abandoned: `samples_df` de-duplicates samples by uuid across all logs it is given, and eval-set retry logs carry completed samples forward with identical uuids — so two retry logs for the same task read in one call each get a partial (and silently wrong) result, which then gets cached by etag. Found empirically on `s3://meridian-test-data/logs`: one retry log dropped from 1000 rows to 158 depending on which files shared a batch. Per-file calls make the de-duplication a no-op and match the semantics of the serial implementation exactly, at equivalent speed and with far less code (no combined-frame splitting, no re-read safeguards, no batch composition rules).

## Key Changes

**`caching.py`**
- `samples_df_with_caching` reads cache misses via `ThreadPoolExecutor` (default 16 workers, `max_workers` param); reader signature stays `Callable[[str], pd.DataFrame]`
- Each file is cached on completion (kvstore stays on the calling thread); on first failure it stops submitting new reads, caches in-flight successes, then re-raises
- `_resolve_logs` uses `run_coroutine`/`tg_collect` with a shared `AsyncFilesystem`: directories expand via a single `iter_files(recursive=True, detail=True)` sweep, files via concurrent `info()`. S3 has no directory objects so `info()` raises for prefixes; a path that lists non-empty is treated as a directory, anything else re-raises

**`eval_log.py`**
- `_index_logs` reader takes a single path; `trace_action` wraps each file read

**Tests**
- Mock `AsyncFilesystem` (async `info`/`iter_files`) replaces the fsspec mocks
- New coverage: S3 directory-prefix expansion, missing-path error propagation, concurrent etag→path association, failure preserving completed reads

## Performance Measurement

Cold-cache indexing of `s3://meridian-test-data/logs` (~336 objects → 295 eval logs). Method: fresh process, `scout_transcript_info_cache.db` deleted before each run, wall time of `_index_logs("s3://meridian-test-data/logs")`.

| Version | Wall time | Samples indexed |
|---------|-----------|-----------------|
| `origin/main` (serial, one file per call) | 1062 s (~17.7 min) | 14,052 |
| This PR (concurrent per-file reads, 16 threads) | 169 s (~2.8 min) — **~6× faster** | 14,052 |

Per-file cache contents verified byte-identical to the serial implementation across all 295 files (including the eval-set retry pair that the batched design corrupted). Wall time on this workload varied roughly ±25% across runs (S3 read throughput), so treat the speedup as approximate.

## Possible Follow-ups

- Each per-file `samples_df` call warms its own S3 client (~0.5–1 s, overlapped across threads). A process-shared warm client in inspect_ai (or `samples_df` accepting an `AsyncFilesystem`) would shave this.
- `_resolve_logs`'s etag logic could migrate into inspect_ai's `resolve_logs()`.

https://claude.ai/code/session_01ThJmjTpPzpe4M798ScGu1W

### DIFF
--- a/src/inspect_scout/_transcript/caching.py
+++ b/src/inspect_scout/_transcript/caching.py
@@ -2,7 +2,7 @@ import io
 import json
 import warnings
 from collections.abc import Callable, Generator
-from concurrent.futures import ThreadPoolExecutor, as_completed
+from concurrent.futures import Future, ThreadPoolExecutor, as_completed
 from contextlib import contextmanager
 from functools import partial
 from os import PathLike
@@ -125,6 +125,7 @@ def samples_df_with_caching(
         # reads, cache the reads still in flight, then re-raise.
         miss_df_by_path: dict[str, pd.DataFrame] = {}
         first_error: Exception | None = None
+        futures: dict[Future[pd.DataFrame], tuple[str, str]] = {}
         if cache_misses:
             with ThreadPoolExecutor(
                 max_workers=min(max_workers, len(cache_misses))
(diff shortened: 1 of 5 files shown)


epatey:
@claude

epatey:
@claude please review

PR #530, opened 2026-07-15
fix: reject scanners whose union types don't cover all filtered message/event types

craig-meridian:
## Summary

`_is_compatible_with_type` had an origin-equality fallback that treated *any* union as compatible with a union target — e.g. a scanner typed `ChatMessageSystem | ChatMessageUser` passed validation as `ChatMessage`. A union now satisfies a union target only when it covers every target member.

The old check compared union origins, never union members, so partial unions written as `typing.Union[X, Y]` passed validation on every Python version. Pipe-spelled unions (`X | Y`) were broken before 3.14 too, just in the opposite direction: their origin (`types.UnionType`) never matched `typing.Union`, so any pipe union not exactly equal to the target was rejected outright — even when its members covered it. Python 3.14 unified the two spellings, flipping the pipe form from over-rejection to over-acceptance, which is how this surfaced (see #531).

## ⚠️ Change in behaviour

This tightens validation: scanners with partial-union types that previously passed validation **incorrectly** will now be rejected at registration with a `TypeError`. Those scanners were already receiving message types they didn't declare — the new error is surfacing a real mismatch, but existing code may need its type annotations (or filters) fixed.

This is unlikely to affect many users, so we're not classifying this as a breaking change. Here is an example of a scanner which would be implicated in this change:

```py
from typing import Union

from inspect_ai.model import ChatMessageSystem, ChatMessageUser
from inspect_scout import Result, Scanner, scanner


@scanner(messages=["system", "user", "assistant"])
def broken() -> Scanner[Union[ChatMessageSystem, ChatMessageUser]]:  # <- missing assistant, should fail
    async def scan(message: Union[ChatMessageSystem, ChatMessageUser]) -> Result:
        return Result(value=True)

    return scan


broken()  # before: registers fine; after: TypeError "must be able to handle all types"
```

### DIFF
--- a/src/inspect_scout/_scanner/validate.py
+++ b/src/inspect_scout/_scanner/validate.py
@@ -465,18 +465,19 @@ def _can_handle_all_types(
     return _is_compatible_with_type(scanner_type, required_type)
 
 
+def _is_union_type(type_hint: Any) -> bool:
+    """Check if a type hint is a union, in either spelling.
+
+    Covers ``typing.Union[X, Y]`` and the ``X | Y`` syntax (distinct types
+    before Python 3.14, unified from 3.14 on).
+    """
+    return get_origin(type_hint) is Union or isinstance(type_hint, UnionType)
+
+
 def _get_union_members(type_hint: Any) -> set[Type[Any]] | None:
     """Get the member types of a Union, or None if not a Union."""
-    origin = get_origin(type_hint)
-
-    # Handle Union from typing module
-    if origin is Union:
+    if _is_union_type(type_hint):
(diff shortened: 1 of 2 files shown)


craig-meridian:
Worth a discussion first to discuss:
- Do we classify this as a breaking change?

craig-meridian:
Discussed with Eric, and we thought this is quite unlikely to affect any users:
- Most don't supply unions of specific message types - they'd either take all or none.
- Any which have been subject to this bug would probably have run into runtime errors and adjusted their type hints accordingly already.

epatey (approved):

PR #534, opened 2026-07-15
fix: robust union detection in import CLI on Python 3.14

craig-meridian:
## Summary

Follow-up from Python 3.14 testing (see #530 and #531). This fixes two bugs in how we inspect type annotations in the import CLI:

1. `_get_type_name` and `_is_str_only_annotation` detected unions by checking `str(origin) == "typing.Union"`. On Python 3.14 that string is now `"<class 'typing.Union'>"`, so the check never matches. Nothing actually breaks today because the `types.UnionType` fallback happens to catch both union spellings on 3.14 — but the check is dead code waiting to bite.

2. `_has_datetime_annotation` and `_has_int_annotation` looked at `__args__` on *any* parameterized generic, not just unions. No current source has a parameter that trips this, so it's latent — but the first source to add a param like `list[int]` would get its CLI string run through `int()`, and `list[datetime]` through `datetime.fromisoformat()`.

## Changes

- Added `_util/type_hints.is_union_type` using the [documented cross-version idiom](https://docs.python.org/3.14/library/typing.html#typing.Union), and switched all the ad-hoc union checks over to it (`_cli/import_command.py`, `_scanner/validate.py` from #530, and `_scanner/_loaders.py`).
- The CLI now only inspects union members when the annotation is actually a union.
- `Annotated` metadata wrappers are unwrapped before scalar classification, including when an individual union member is annotated.
- Added tests covering both union spellings (`typing.Union[X, None]` and `X | None`) and confirming parameterized generics like `list[int]` aren't treated as unions.

## Known limitation

Postponed annotations are stored as strings, so `Annotated[...]` used with `from __future__ import annotations` still falls back to YAML coercion. This pre-existing behavior is intentionally left unchanged: evaluating annotation strings during inspection can execute annotation expressions or fail when names are unavailable at runtime, such as imports guarded by `TYPE_CHECKING`.


src/inspect_scout/_scanner/validate.py
```diff
@@ -465,18 +464,9 @@ def _can_handle_all_types(
     return _is_compatible_with_type(scanner_type, required_type)
 
 
-def _is_union_type(type_hint: Any) -> bool:
```
craig-meridian, line 468: Moved to shared `_util.type_hints.py`

src/inspect_scout/_util/type_hints.py
```diff
@@ -0,0 +1,13 @@
...
+from typing import Any, Union, get_origin
+
+
+def is_union_type(type_hint: Any) -> bool:
+    """Check if a type hint is a union, in either spelling.
+
+    Covers ``typing.Union[X, Y]`` and the ``X | Y`` syntax (distinct types
+    before Python 3.14, unified from 3.14 on).
+    """
+    return get_origin(type_hint) in (Union, UnionType)
```
craig-meridian, line 13: This is the cross-version union check recommended by the [typing docs](https://docs.python.org/3.14/library/typing.html#typing.Union) 

> For compatibility with earlier versions of Python, use `get_origin(obj) is typing.Union or get_origin(obj) is types.UnionType`

 Spelled as tuple membership, which is equivalent since neither object defines `__eq__`. The docs' preferred `isinstance(obj, Union)` needs 3.14+ (it raises `TypeError` on ≤3.13), so once we drop support for ≤3.13 this body can shrink to that.

tests/cli/test_coerce_value.py
```diff
@@ -1,62 +1,100 @@
...
-        "str_or_list_or_none",
-        "bare_int",
-        "int_or_none",
-        "list_str",
-        "string_str",
-        "string_str_or_none",
-        "string_str_or_list",
-        "string_str_or_list_or_none",
-        "string_int",
+    [
```
craig-meridian, line 20: Apologies for the noisy diff here. I wasn't a fan of the split `_STR_ONLY_CASES` and `ids=` which were hard to reconcile and easy to end up out of sync. I added new tests (below) so thought I'd update this test to match the (IMO) much more maintainable style.

Issue #502, opened 2026-07-10
Default exclude_columns to heavy columns (input, input_data, scan_events) in scan_results_df and ScanResultsArrow.reader


epatey:
## Motivation

Scan results parquet files contain three "heavy" JSON columns that dominate file size and memory usage:

| Column | Contents |
|--------|----------|
| `input` | Full serialized scanner input (transcript / messages / events JSON) |
| `input_data` | Deduplicated message/call pools referenced by condensed event refs |
| `scan_events` | Scanner execution events (model calls, logging, etc.) |

Today `exclude_columns=None` means "exclude nothing", so every consumer pays the cost of these columns unless they know to opt out. The frontend already works around this by explicitly passing `["input", "scan_events"]` (`useSelectedScanDataframe.ts`). The common case (analysis over values/scores/metadata) never needs the heavy columns — the default should reflect that.

## Proposal

Change the default for `exclude_columns` to exclude the heavy columns (`input`, `input_data`, `scan_events`) in:

- `scan_results_df()` / `scan_results_df_async()` (`_scanresults.py`)
- `ScanRecorder.results_df()` ABC + file recorder implementation (`_recorder/recorder.py`, `_recorder/file.py`)
- `ScanResultsArrow.reader()` (`_recorder/recorder.py:59`, impl in `_recorder/file.py:394`)

### New semantics

| `exclude_columns` | Behavior |
|-------------------|----------|
| `None` (default) | Exclude heavy columns (`input`, `input_data`, `scan_events`) |
| `[]` | Exclude nothing — all columns |
| explicit list | Exclude exactly those columns |

Define the heavy set once (e.g. `HEAVY_COLUMNS` constant) rather than duplicating the list.

### Notes

- **df path**: when `input` is excluded, `_expand_events_in_df()` has nothing to expand — `input_data` should be excluded from the parquet read as well (excluding it is required for the expansion logic to be skipped cleanly).
- **HTTP API unchanged**: the `scan_df` endpoint defaults `exclude_column=[]`, which under the new semantics still means "exclude nothing", so endpoint behavior is preserved and clients (like the frontend) keep explicit control. The frontend could optionally drop its hand-rolled exclusion list later.
- **`scan_detail` endpoint**: calls `scan_results_df_async()` without `exclude_columns` but its `response_model=ScanStatus` never serializes the dataframes, so no behavior change.
[...]

PR #538, opened 2026-07-18
perf: fix quadratic `_expand_events_in_df`

nix-aisi:
The old code did
```
for idx in df.index[mask]:
   df.at[idx, "input"] = json.dumps(transcript)
```

But this requires rebuilding the arrow-backed string column every time, meaning it's O(rows * column_bytes). This is pretty slow as inputs are sometimes big!

It's easy to make it much faster, by assembling the column in a list and writing it once. On a 391 MB scan I had on s3 this speeds up `scan_results_df` from 442s -> 26.6s end-to-end.

This is less necessary if #502 is already fixed (since this code will not run by default) but no downside to the fix. 

Testing: I verified the output was identical via `pd.testing.assert_frame_equal`


src/inspect_scout/_scanresults.py
```diff
@@ -347,21 +347,27 @@ def _expand_events_in_df(df: pd.DataFrame) -> pd.DataFrame:
     df = df.copy()
     mask = df["input_data"].notna()
 
-    for idx in df.index[mask]:
-        input_json = str(df.at[idx, "input"])
-        input_data_json = str(df.at[idx, "input_data"])
-        input_type = str(df.at[idx, "input_type"])
+    # Build replacement values in a list and assign the column once: per-row
+    # df.at writes into an Arrow-backed string column rebuild the entire column
+    # each time, making this loop O(rows * column_bytes).
```
nix-aisi:
```suggestion
```

epatey:
@review

epatey (approved):

Issue #537, opened 2026-07-16
copyreg.pickle(Model, ...) registration hijacks copy.copy() on inspect_ai Model process-wide


ransomr:
## Problem

`src/inspect_scout/_concurrency/_mp_common.py` registers a global pickle reducer for inspect_ai's `Model` at import time:

```python
copyreg.pickle(Model, _reduce_model)
```

The intent is pickle-only (letting scanner closures that capture a `Model` cross spawn multiprocessing boundaries via cloudpickle). But `copyreg.dispatch_table` is also consulted by `copy.copy()` and `copy.deepcopy()` — the copy module shares the pickle reduction protocol. So merely importing `inspect_scout` changes what `copy(model)` means process-wide: instead of a shallow copy, it round-trips through `_reconstruct_model` → `get_model(...)`, which **memoizes** — returning a shared cached instance rather than a distinct object.

## Impact on inspect_ai

inspect_ai relies on `copy(model)` producing a genuinely distinct instance when stamping model roles (`Model._set_role`) and applying role-scoped config overrides (`get_model(role=..., config=...)` copies the role's model before rebinding `.config`). With scout imported, those copies alias the memoized instance, so:

- a role-scoped config override mutates the shared model other callers are using
- role stamping collapses distinct roles onto one instance, misattributing `role_usage`

This surfaced as CI failures in UKGovernmentBEIS/inspect_ai#4465 (see the analysis comment there: tests passed in isolation but failed in any pytest-xdist worker where a scout-importing test had run first). inspect_ai is adding a defensive `Model.__copy__` (which `copy.copy()` checks before the dispatch table), but the global registration is worth fixing here too — it affects any `copy()` caller in any process that imports scout, not just inspect_ai internals.

## Suggested fix

Scope the reducer to pickling. Only the dump side needs it (the pickled data embeds a reference to `_reconstruct_model`, so unpickling in workers requires no registration). Two options:

**`reducer_override` (recommended, Python ≥ 3.8):** checked per-pickler before any table, invisible to `copy`:

```python
class ScoutPickler(cloudpickle.Pickler):
    def reducer_override(self, obj):
        if isinstance(obj, Model):
            return _reduce_model(obj)
        return super().reducer_override(obj)
```

then use `ScoutPickler` at the `cloudpickle.dumps` call sites (e.g. `DillCallable.__init__`).

**Per-pickler `dispatch_table` (classic alternative):**

```python
buf = io.BytesIO()
[...]

PR #542, opened 2026-07-20
fix: scope Model reducer to Scout's pickler, not global copyreg

epatey:
## Summary

Fixes issue #537 by moving the `Model` pickle reducer from global `copyreg` registration to a dedicated `ScoutPickler` class that uses `reducer_override`. This prevents the reducer from leaking into `copy.copy()` and `copy.deepcopy()` semantics process-wide, which was breaking inspect_ai's role stamping and role-scoped config overrides.

## Key Changes

- **Replaced global `copyreg.pickle()` registration** with a `ScoutPickler` class that extends `cloudpickle.Pickler` and implements `reducer_override()` to handle `Model` instances
- **Added `scout_dumps()` function** that mirrors `cloudpickle.dumps()` but uses `ScoutPickler`, ensuring scanner closures with captured `Model` instances can be serialized without polluting global pickle/copy behavior
- **Updated `DillCallable.__init__`** to use `scout_dumps()` instead of `cloudpickle.dumps()`
- **Updated tests** to verify that importing scout does not hijack `copy.copy()` semantics and that `scout_dumps()` correctly serializes models in subprocesses

## Implementation Details

The `reducer_override` method is consulted per-pickler before any dispatch table, so it stays scoped to serialization and never affects the copy module's behavior. The unpickle side requires no registration since the pickled data embeds `_reconstruct_model` directly.

This approach maintains the memoization benefits of `get_model()` (reusing cached instances in workers) while preventing unintended aliasing when users call `copy.copy(model)` on their own models.

https://claude.ai/code/session_014aNt2wz3DGc5CM6KZFbveg

### DIFF
--- a/src/inspect_scout/_concurrency/_mp_common.py
+++ b/src/inspect_scout/_concurrency/_mp_common.py
@@ -44,19 +44,28 @@ if TYPE_CHECKING:
 
 
 # ---------------------------------------------------------------------------
-# copyreg reducer: make inspect_ai Model picklable via cloudpickle
+# Pickler-scoped reducer: make inspect_ai Model picklable via cloudpickle
 # ---------------------------------------------------------------------------
 # ModelAPI subclasses contain unpicklable state (anyio.Lock, httpx.AsyncClient,
 # SDK async clients).  When a scanner closure captures a Model instance,
 # cloudpickle fails with "cannot pickle '_thread.RLock' object".
 #
-# We register a copyreg reducer that converts Model → ModelConfig for
-# serialization, then reconstructs via get_model() on unpickle.  get_model()
-# uses memoize=True by default, so repeated unpickles in the same worker
-# reuse the cached instance rather than creating duplicate HTTP clients.
+# We reduce Model → ModelConfig for serialization, then reconstruct via
+# get_model() on unpickle.  get_model() uses memoize=True by default, so
+# repeated unpickles in the same worker reuse the cached instance rather than
+# creating duplicate HTTP clients.
+#
+# The reducer is scoped to a dedicated `ScoutPickler` (via reducer_override)
+# rather than registered globally with `copyreg.pickle(Model, ...)`.  A global
+# registration pollutes `copyreg.dispatch_table`, which `copy.copy()` and
+# `copy.deepcopy()` also consult — so merely importing scout would change what
+# `copy(model)` means process-wide, aliasing the memoized instance instead of
+# producing a distinct copy (see issue #537).  reducer_override is consulted
+# only by our pickler and is invisible to the copy module.  The unpickle side
+# needs no registration: the pickled data embeds `_reconstruct_model` directly.
 # ---------------------------------------------------------------------------
 
-import copyreg
+import io
 
 from inspect_ai.model import get_model
 from inspect_ai.model._model import Model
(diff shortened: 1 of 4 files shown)


epatey:
@auto

epatey:
@auto

i-am-marvin:
**Claude finished @epatey's task in 17m 28s** —— [View job](https://github.com/meridianlabs-ai/inspect_scout/actions/runs/29772366968)

---
### Summary

All CI is now green on this PR ✅ (Build run on `338feffc`: [success](https://github.com/meridianlabs-ai/inspect_scout/actions/runs/29772818181)).

- [x] Bring branch up to date with `origin/main` — already up to date (merge commit `30b410ca` includes latest main)
- [x] Review the PR diff — implementation matches the fix recommended in #537; all spawn-boundary callables route through `DillCallable` → `scout_dumps`
- [x] Diagnose CI failure — `py-test (3.14)` failed on the two previous commits: `tests/llm_scanner` context-window tests hit `assert isinstance(sem, AdaptiveConcurrencyController)` inside inspect_ai whenever a multi-process scan test had run earlier on the same pytest-xdist worker
- [x] Fix root cause and push `338feffc`
- [x] Confirm CI green on the new commit

### What was wrong

The failure was **not** in this PR's pickler change — it was a pre-existing latent bug that the new e2e test exposed by shifting xdist test scheduling: `multi_process.py` installs the cross-process `ParentSemaphoreRegistry` via `init_concurrency(parent_registry)` at scan start but never restored the default registry afterward. The leaked registry downgraded every later `concurrency()` request in the parent process (including inspect_ai's adaptive model connections) to fixed-limit MP semaphores — the same "global state leaks out of Scout" class of bug as #537 itself.

### The fix (`338feffc`)

- `src/inspect_scout/_concurrency/multi_process.py`: restore the default in-process registry with `init_concurrency()` in the strategy's outer `finally` block (child processes need nothing — they exit with the scan)
- `tests/concurrency/test_mp_model_capture_e2e.py`: after the multiprocess scan, assert that `concurrency()` no longer yields an `MPConcurrencySemaphore`, covering the restore

Note: I couldn't run lint/typecheck/tests locally (package installation is blocked in this environment), so verification was done via CI, which runs the full check suite and is now green.
 | [Branch](https://github.com/meridianlabs-ai/inspect_scout/tree/claude/issue-537-fix-76eg8w)

i-am-marvin:
@review

epatey:
@auto

i-am-marvin:
**Claude finished @epatey's task in 4m 31s** —— [View job](https://github.com/meridianlabs-ai/inspect_scout/actions/runs/29775552683)

---
### Review feedback addressed ✅

- [x] Bring branch up to date with `origin/main` — already up to date
- [x] Read `multi_process.py` and confirm the gap
- [x] Widen the `try` to cover the registry install (pre-initialize `original_sigint_handler`)
- [x] Run local checks — `ruff format --check`, `ruff check`, and a standalone `mypy` pass on the file are clean (full test suite can't run here since package installation is blocked; CI covers it)
- [x] Commit and push `bfec28d7`
- [x] Resolve the addressed review thread
- [x] Post `@review` to re-run the automated reviewer

### What changed (`bfec28d7`)

The review's finding on `src/inspect_scout/_concurrency/multi_process.py`: the `finally` that restores the default concurrency registry only covered exceptions raised inside the `try`, but the cross-process `ParentSemaphoreRegistry` was installed *before* it — so `find_inspect_log_handler()` raising `RuntimeError` or `signal.signal` raising `ValueError` (off the main thread) would leak the registry and leave `_active = True`, blocking all subsequent scans in the process.

Fix, exactly as suggested in the review:
- The `try` now begins immediately after `_active = True`, so the `finally` (`init_concurrency()` restore + `_active = False`) covers every setup failure path, including `Manager()` creation
- `original_sigint_handler` is pre-initialized to `None` (typed via a new module-level `SignalHandler` alias mirroring typeshed's `signal._HANDLER`), and the `finally` only restores it when SIG_IGN was actually installed — previously an early failure would have restored a handler that was never captured

No behavior change on the happy path; the restructure is purely widening the cleanup scope.
 | [Branch](https://github.com/meridianlabs-ai/inspect_scout/tree/claude/issue-537-fix-76eg8w)

i-am-marvin:
@review

PR #545, opened 2026-07-21
fix: absence-implied true-negative validation rows suppressed by labels on other transcripts

tbroadley:
(written by Claude, edited by Thomas)

## Repro case

Suppose we have two transcripts and one scanner that detects multiple kinds of findings and returns labeled Results in a ResultSet. Also, we have a validation set saying what each transcript should contain:

```yaml
- id: t1
  labels: { deception: true,  phishing: false }   # t1 really contains deception, no phishing
- id: t2
  labels: { deception: false, phishing: false }   # t2 is clean
```

The scanner runs once per transcript:

- On `t1` it returns `[Result(label="deception", value="found")]`. This is a true positive for deception and a true negative for phishing.
- On `t2` it returns `[Result(label="phishing", value="found")]`. This is a true negative for deception and a false positive for phishing.

`validation_result` is stored as follows:

| row | value (resultset JSON) | validation_target | validation_result |
|---|---|---|---|
| t1 | `[{label: deception, value: found}]` | `{deception: true, phishing: false}` | `{deception: true, phishing: true}` |
| t2 | `[{label: phishing, value: found}]` | `{deception: false, phishing: false}` | `{deception: true, phishing: false}` |

`scan_results_df()` expands each ResultSet into a single DataFrame row per Result. It adds synthetic rows for true negatives that otherwise wouldn't appear in the DataFrame. The correct output is one row per transcript and label:

| transcript | label | value | validation_result | |
|---|---|---|---|---|
| t1 | deception | found | True | real |
| t1 | phishing | False | True | **synthetic** (true negative) |
| t2 | phishing | found | False | real (false positive) |
| t2 | deception | False | True | **synthetic** (true negative) |

However, `_handle_label_validation()` computes whether each label is present by looking over the entire DataFrame. Both labels are present on at least one row in the DataFrame, causing this function to skip adding true negatives.

## The fix

Instead of computing label presence over the entire DataFrame, compute it for each row with a ResultSet value separately. That way, the logic correctly detects that deception is missing for one transcript and phishing for the other, and creates the true negative rows.
Also, the per-label result propagation now assigns only the `validation_result` column instead of applying a whole-row transform. `DataFrame.apply(axis=1)` over whole rows lets pandas re-infer dtypes of untouched columns (e.g. coercing an object column of Python bools to numpy bool dtype, which conflicts with #544's guarantee that boolean resultset values come back as real `True`/`False`).

### DIFF
--- a/src/inspect_scout/_scanresults.py
+++ b/src/inspect_scout/_scanresults.py
@@ -215,21 +215,13 @@ def _handle_label_validation(
 ) -> tuple[pd.DataFrame, pd.DataFrame]:
     """Handle label-based validation for expanded resultset rows.
 
-    Label validation is computed at scan time per scanner invocation (see
-    `_validate_labels()`), so its presentation is scoped the same way: each
-    resultset row is checked against its own `validation_target`.
-
     This function:
     1. Propagates per-label validation results to individual expanded rows
-    2. Creates synthetic rows for expected-absent labels (negative expected
-       value) with no result in their own resultset row, making true
-       negatives visible in the DataFrame
+    2. Creates synthetic rows for missing labels (where expected value is negative)
 
     Args:
-        expanded: DataFrame of expanded result rows (carrying
-            `_RESULTSET_INDEX`); may be empty
-        resultset_rows: Original resultset rows before expansion (carrying
-            `_RESULTSET_INDEX`)
+        expanded: DataFrame of expanded result rows
+        resultset_rows: Original resultset rows before expansion
 
     Returns:
         Tuple of (expanded DataFrame with validation, synthetic rows DataFrame)
--- a/src/inspect_scout/_scanresults.py
+++ b/src/inspect_scout/_scanresults.py
@@ -193,140 +193,132 @@ def remove_scan_results(scan_location: str) -> None:
         scan_path.rmdir(recursive=True)
 
 
+_RESULTSET_INDEX = "_resultset_index"
+"""Temporary column tying expanded rows back to their originating resultset row."""
+
+
+def _parse_json_dict(value: Any) -> dict[str, Any] | None:
+    """Parse a value as a JSON object, returning None if it isn't one."""
+    if isinstance(value, dict):
+        return value
+    if not isinstance(value, str) or not value:
+        return None
+    try:
+        parsed = json.loads(value)
+    except json.JSONDecodeError:
+        return None
+    return parsed if isinstance(parsed, dict) else None
+
+
 def _handle_label_validation(
     expanded: pd.DataFrame, resultset_rows: pd.DataFrame
 ) -> tuple[pd.DataFrame, pd.DataFrame]:
     """Handle label-based validation for expanded resultset rows.
 
+    Label validation is computed at scan time per scanner invocation (see
+    `_validate_labels()`), so its presentation is scoped the same way: each
+    resultset row is checked against its own `validation_target`.
+
     This function:
     1. Propagates per-label validation results to individual expanded rows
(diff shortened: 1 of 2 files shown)


src/inspect_scout/_scanresults.py
```diff
@@ -193,140 +193,132 @@ def remove_scan_results(scan_location: str) -> None:
...
+    Label validation is computed at scan time per scanner invocation (see
+    `_validate_labels()`), so its presentation is scoped the same way: each
+    resultset row is checked against its own `validation_target`.
+
     This function:
     1. Propagates per-label validation results to individual expanded rows
-    2. Creates synthetic rows for missing labels (where expected value is negative)
+    2. Creates synthetic rows for expected-absent labels (negative expected
+       value) with no result in their own resultset row, making true
+       negatives visible in the DataFrame
```
tbroadley: I think we can drop these changes to this docstring, doesn't seem useful to document this.

src/inspect_scout/_scanresults.py
```diff
@@ -407,8 +403,13 @@ def _expand_resultset_rows(df: pd.DataFrame) -> pd.DataFrame:
...
     if expanded.empty:
-        # No actual results to expand, just return other rows
-        return other_rows.reset_index(drop=True)
+        # No actual results to expand, but synthetic missing-label rows may
+        # still apply (e.g. empty resultsets with negative label expectations)
+        _, synthetic_rows = _handle_label_validation(expanded, resultset_rows)
+        if synthetic_rows.empty:
+            return other_rows.reset_index(drop=True)
+        synthetic_rows = synthetic_rows.drop(columns=[_RESULTSET_INDEX])
+        return _concat_aligned([other_rows, synthetic_rows])
```
tbroadley, line 412: Do we have a test for this case where the expanded set is empty but we still need to add synthetic rows?

src/inspect_scout/_scanresults.py
tbroadley: Reverted the docstring to its original wording in 9eb020c8.

src/inspect_scout/_scanresults.py
tbroadley: Yes — `test_synthetic_rows_for_empty_resultset` hits this path (single resultset row with `results: []`, so the whole expanded frame is empty), and `test_synthetic_rows_use_each_rows_own_target` also goes through it (both transcripts have empty resultsets). `test_synthetic_rows_without_stored_validation_result` covers the neighboring non-empty path.

dragonstyle:
@review

src/inspect_scout/_scanresults.py
```diff
@@ -211,122 +228,97 @@ def _handle_label_validation(
...
+        present = present_labels.get(resultset_row[_RESULTSET_INDEX], set())
+        missing_labels = set(parsed_target.keys()) - present
 
-    # Missing labels = expected but not present
-    missing_labels = expected_labels - present_labels
+        for label in sorted(missing_labels):
+            expected_value = parsed_target[label]
+            # Only create synthetic row if expected value is negative
+            negative_values = (False, None, "NONE", "none", 0, "")
+            if expected_value not in negative_values:
```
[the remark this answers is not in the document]
tbroadley: Done in 256dc40e — replaced the inline tuple with `is_positive_value()` from `_validation/validate.py` so synthesis uses the same negativity semantics as scan-time validation.

epatey:
Retitled the PR (was: "fix: scope label validation in results DataFrames per resultset row"). I literally had to back solve from the fix to figure out what the symptom was.

epatey (approved):

PR #546, opened 2026-07-21
fix: bound memory usage on large scans (streamed compaction + bounded results queue)

Axel-At-Apollo:
## Problem

On large scans (tens of thousands of transcripts), scout's main-process memory usage can grow far beyond the working set of the scan itself — in our case repeatedly freezing a 48GB workstation (macOS swap storm) late in a ~48k-transcript scan. Two mechanisms compound:

1. **Finalize compaction materializes the entire output in RAM.** `scanner_table` writes the compacted parquet into a `pa.BufferOutputStream` and then copies it out with `to_pybytes()`, so `sync()` peaks at ~2x the compacted parquet size per scanner. Because each result row embeds the scanned transcript (`input`) and the scan events, large scans compact to multi-GB parquets (we measured ~310KB/row on agentic transcripts, i.e. a ~15GB buffer dir for one scan), making that spike large enough to take down the host. Interrupts hit the same path, since Ctrl-C handling also runs `sync()`.

2. **The upstream results queue is unbounded.** The parse-job queue is bounded (with a comment explaining exactly this risk), but the results direction is not: each `ResultItem` embeds full transcripts, so whenever workers produce results faster than the main process can record them the backlog grows without limit for the life of the scan.

## Changes

**Streamed compaction** (`_recorder/buffer.py`, `_recorder/file.py`):
- `scanner_table` now streams the compacted parquet directly to a local output file via `ParquetWriter` instead of an in-memory buffer + bytes copy (this also resolves the existing TODO about avoiding the `to_pybytes` copy). Signature changes from returning `bytes | None` to writing `output_file` and returning `bool`.
- `FileRecorder.sync` compacts straight into the sibling `.tmp` when the scan dir is local (same atomic-rename semantics as before), or into a temp file that is uploaded with `AsyncFilesystem.write_file_streaming` for remote scan dirs. Compaction memory is now bounded by the existing ~100MB input-accumulation cap regardless of output size.

**Bounded upstream queue** (`_concurrency/multi_process.py`, `_mp_subprocess.py`, `_mp_shutdown.py`):
- `upstream_queue` gets `maxsize=UPSTREAM_QUEUE_MAXSIZE` (100), so workers experience backpressure instead of the main process accumulating results.
- Worker-side puts that can now block (`ResultItem`, `WorkerReady`, `WorkerComplete`, work-task exceptions) are wrapped in `run_sync_on_thread` so only the putting task waits, never the worker's event loop.
- Sync-context puts use `put_nowait` and drop on `Full`: metrics snapshots (the next update supersedes) and forwarded log records (dropping a record under backpressure beats stalling the event loop from inside a logging handler).
- The shutdown sentinel injection uses `put_nowait`: a blocking put on a full queue would hang shutdown, and if the queue is full the collector cannot be blocked on `get()` anyway.

## Testing

- Adapted `tests/scanner/test_recorder_buffer.py` to the new `scanner_table` signature.
- `pytest tests/recorder tests/concurrency tests/scan tests/scanner`: 679 passed, 1 skipped.
- Full suite run: the only failures (`test_project.py`, langsmith source, anthropic cache-control import) fail identically on `main` in my environment (missing optional deps), i.e. pre-existing.
- `ruff check`/`ruff format` and `mypy` clean on all touched files.

Co-Authored-By: Claude

### DIFF
--- a/src/inspect_scout/_concurrency/_mp_shutdown.py
+++ b/src/inspect_scout/_concurrency/_mp_shutdown.py
@@ -4,7 +4,7 @@ import time
 from collections.abc import Sequence
 from multiprocessing.context import SpawnProcess
 from multiprocessing.queues import Queue
-from queue import Empty
+from queue import Empty, Full
 from typing import Any, Callable
 
 import anyio
@@ -125,8 +125,15 @@ async def shutdown_subprocesses(
     # PHASE 5: Inject shutdown sentinel to wake collector
     print_diagnostics("SubprocessShutdown", "Phase 5: Injecting shutdown sentinel")
     try:
-        ctx.upstream_queue.put(shutdown_sentinel)
+        # put_nowait: the queue is bounded, and a blocking put on a full queue
+        # would hang shutdown. If the queue is full the collector cannot be
+        # blocked on get() anyway, so the wake-up sentinel isn't needed.
+        ctx.upstream_queue.put_nowait(shutdown_sentinel)
         print_diagnostics("SubprocessShutdown", "Injected upstream queue sentinel")
+    except Full:
+        print_diagnostics(
+            "SubprocessShutdown", "Upstream queue full - sentinel not needed"
+        )
     except (ValueError, OSError) as e:
         # Queue already closed - collector likely already exited via cancellation
         print_diagnostics("SubprocessShutdown", f"Upstream queue closed: {e}")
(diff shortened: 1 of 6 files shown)


Axel-At-Apollo:
This is fully vibecoded, and I have not set off time to review this, will close for now, but the problem of my machine crashing when grading many samples in parallel (like 3k concurrents) stands

PR #550, opened 2026-07-30
feat: add results_buffer option for periodic sync of in-progress results

epatey:
Adds a `results_buffer` scan option that enables periodic syncing of in-progress results to the scan location while a scan is still running, allowing users to inspect partial results without waiting for completion.

> **Companion PR:** meridianlabs-ai/ts-mono#480 (regenerated `generated.ts`). Two-phase landing: this PR's gitlink currently pins that branch's head, so `submodule-on-main` stays red until ts-mono#480 merges; then the gitlink gets bumped to the merged main SHA and this PR can land.

## Summary

This feature addresses the need to monitor long-running scans by syncing compacted results to the scan location every N recorded results. Results are written as background tasks so their latency doesn't block the scanning pipeline, and syncs are skipped if one is already in flight.

## Key Changes

- **New `results_buffer` option**: Added to `scan()`, `scan_async()`, `ScanJob`, `ScanJobConfig`, and `ScanOptions`. Accepts an integer N to sync every N results, or None (default) to sync only on completion.

- **Periodic sync implementation**: In the scan event loop, a background task group manages periodic syncs triggered by `record_results()`. Syncs run asynchronously via `recorder.sync(complete=False)` and are skipped if one is already in flight to avoid queueing.

- **Optimized compaction for remote filesystems**: 
  - Added `_prior_parquet_cache` to cache downloaded prior parquets per scan run, avoiding re-downloads on each periodic sync
  - Moved `scanner_table()` compaction to a worker thread via `anyio.to_thread.run_sync()` to keep the event loop responsive during mid-scan syncs
  - Made `_sync_status_files()` async to write summary/errors through the async filesystem for remote locations

- **Atomic writes for remote/local filesystems**:
  - Remote object stores (S3, etc.) write directly to the final key (atomic on upload completion)
  - Local filesystems write to `.tmp` then atomically rename to avoid read/write overlap

- **Cache invalidation**: Clears the prior parquet cache when a run starts (`init`/`resume`/`attach`) and when a sync completes the scan, ensuring stale cached files don't persist across runs.

- **CLI and documentation**: Added `--results-buffer` option to the scan command with environment variable support, and documented the feature in the workflow guide.

## Implementation Details

- Syncs are triggered in `record_results()` after each result is recorded, checking if the count since the last sync has reached the buffer threshold
- Failed periodic syncs log a warning but don't interrupt the scan; the final sync (or a later periodic one) will retry
- The feature is fully compatible with both `scan()` and `scan_async()`, and works across different concurrency strategies and process counts
- Comprehensive end-to-end tests verify partial results appear mid-scan, final results are unchanged, and the option is honored in async contexts

https://claude.ai/code/session_0127Dxge9RML8K6Ajej9BfNa


epatey:
This PR supersedes #500 (now closed) — see the [analysis](https://github.com/meridianlabs-ai/inspect_scout/pull/500#issuecomment-5136017778) and [adaptation notes](https://github.com/meridianlabs-ai/inspect_scout/pull/500#issuecomment-5136275404) there for why the original inline-sync approach was slow on S3 and what changed.

@dragonstyle — could you rerun your S3 perf measurement from [#500](https://github.com/meridianlabs-ai/inspect_scout/pull/500#issuecomment-5135874808) against this branch? Against a local moto S3 server I measured no wall-clock overhead with `--results-buffer 10`/`25` vs. buffer off (7.42s / 7.89s vs 7.98s baseline, 200 transcripts), but confirmation under real S3 latency would validate the fix.

**CI status:** `check-schema-and-types` is failing as expected — `results_buffer` is persisted on `ScanOptions`/`ScanJobConfig`, so the OpenAPI schema changed. `openapi.json` is regenerated in this PR, but the matching `generated.ts` lives in the ts-mono submodule, which this session can't push to. To go green, a maintainer needs to land the companion ts-mono commit and bump the submodule:

```bash
git submodule update --init src/inspect_scout/_view/ts-mono
(cd src/inspect_scout/_view/ts-mono && pnpm install --frozen-lockfile)
.venv/bin/python scripts/export_openapi_schema.py
# commit generated.ts in ts-mono (merge to its main), then bump the submodule pointer here
```

The diff is just the optional `results_buffer?: number | null` field added to five schemas in `generated.ts`.

---

epatey:
**Companion ts-mono change is now in place** (meridianlabs-ai/ts-mono#480, regenerated `generated.ts`), and this PR's gitlink points at that branch's head — phase 1 of the two-phase landing.

Current expected CI state on this PR:
- `check-schema-and-types` — green (artifacts now in sync)
- `submodule-on-main` — **red by design** until ts-mono#480 merges; this is the "waiting on ts-mono merge" signal, not a defect
- everything else — green (all checks passed on the previous SHA, including both py-test matrices and `js-dist-validation`; the ts-mono delta is types-only so the viewer bundle is unchanged)

**Landing order** (from the land-ts-mono procedure): get this PR reviewed/approved with `submodule-on-main` as the sole red → a human merges ts-mono#480 (no auto-merge there — its merge opens the window where ts-mono main references a not-yet-merged scout schema) → I bump the gitlink to the merged main SHA → this PR goes fully green and can land.

---

PR #554, opened 2026-08-06
fix: stream scan result compaction to disk

herbert-apollo:
Large scans can freeze the host when results are synced. Each result row embeds the scanned transcript and its scan events, so a ~48k transcript scan compacts to multi-GB parquets (a ~15GB buffer dir measured on agentic transcripts), and finalizing one repeatedly put a 48GB workstation into a swap storm. Ctrl-C runs the same sync, so interrupting does not help.

The cause is that `scanner_table` compacts into an in-memory buffer and copies it out as bytes, so each scanner's sync peaks at about twice its compacted parquet size. With `results_buffer` (#550) this happens on every periodic sync, not just at finalize.

`scanner_table` now streams the compacted parquet to an output file instead of returning bytes. Local scan dirs compact straight into the sibling `.tmp` and rename as before, remote ones compact into a local temp file that is streamed to the final key. Peak memory is the existing ~100MB batch accumulation cap regardless of output size.

### DIFF
--- a/src/inspect_scout/_recorder/buffer.py
+++ b/src/inspect_scout/_recorder/buffer.py
@@ -316,9 +316,20 @@ def resolve_success_value(value: bool | None, score: JsonValue | None) -> bool |
 def scanner_table(
     buffer_dir: UPath,
     scanner: str,
+    output_file: str,
     *,
     extra_inputs: list[UPath] | None = None,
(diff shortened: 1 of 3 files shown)


epatey:
@review

epatey:
@review

PR #562, opened 2026-08-13
fix: bump openai to >=3.0.0; pass allow_remote_mcp to inspect_ai bridge tool converters

epatey:
inspect_ai ≥0.3.255 added a required `allow_remote_mcp` param to `tool_from_responses_tool` and `tools_from_anthropic_tools`, breaking mypy and (silently, via `try/except`) dropping tool info from observe events. Pass `True` to preserve recorded MCP tools when reconstructing observe events — matches inspect_ai's own default.

Also bumps the dev-group `openai` floor to >=3.0.0 (openai 3.x types its exceptions against httpx2, which the tests construct directly) and adapts the affected tests.

Verified with `make check` and `tests/observe`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -106,7 +106,9 @@ dev = [
     "huggingface_hub",
     "langsmith",
     "logfire",
-    "openai",
+    # >=3.0.0: openai 3.x types its exceptions against httpx2, which the
+    # tests construct directly (httpx2 arrives transitively via openai)
+    "openai>=3.0.0",
     "pytest-dotenv",
     "pytest-asyncio",
     "mypy",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -67,7 +67,9 @@ dev = [
     "huggingface_hub",
     "langsmith",
     "logfire>=3",
-    "openai>=2.26.0",
+    # >=3.0.0: openai 3.x types its exceptions against httpx2, which the
+    # tests construct directly (httpx2 arrives transitively via openai)
+    "openai>=3.0.0",
     "pytest-dotenv",
     "pytest-asyncio",
     "mypy",
--- a/src/inspect_scout/_observe/providers/anthropic.py
+++ b/src/inspect_scout/_observe/providers/anthropic.py
@@ -157,6 +157,7 @@ class AnthropicProvider:
                 request.get("mcp_servers"),  # MCP servers if present
                 {},  # web_search_providers
                 {},  # code_execution_providers
+                allow_remote_mcp=True,
             )
             # Convert Tool to ToolInfo if needed
             for t in raw_tools:
--- a/src/inspect_scout/_observe/providers/openai.py
+++ b/src/inspect_scout/_observe/providers/openai.py
@@ -396,7 +396,9 @@ class OpenAIProvider:
             from inspect_ai.tool._tool_util import tool_to_tool_info
 
             for tool in request.get("tools", []):
-                tool_or_info = tool_from_responses_tool(tool, {}, {})
+                tool_or_info = tool_from_responses_tool(
+                    tool, {}, {}, allow_remote_mcp=True
+                )
                 # Convert Tool to ToolInfo if needed
                 if isinstance(tool_or_info, ToolInfo):
                     tools.append(tool_or_info)
--- a/tests/observe/test_observe_stream_errors.py
+++ b/tests/observe/test_observe_stream_errors.py
@@ -14,6 +14,7 @@ from dataclasses import dataclass
 from typing import Any, AsyncIterator, Callable, Iterator
 
 import httpx
+import httpx2
 import pytest
 
 
@@ -133,7 +134,7 @@ def _openai_error() -> Exception:
     import openai
 
     return openai.APIConnectionError(
-        message="connection reset", request=httpx.Request("POST", "http://test")
+        message="connection reset", request=httpx2.Request("POST", "http://test")
     )
 
 
@@ -667,7 +668,7 @@ def _openai_moderation_error() -> Exception:
 
     err = openai.APIError(
         "limited access for safety reasons",
-        request=httpx.Request("POST", "http://test"),
+        request=httpx2.Request("POST", "http://test"),
         body={
             "type": "invalid_request_error",
             "code": "invalid_prompt",
@@ -759,7 +760,7 @@ def _openai_bad_request_error() -> Exception:
 
     err = openai.BadRequestError(
         message="blocked by content policy",
-        response=httpx.Response(400, request=httpx.Request("POST", "http://test")),
+        response=httpx2.Response(400, request=httpx2.Request("POST", "http://test")),
         body={
             "type": "invalid_request_error",
             "code": "content_policy_violation",
--- a/tests/view/test_api_v2_search.py
+++ b/tests/view/test_api_v2_search.py
@@ -7,6 +7,7 @@ from unittest.mock import patch
 
 import anthropic
 import httpx
+import httpx2
 import openai
 import pytest
 import pytest_asyncio
@@ -24,6 +25,10 @@ from pydantic import TypeAdapter
 
 def _fake_httpx_response(status_code: int) -> httpx.Response:
     return httpx.Response(status_code, request=httpx.Request("POST", "https://test"))
+
+
+def _fake_httpx2_response(status_code: int) -> httpx2.Response:
+    return httpx2.Response(status_code, request=httpx2.Request("POST", "https://test"))
 
 
 def _base64url(s: str) -> str:
@@ -352,7 +357,7 @@ class TestSearchEndpoint:
             (
                 openai.NotFoundError(
                     "The model `gpt-nope` does not exist or you do not have access to it.",
-                    response=_fake_httpx_response(404),
+                    response=_fake_httpx2_response(404),
                     body=None,
                 ),
                 "The model `gpt-nope` does not exist or you do not have access to it.",
--- a/uv.lock
+++ b/uv.lock
@@ -694,6 +694,24 @@ wheels = [
 ]
 
 [[package]]
+name = "claude-agent-sdk"
+version = "0.2.138"
+source = { registry = "https://pypi.org/simple" }
+dependencies = [
+    { name = "anyio" },
+    { name = "mcp" },
+    { name = "sniffio" },
+]
+sdist = { url = "https://files.pythonhosted.org/packages/5e/8f/07da8bf64606cc06c82a8f478c16509bf332d17da7398d31e1126ec3702b/claude_agent_sdk-0.2.138.tar.gz", hash = "sha256:bb09194a877869fda84ac9980f04dcef2abbb0c7534b4cd555283f0e5dfb827a", size = 319524, upload-time = "2026-08-13T23:43:46.281Z" }
+wheels = [
+    { url = "https://files.pythonhosted.org/packages/0c/e5/c869afe2c8ff4625331ce9b0a62b2602c01418a12fd4ea092515da8c1678/claude_agent_sdk-0.2.138-py3-none-macosx_11_0_arm64.whl", hash = "sha256:2b34fdcdaed8a750521dbd800590c9a411b9b13803425133f38d3afe7c2ce893", size = 87842823, upload-time = "2026-08-13T23:43:49.74Z" },
+    { url = "https://files.pythonhosted.org/packages/19/81/0fe3d111b6c1a3f7bc917fae38797389795aee74a05fa78e0001d679a764/claude_agent_sdk-0.2.138-py3-none-macosx_11_0_x86_64.whl", hash = "sha256:68b973423e20afc39ce69ffc47a8dc498bb4779d3da8d00ec1087ed6a4d748fa", size = 92734215, upload-time = "2026-08-13T23:43:53.685Z" },
+    { url = "https://files.pythonhosted.org/packages/a2/9d/3735711bb98d278773b86ef82a153ae69549d8bef4907acced7fa2994dab/claude_agent_sdk-0.2.138-py3-none-manylinux_2_17_aarch64.whl", hash = "sha256:564a619126e1194b4c5ae9be00505173fa904f7dcf75d10bd891d634d78270ad", size = 96736541, upload-time = "2026-08-13T23:43:57.243Z" },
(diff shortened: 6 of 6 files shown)


epatey:
@i-am-marvin, fix the mypy workflow error.

i-am-marvin:
**Claude finished @epatey's task in 16m 35s** —— [View job](https://github.com/meridianlabs-ai/inspect_scout/actions/runs/31728383516)

---
### Fixed — mypy CI is green again ✅

- [x] Check CI status and download the failing mypy job log
- [x] Diagnose the mypy error
- [x] Fix the errors and push (commit 8bd46f3)
- [x] Confirm CI passes — [Build run](https://github.com/meridianlabs-ai/inspect_scout/actions/runs/31729198233) fully green (`py-mypy`, `py-ruff`, `py-test` on 3.10 & 3.14, plus all other jobs)

### What was wrong

The failure was unrelated to this PR's diff. **openai 3.0.0** (released after main's last green build on Aug 5) switched its HTTP layer from `httpx` to the new [`httpx2`](https://github.com/pydantic/httpx2) package, so its exception constructors are now typed against `httpx2.Request`/`httpx2.Response`. Four test sites construct openai exceptions with `httpx` objects, which mypy now rejects:

- `tests/observe/test_observe_stream_errors.py:136, 670, 762` — `APIConnectionError`/`APIError`/`BadRequestError`
- `tests/view/test_api_v2_search.py:355` — `NotFoundError`

(They passed at runtime — the SDK duck-types — which is why only `py-mypy` failed while `py-test` stayed green.)

### The fix

- Construct openai exceptions with `httpx2.Request`/`httpx2.Response` at those four sites (anthropic 0.121.0 still uses `httpx`, so its sites are unchanged)
- Bump the dev extra floor from `openai>=2.26.0` to `openai>=3.0.0` so `httpx2` is guaranteed present (it arrives transitively via openai) and dev envs match what the tests now require

Note: `pip install` isn't permitted on this runner, so I couldn't run mypy locally — I verified the fix against openai 3.0.0's and httpx2's sources and then confirmed via the PR's CI, which is fully green.

--- · [Branch](https://github.com/meridianlabs-ai/inspect_scout/tree/fix/bridge-allow-remote-mcp)

epatey:
@i-am-marvin, are you still with me? You seem very stuck

i-am-marvin:
**Claude finished @epatey's task in 6m 8s** —— [View job](https://github.com/meridianlabs-ai/inspect_scout/actions/runs/31728994204)

---
### Mypy fix: done — and I owe you an explanation for the "stuck" appearance

- [x] Gather context (reproduced the mypy failure from CI logs)
- [x] Sync branch with `origin/main` (already up to date)
- [x] Fix the mypy errors (`httpx` vs `httpx2` mismatches in tests)
- [x] Verify via CI (couldn't run mypy locally — package installs are blocked in this runner)
- [x] ~~Commit and push~~ — not needed; see below

### What happened

The earlier run wasn't dead, just slow: while I was preparing my commit, it pushed `8bd46f31` ("test: use httpx2 types with openai 3.x exceptions") to this branch at 18:07. I had independently produced a functionally identical fix; when my push was rejected I diffed the two and they match exactly (same four call sites, same `openai>=3.0.0` dev floor bump, only comment wording differs). So I dropped my duplicate rather than adding noise.

### The fix itself

Root cause: CI resolves the `openai>=2.26.0` dev dependency to **openai 3.0.0**, which switched its HTTP layer from `httpx` to `httpx2` — its exception classes now expect `httpx2.Request`/`httpx2.Response`. Four test sites built them with old `httpx` objects (`tests/observe/test_observe_stream_errors.py:136,670,762` and `tests/view/test_api_v2_search.py:355`). The fix constructs `httpx2` objects for openai errors (anthropic still uses `httpx`, so those sites are unchanged) and bumps the openai dev floor to `>=3.0.0` so `httpx2` is always present.

### CI status on `8bd46f31`

✅ **py-mypy (3.10) and py-mypy (3.14) both pass** — the failure you asked about is fixed. ruff, build, schema/types checks are green; py-test (3.10/3.14) still running at time of writing but passed on the previous commit and this change only touches test fixture types.

One leftover from the earlier review, your call: `pyproject.toml` still declares `inspect-ai>=0.3.252` while this PR's code requires ≥0.3.255 (older versions silently drop tool info via the `try/except`). [Fix this →](https://claude.ai/code?q=On%20branch%20fix%2Fbridge-allow-remote-mcp%2C%20bump%20the%20inspect-ai%20floor%20in%20pyproject.toml%20from%20%3E%3D0.3.252%20to%20%3E%3D0.3.255%20because%20the%20allow_remote_mcp%20kwarg%20passed%20in%20src%2Finspect_scout%2F_observe%2Fproviders%2Fanthropic.py%20and%20openai.py%20only%20exists%20in%20inspect_ai%200.3.255%2B&repo=meridianlabs-ai/inspect_scout)
 · [Branch](https://github.com/meridianlabs-ai/inspect_scout/tree/fix/bridge-allow-remote-mcp)

pyproject.toml
```diff
@@ -67,7 +67,9 @@ dev = [
     "huggingface_hub",
     "langsmith",
     "logfire>=3",
-    "openai>=2.26.0",
+    # >=3.0.0: openai 3.x types its exceptions against httpx2, which the
+    # tests construct directly (httpx2 arrives transitively via openai)
+    "openai>=3.0.0",
```
epatey, line 72: I'm not so sure about this. inspect_ai didn't do this. How does inspect_ai deal with this without needing to bump the min version?

epatey:
@auto

i-am-marvin:
**Claude finished @epatey's task in 3m 44s** —— [View job](https://github.com/meridianlabs-ai/inspect_scout/actions/runs/31729701416)

---
### How inspect_ai handles this without bumping min versions

Good question — I went and looked at how inspect_ai actually does it. It uses two different strategies depending on the kind of dependency, and which one applies here is the crux:

**1. Provider SDKs (openai, anthropic, google-genai, …): there is no min version to bump, because they aren't declared dependencies at all.** inspect_ai's [`requirements.txt`](https://github.com/UKGovernmentBEIS/inspect_ai/blob/main/requirements.txt) contains no `openai` or `anthropic` entries. The SDKs are optional, lazily imported, and every entry point enforces the version *at runtime* via [`verify_required_version()`](https://github.com/UKGovernmentBEIS/inspect_ai/blob/main/src/inspect_ai/_util/version.py) — e.g. `model/_providers/providers.py` requires `anthropic >= 0.115.0` and raises an explicit "pip install --upgrade" error if the installed SDK is too old. So the answer to "how does inspect_ai deal with this?" is: the requirement lives in code and **fails loudly**, instead of living in packaging metadata.

**2. Hard (declared) dependencies: inspect_ai bumps the floor, every time.** Its `requirements.txt` is full of exactly the pattern I suggested, comment and all:
- `agent-client-protocol>=0.12.0  # 0.12 renamed UntitledMultiSelectItems and added ...`
- `anyio>=4.14.0  # 4.14 fixes asyncio Lock/Semaphore waiter deadlock after cancellation`
- `click>=8.1.3,!=8.2.0,!=8.2.2,...  # 8.2.2/8.3.0/8.3.1 break optional flag values`

Scout's dependency on `inspect-ai` is the second kind — a required dep in `pyproject.toml` — so the inspect_ai-consistent move here is actually the floor bump.

### That said, the stakes are genuinely low

`release-pin-deps.yml` rewrites the floor to the latest shipped inspect_ai on every release PR, so **published scout wheels can never carry the stale floor**. The only exposure is a source install from `main`, in the window before the next release, with inspect_ai manually held at 0.3.252–0.3.254 — and dev setups here typically track inspect_ai `main` anyway. What tips me toward bumping regardless is that the failure mode is silent: the `TypeError` is swallowed by the surrounding `try/except`, so tool info just quietly disappears (the exact bug this PR fixes). But if that window doesn't concern you, leaving it is defensible — the automation makes it right at release either way. A runtime `verify_required_version` check (option 1's pattern) would be overkill for a hard dep.

### CI status

All green on `8bd46f31` — Build ✅ (mypy 3.10/3.14, py-test 3.10/3.14, ruff, schema/types), PR Title Lint ✅. No code changes made in this run, so nothing was pushed.

---
- [x] Gather context (PR diff, pyproject, provider call sites)
- [x] Research how inspect_ai handles signature changes in its dependencies
- [x] Check scout/inspect_ai precedent for runtime version checks vs floor bumps
- [x] Answer the review question with a recommendation
- [x] Check CI status on the latest commit
 · [Branch](https://github.com/meridianlabs-ai/inspect_scout/tree/fix/bridge-allow-remote-mcp)

epatey:
@i-am-marvin, I'm still confused. IIRC, inspect_ai does have a requirements-dev.txt which gets installed during CI. It did not get bumped. So what's different about scout's consumption of opanai sdk?

i-am-marvin:
**Claude finished @epatey's task in 4m 25s** —— [View job](https://github.com/meridianlabs-ai/inspect_scout/actions/runs/31730300506)

---
### What's different isn't the SDK consumption — it's how CI resolves dependencies

- [x] Check how scout's tests/dev deps consume the openai SDK
- [x] Check how inspect_ai's requirements-dev.txt declares openai and how its tests use the SDK
- [x] Explain the difference

Your recollection is right: inspect_ai's `requirements-dev.txt` has `openai>=2.45.0`, it gets installed in CI, and it wasn't bumped. And here's the kicker — inspect_ai consumes the openai SDK **the same way scout does**: [`tests/model/providers/test_openai_reasoning_summaries.py`](https://github.com/UKGovernmentBEIS/inspect_ai/blob/main/tests/model/providers/test_openai_reasoning_summaries.py) constructs `BadRequestError(message=..., response=httpx.Response(...), body=None)` with plain `httpx` objects, in a file their CI type-checks (`uv run mypy src tests`). Under openai 3.0.0 that fails mypy exactly like scout's tests did.

So why is inspect_ai's CI green (it ran successfully on main today, hours ago)?

**inspect_ai's CI never sees openai 3.0.0.** Its mypy job does `uv pip install .[dev]` but then runs `uv run mypy ...` — and because inspect_ai is a uv-managed project with a committed `uv.lock` (plus `default-groups = ["dev"]` in `[tool.uv]`), `uv run` syncs the venv **to the lockfile** before executing. Their lock pins `openai == 2.45.0`. The `>=2.45.0` floor in requirements-dev.txt isn't what decides the installed version — the lock is. They'll only meet openai 3.0.0 when someone deliberately runs `uv lock --upgrade`, at which point they'll hit this identical mypy break and need the same httpx2 fix. "Didn't bump" means "hasn't crossed openai 3 yet," not "found a way around it."

**Scout's CI resolves fresh from PyPI every run.** `build.yaml` does `python -m pip install .[dev]` with no lock involved (scout has a `uv.lock`, pinning openai 2.32.0, but CI never consults it — and CLAUDE.md forbids `uv sync`/`uv run` for dev envs). So the day openai 3.0.0 shipped, `openai>=2.26.0` resolved to 3.0.0 and mypy broke with zero changes on our side.

### Why the floor bump was still the right fix for scout

With no lockfile in scout's CI/dev path, the `>=` floor is the *only* mechanism ensuring the installed openai matches what the tests assume. The tests now `import httpx2` and build openai exceptions with it — that only type-checks against openai 3.x. Left at `>=2.26.0`, any environment resolving to openai 2.x would fail mypy in the opposite direction. In inspect_ai the floor is documentation of the minimum the code needs while the lock does the pinning; in scout the floor has to do both jobs.

(There's a legitimate separate question of whether scout's CI *should* install from its lockfile like inspect_ai effectively does — that would insulate CI from day-one upstream releases, at the cost of the early-drift detection that made this PR catch the openai 3 break immediately. Happy to explore that if you want, but it's an infra decision beyond this PR.)
 · [Branch](https://github.com/meridianlabs-ai/inspect_scout/tree/fix/bridge-allow-remote-mcp)

epatey:
@i-am-marvin, I've spent quite some time on this. We ultimately decided to bump inspect_ai's min for openai to 3.00. That is now the appropriate path for us to take in this pr as well.

i-am-marvin:
**Claude finished @epatey's task in 13m 1s** —— [View job](https://github.com/meridianlabs-ai/inspect_scout/actions/runs/31806358904)

---
### Done — the openai ≥3.0.0 path is now complete across the repo ✅

- [x] Gather context (pyproject, uv.lock, branch state vs main — branch already contains `origin/main`)
- [x] Confirm/complete the `openai>=3.0.0` floor bump everywhere it's declared
- [x] Push (commit `9e17a08b`) and verify CI — **Build fully green** (mypy 3.10/3.14, py-test 3.10/3.14, ruff, build, schema/types)
- [x] Resolve the settled review thread on pyproject.toml

### What was done

The pip-facing half was already in place: the `[project.optional-dependencies]` dev extra got `openai>=3.0.0` in commit `8bd46f31`, which is what CI installs. What remained was the uv-managed half, which I've now aligned in `9e17a08b`:

- **`[dependency-groups]` dev**: `openai` (unpinned) → `openai>=3.0.0`, same comment as the dev extra
- **`uv.lock` relocked**: openai 2.32.0 → 3.0.0, adding `httpx2` (openai 3.x's HTTP layer). Without this, a uv-managed dev env would keep openai 2.32.0, where the httpx2-typed tests can't even import.

### One relock side effect worth knowing about

Bumping openai to 3.x forces two backtracks in the lock: **litellm 1.90.0 → 1.83.0** (newer litellm caps openai below 3.x; 1.83.0 has no cap) and **harbor 0.16.1 → 0.6.1** (harbor is the only litellm consumer; newer harbor needs the newer litellm). This isn't uv being weird — I verified a fresh `pip install .[dev]` resolves to exactly the same versions (openai 3.0.0, harbor 0.6.1, litellm 1.83.0), so this is what CI has been installing since the openai floor landed, and py-test is green with it. The lock now just tells the truth. When the harbor/litellm ecosystem catches up to openai 3.x, a routine `uv lock` will move them forward again.

### Not touched

The other open review thread (`anthropic.py:160`, suggesting `inspect-ai>=0.3.255`) is about a different dependency and wasn't part of this ask, so I left it unresolved — your call, and `release-pin-deps.yml` bumps that floor at release regardless.

*Note: local `mypy`/`pytest` invocations are blocked by this runner's permissions, so verification is via the PR's CI, which completed fully green on `9e17a08b`.*

--- · [Branch](https://github.com/meridianlabs-ai/inspect_scout/tree/fix/bridge-allow-remote-mcp)

PR #564, opened 2026-08-17
fix: prevent deferred metrics write to closed active scans store

### LINKED ISSUE #558: active_scans_store: throttled put_metrics can fire after the kvstore connection is closed (sqlite3.ProgrammingError)
✨ claude-opus-5 | Comment not reviewed by tbroadley

## Summary

`active_scans_store()` yields a store object whose methods close over a `KVStore` owned by the context manager's `with` block. `_scan.py` passes a callback that writes through that store into `@throttle(1)`, which can defer the call to a background task. When that trailing-edge fire lands after the `with` block has exited, the sqlite connection is already closed and the write raises:

```
sqlite3.ProgrammingError: Cannot operate on a closed database.
```

Surfaced when running tests in parallel workers, where the extra contention widens the window between the deferred write being scheduled and the block exiting. It is not a cross-process race on the shared DB file — the connection being used after close belongs to the same process that closed it.

## Mechanism

- `_recorder/active_scans_store.py` — `active_scans_store()` is a `@contextmanager` that does `with inspect_kvstore(_STORE_NAME) as kvstore:` and then `yield _Store()`. Every `_Store` method (`put_metrics`, `put_spec`, `put_scanner_results`, `read_all`, …) closes over that `kvstore`. On exit, `KVStore.__exit__` calls `self.conn.close()`, but the yielded `_Store` can still be referenced.
- `_scan.py` — inside `with active_scans_store() as active_store:`, `update_metrics` is defined as a plain function calling `active_store.put_metrics(...)` and passed to `strategy(..., update_metrics=update_metrics)`.
- `_concurrency/single_process.py` — the strategy wraps it: `@throttle(1) def _update_metrics(): ... update_metrics(metrics)`.
- `inspect_ai/util/_throttle.py` — a call inside the throttle window is stored as `pending` and scheduled with `run_in_background(_deferred_fire)`. That task is fire-and-forget: it is not awaited, and nothing ties its lifetime to the `active_scans_store` block. `_deferred_fire` sleeps out the remaining window and then invokes the wrapped function.

So a metrics update issued shortly before the scan finishes can fire up to ~1s later, after `active_scans_store()` has closed the connection.

## Reproduction

Deterministic, using only public `inspect_scout` APIs plus the throttle decorator the scan path itself uses:

```python
"""Minimal reproduction: a throttled writer outliving the active_scans_store block."""

import anyio
from inspect_ai.util._throttle import throttle
from inspect_scout._concurrency.common import ScanMetrics
from inspect_scout._recorder.active_scans_store import active_scans_store


async def main() -> None:
    with active_scans_store() as store:

        @throttle(1)
        def update_metrics() -> None:
            store.put_metrics("scan-id", ScanMetrics())

        # Two calls inside the window: the first fires now, the second is saved as
        # pending and scheduled on a background task that sleeps out the window.
        update_metrics()
        update_metrics()

    # The with-block has exited, so KVStore.__exit__ has closed the sqlite connection.
    # The deferred trailing-edge fire is still queued.
    await anyio.sleep(2)


if __name__ == "__main__":
    anyio.run(main)
```

Output (3/3 runs):

```
Task exception was never retrieved
future: <Task finished name='Task-2' coro=<run_in_background.<locals>.wrapper() done, ...> exception=RuntimeError('Exception escaped from background task')>
Traceback (most recent call last):
  File ".../inspect_ai/_util/background.py", line 34, in wrapper
    await func(*args)
  File ".../inspect_ai/util/_throttle.py", line 67, in _deferred_fire
    last_result = func(*pending[0], **pending[1])
  File "repro.py", line 19, in update_metrics
    store.put_metrics("scan-id", ScanMetrics())
  File ".../inspect_scout/_recorder/active_scans_store.py", line 138, in put_metrics
    existing = json.loads(kvstore.get(pid_key) or "{}")
  File ".../inspect_ai/_util/kvstore.py", line 57, in get
    cursor = self.conn.execute("SELECT value FROM kv_store WHERE key = ?", (key,))
sqlite3.ProgrammingError: Cannot operate on a closed database.
```

In a real scan the same call arrives through `single_process._update_metrics` → `_scan.py:update_metrics` → `put_metrics`, so the failing frames below `_deferred_fire` are identical.

## Versions

- `inspect-scout` 0.4.46 (also seen on 0.4.45)
- `inspect-ai` 0.3.257 (also seen on 0.3.251)
- Python 3.13.14, Linux

## Suggested fix direction

Any of these would close it; the first seems the most contained:

1. **Flush or cancel the throttled write before the store goes away.** `_throttle` currently offers no handle to do this, so it would need to expose something like `.flush()` / `.cancel()` on the wrapped function, with `_scan.py` calling it in the `finally:` that already runs `active_store.delete_current()`.
2. **Decouple the store's lifetime from the connection** — have `active_scans_store` open a short-lived connection per operation instead of closing over one, so a late write is merely a no-op-ish extra open rather than an error.
3. **Make the write tolerant of a closed connection** — cheapest, but it hides late writes rather than ordering them, and the metrics they carry are silently dropped.

Worth noting independently: `run_in_background`'s wrapper re-raises as `RuntimeError("Exception escaped from background task")`, so whatever the resolution, a failure in a deferred throttle callback surfaces as an unretrieved task exception rather than failing the scan.

sunnyadn:
Fixes #558.

**What happens**

The scan reports metrics through a callback that the single process strategy wraps in `@throttle(1)` with trailing-edge semantics. A call arriving inside the throttle window is saved and fired up to a second later by a background task in the scan's outer task group, so the write can outlive the `active_scans_store()` block that owns the sqlite connection.

The path where this actually fires in a real scan is the interrupted one. When a scanner error ends the scan, the store's connection closes and its entry is deleted, then `handle_scan_interrupted` still awaits a final `recorder.sync()`. If the pending trailing-edge write comes due during that sync, `put_metrics` runs against the closed connection and raises `sqlite3.ProgrammingError: Cannot operate on a closed database`. Past the connection error it would also resurrect the deleted entry, because `put_metrics` starts from `kvstore.get(pid_key) or "{}"`.

**The fix**

`update_metrics` now checks a `scan_finished` flag set in the same `finally` that deletes the store entry. Nothing awaits between setting the flag and the delete, so a deferred write cannot land in the gap. The `finally` runs on the success, error, and cancellation paths alike, so the guard covers all three. This is a more contained version of the issue's first suggested direction. Rather than adding a flush or cancel handle to the throttle in inspect_ai, the callback itself refuses to write once the scan is over. The callback is also the one choke point that covers both targets a late fire would hit, the store write and the display update, which is why the guard sits here rather than inside `put_metrics`.

**Verification**

Reproduced end to end through the public `scan()` API. The new test holds the interrupted-path sync open past the throttle window with a monkeypatch, so the pending write always comes due in the vulnerable gap, with no wall-clock aiming. On current main it fails 3 of 3 runs. The surfaced error varies because `_scan_async` re-raises with `from None` and severs the cause chain. Two runs showed the ProgrammingError above directly, and one showed only a downstream `RuntimeError: no running event loop` from the crashed scan's teardown. The root is the same deferred write either way, and it is the only background task in the group. With the fix the test passes 3 of 3 runs and the full suite passes.

I also probed the other two paths with instrumented runs. On successful completion nothing yields between the entry delete and the task group exit, and the pending timer is cancelled with the task group about a second before it would fire (5 of 5 runs). Under a real SIGINT the global cancellation kills the timer at unwind time (3 of 3 runs). So the interrupted path is the only one I could make fire, and the guard covers the others as insurance.

One thing I did not check is remote (S3) transcript sources, where reader shutdown does real async work and might open a success-path window too. The guard would cover that case the same way.

cc @epatey

### DIFF
--- a/src/inspect_scout/_scan.py
+++ b/src/inspect_scout/_scan.py
@@ -756,7 +756,15 @@ async def _scan_async_inner(
                         )
                         maybe_start_results_sync()
 
+                    # metrics reporting is throttled, so a trailing-edge write can
+                    # be deferred past the end of the scan. by then the store's
+                    # connection is closed and its entry has been deleted, so the
+                    # write must not run at all.
+                    scan_finished = False
+
                     def update_metrics(metrics: ScanMetrics) -> None:
+                        if scan_finished:
+                            return
                         active_store.put_metrics(scan.spec.scan_id, metrics)
                         scan_display.metrics(metrics)
 
@@ -792,6 +800,9 @@ async def _scan_async_inner(
                             await recorder.location(), complete=len(errors) == 0
                         )
                     finally:
+                        # nothing awaits between here and the delete, so a deferred
+                        # metrics write cannot land in the gap
+                        scan_finished = True
                         active_store.delete_current()
 
         # report scan complete
--- a/src/inspect_scout/_scan.py
+++ b/src/inspect_scout/_scan.py
@@ -756,10 +756,9 @@ async def _scan_async_inner(
                         )
                         maybe_start_results_sync()
 
-                    # metrics reporting is throttled, so a trailing-edge write can
-                    # be deferred past the end of the scan. by then the store's
-                    # connection is closed and its entry has been deleted, so the
-                    # write must not run at all.
+                    # the store ignores writes once closed (see _Store); this
+                    # guard additionally keeps a late trailing-edge fire away
+                    # from the finished display.
                     scan_finished = False
 
                     def update_metrics(metrics: ScanMetrics) -> None:
@@ -800,8 +799,6 @@ async def _scan_async_inner(
                             await recorder.location(), complete=len(errors) == 0
                         )
                     finally:
-                        # nothing awaits between here and the delete, so a deferred
-                        # metrics write cannot land in the gap
                         scan_finished = True
                         active_store.delete_current()
 
--- a/tests/scan/test_active_scan_cleanup.py
+++ b/tests/scan/test_active_scan_cleanup.py
@@ -1,15 +1,32 @@
-import os
+import time
+from contextlib import contextmanager
 from pathlib import Path
+from typing import Any, Callable, Iterator
 
 import anyio
+import inspect_scout._scan as scan_module
 import pytest
+from inspect_ai.util import throttle
 from inspect_scout import Result, Scanner, Status, scan, scanner
-from inspect_scout._recorder.active_scans_store import active_scans_store
+from inspect_scout._concurrency import single_process
+from inspect_scout._recorder.active_scans_store import (
+    ActiveScansStore,
+    active_scans_store,
+)
 from inspect_scout._recorder.file import FileRecorder
 from inspect_scout._transcript.factory import transcripts_from
 from inspect_scout._transcript.types import Transcript
 
+from tests.helpers import temp_active_scans_store
+
 LOGS_DIR = Path(__file__).parent.parent.parent / "examples" / "scanner" / "logs"
+
+# the strategy hardcodes @throttle(1); shrinking the window keeps the same
+# trailing-edge mechanics while the test holds the vulnerable gap open for
+# fractions of a second instead of multiples of it. the window must still be
+# generous enough that the strategy's teardown metrics call lands inside it
+# even on a loaded CI machine, or no trailing-edge fire gets scheduled.
+THROTTLE_WINDOW = 0.3
 
 
 @scanner(name="cleanup_probe_scanner", messages="all")
@@ -22,52 +39,35 @@ def cleanup_probe_scanner() -> Scanner[Transcript]:
 
 @scanner(name="failing_probe_scanner", messages="all")
 def failing_probe_scanner() -> Scanner[Transcript]:
-    first_transcript = True
-
     async def scan_transcript(transcript: Transcript) -> Result:
-        nonlocal first_transcript
-        if first_transcript:
-            first_transcript = False
-            return Result(value=True)
-        # let the first result get recorded before failing (the test docstring
-        # explains why this ordering leaves a trailing write pending)
-        await anyio.sleep(0.3)
         raise ValueError("injected scanner failure")
 
     return scan_transcript
 
 
-def assert_no_active_scan_entry() -> None:
-    """The active-scans entry for this process must be gone after a scan.
+def test_completed_scan_leaves_no_active_scan_entry(tmp_path: Path) -> None:
+    """Cleanup smoke test for the success path.
 
-    Checked by PID rather than by an empty store: under pytest-xdist other
-    worker processes have live entries of their own.
+    A completed scan must delete its active-scans entry. (The late
+    metrics-write hazard is only reachable on the interrupted path -- on
+    completion the pending trailing-edge write is cancelled with the scan's
+    task group -- so this test pins only the cleanup behavior.)
     """
-    with active_scans_store() as store:
-        assert store.read_by_pid(os.getpid()) is None
+    with temp_active_scans_store():
+        status = scan(
+            scanners=[cleanup_probe_scanner()],
+            transcripts=transcripts_from(LOGS_DIR),
+            scans=str(tmp_path),
+            limit=2,
+            # single-process path: the strategy whose metrics callback and
+            # store cleanup are under test here
+            max_processes=1,
+            display="none",
+        )
 
-
-def test_scan_completes_with_a_metrics_write_still_throttled(tmp_path: Path) -> None:
-    """A scan short enough to end while a throttled metrics write is pending.
-
-    The scan issues its metrics updates within a few tens of milliseconds and
-    finishes well inside the one second throttle window, so the trailing-edge
-    write is still pending when the scan ends. It must not fire against the
-    closed store, and the scan must leave no active-scans entry behind.
-    """
-    status = scan(
-        scanners=[cleanup_probe_scanner()],
-        transcripts=transcripts_from(LOGS_DIR),
-        scans=str(tmp_path),
-        limit=2,
-        # the deferred write only exists on the single-process path, where
-        # metrics reporting goes through a throttled callback
-        max_processes=1,
-        display="none",
-    )
-
-    assert status.complete
-    assert_no_active_scan_entry()
+        assert status.complete
+        with active_scans_store() as store:
+            assert store.read_all() == {}
 
 
 def test_interrupted_scan_survives_a_late_metrics_write(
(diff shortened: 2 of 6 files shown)


tests/scan/test_active_scan_cleanup.py
```diff
@@ -0,0 +1,107 @@
...
+    """The active-scans entry for this process must be gone after a scan.
+
+    Checked by PID rather than by an empty store: under pytest-xdist other
+    worker processes have live entries of their own.
+    """
+    with active_scans_store() as store:
+        assert store.read_by_pid(os.getpid()) is None
+
+
+def test_scan_completes_with_a_metrics_write_still_throttled(tmp_path: Path) -> None:
```
epatey: **[code-review] Test passes with or without the fix (vacuous as a regression test).** Verified empirically: with the `scan_finished` guard reverted, this test stayed green 4/4 runs (the interrupted-path test does fail without the fix). Mechanism: the throttle's `_deferred_fire` sleeps inside the scan-level task group (`set_background_task_group(tg)`), and on the completed path `run`'s `finally: tg.cancel_scope.cancel()` cancels it before the 1s window expires. Even if finalization awaits (`record_metrics`/`errors`/final `sync`) stretch past the window, the fire lands before `delete_current()` against the still-open store; after close there is no await before the cancel. The docstring's "must not fire against the closed store" scenario can't occur on this path. Suggest dropping it or reframing as a plain cleanup smoke test (its docstring currently claims coverage it doesn't provide).

tests/scan/test_active_scan_cleanup.py
```diff
@@ -0,0 +1,107 @@
...
+        # metrics reporting goes through a throttled callback
+        max_processes=1,
+        display="none",
+    )
+
+    assert status.complete
+    assert_no_active_scan_entry()
+
+
+def test_interrupted_scan_survives_a_late_metrics_write(
```
epatey, line 69: **[code-review] Detection is indirect: the test's own assertions never see the bug.** Verified empirically: without the guard, `put_metrics` raises `sqlite3.ProgrammingError` at `kvstore.get` on the closed connection, which propagates out of the deferred background task and makes `scan()` itself raise — so `assert not status.complete` and `assert_no_active_scan_entry()` never execute. They would also pass if they did: the `get` fails before any `put`, so resurrection is unreachable on this path and the store entry is already deleted. The test therefore guards the fix only via `run_in_background`'s exception-propagation behavior; its asyncio fallback (no task group set) merely logs orphan-task exceptions, so a change there would turn this test silently green. Consider asserting the failure mode directly (e.g. wrap scan() and assert no exception escapes, with a comment naming the dependency).

src/inspect_scout/_scan.py
```diff
@@ -756,7 +756,15 @@ async def record_results(
                         )
                         maybe_start_results_sync()
 
+                    # metrics reporting is throttled, so a trailing-edge write can
+                    # be deferred past the end of the scan. by then the store's
+                    # connection is closed and its entry has been deleted, so the
+                    # write must not run at all.
+                    scan_finished = False
```
epatey, line 762: **[code-review] Consider moving the guard into the store (closes the whole class, not one call site).** Every `_Store.put_*` in `active_scans_store.py` does `kvstore.get(pid_key) or "{}"` then `put`, so any late write raises on the closed connection — and would resurrect a deleted entry if the connection were open. A `closed` flag set in `delete_current()`/context exit making `put_*` no-ops protects all writers in one place and matches the existing `self._closed` idiom in `_transcript/eval_log.py:111-131`; it also removes this flag plus the comments documenting a callee's throttling two layers down (the deferred fire lives in `single_process.py`'s `@throttle(1)` from `inspect_ai.util`). Counterpoint acknowledged: the current flag also suppresses the late `scan_display.metrics` call, which a store-side guard wouldn't. Minor: the "nothing awaits between here and the delete" comment isn't load-bearing — the deferred fire only runs at an await point, and it would see the flag even if an await were inserted there.

tests/scan/test_active_scan_cleanup.py
```diff
@@ -0,0 +1,107 @@
...
+    write would raise (and resurrect the deleted entry if it got further).
+
+    Holding the interrupted-path sync open past the one second throttle window
+    makes the write come due inside that gap deterministically, no wall-clock
+    aiming involved.
+    """
+    real_sync = FileRecorder.sync
+
+    async def slow_sync(scan_location: str, complete: bool) -> Status:
+        await anyio.sleep(2)
```
epatey: **[code-review] ~2.3s of fixed sleep per suite run is avoidable.** The 2s here (and the 0.3s in `failing_probe_scanner`) exist only to straddle the hardcoded 1s throttle window. `single_process.py` imports `throttle` by name and applies `@throttle(1)` at scan time, so `monkeypatch.setattr(inspect_scout._concurrency.single_process, "throttle", ...)` with a ~0.05-0.1s window before `scan()` takes effect; both sleeps then shrink proportionally (keep generous ratios, e.g. window 0.05s / sync gap 0.3-0.5s), saving ~2s per run with the same determinism. The 0.3s sleep is also not needed for the assertions as written: workers call `_update_metrics` on every action (and again in the worker/strategy `finally` during unwind), so a pending trailing write exists even with an unconditionally-raising scanner.

tests/scan/test_active_scan_cleanup.py
```diff
@@ -0,0 +1,107 @@
...
+            return Result(value=True)
+        # let the first result get recorded before failing (the test docstring
+        # explains why this ordering leaves a trailing write pending)
+        await anyio.sleep(0.3)
+        raise ValueError("injected scanner failure")
+
+    return scan_transcript
+
+
+def assert_no_active_scan_entry() -> None:
```
epatey: **[code-review] Reuse temp-store isolation instead of PID-scoped reads against the real store.** `tests/recorder/test_active_scans_store.py:32-43` already has `_temp_store_name()` (patches `_STORE_NAME` to a random temp kvstore and cleans up). `active_scans_store()` reads `_STORE_NAME` at call time and these tests run in-process (`max_processes=1`), so wrapping the `scan()` call in it works; the assertion becomes `read_all() == {}` with no xdist caveat. As written, the tests write live entries into the developer's real `scout_active_scans` store (transiently visible to a running scout view) and run its version-check/stale-PID sweep against real data. Suggest hoisting `_temp_store_name` to `tests/helpers.py` and using it here.

epatey:
**[code-review]** Two pre-existing issues found while tracing this fix's mechanism — outside the diff, so noting them here rather than inline:

1. **`set_batch_log_callback` is never reset** (`src/inspect_scout/_concurrency/single_process.py:118`). The strategy's `finally` resets only `set_batch_status_callback(None)`. inspect_ai's `log_batch` dispatches to `(_batch_log_callback or print)`, so after any scout scan the no-op `_on_batch_log` stays installed process-wide and all subsequent evals/scans in that process have batch log output silently suppressed. Same callback-lifetime bug class this PR fixes; fix is a one-liner `set_batch_log_callback(None)` in the same `finally`.

2. **MP worker subprocesses drop their final metrics update** (`src/inspect_scout/_concurrency/_mp_subprocess.py`). Subprocesses never call `set_background_task_group`, so the throttle's deferred fire falls back to a bare `asyncio.create_task`; the worker sends `WorkerComplete` and its loop tears down (anyio cancels the still-sleeping task), so the final zeroed metrics never reach the parent. Cosmetic — stale per-worker metrics in the display at worker exit — but the same trailing-fire-outlives-scope pattern.

epatey (changes requested):
See comments above

tests/scan/test_active_scan_cleanup.py
sunnyadn: Agreed, this matches what the PR description found on the success path, and the docstring oversold it. Reframed as a cleanup smoke test pinning that a completed scan deletes its entry, which isn't covered elsewhere. Happy to drop it instead if you prefer.

tests/scan/test_active_scan_cleanup.py
sunnyadn: Went a step further than wrapping the call. The test now records throttled fires and the entry delete, and asserts a fire came due after the delete, so it reddens if the window stops opening instead of going silently green. Note the revert check now needs both guards removed, since with only the callback guard reverted the store guard keeps it green (verified both ways).

src/inspect_scout/_scan.py
sunnyadn: Done, with one deviation. The flag flips on context exit only and delete_current() stays pure per its Protocol doc, and nothing awaits between the delete and the exit so the hazard is still covered. scan_finished stays for the display suppression, and the "nothing awaits" comment is gone.

tests/scan/test_active_scan_cleanup.py
sunnyadn: Done. throttle is monkeypatched to a 0.3s window with the sync hold at 2x, dropping fixed sleeping from ~2.3s to 0.6s, and the scanner now raises unconditionally as you noted. Kept the window generous so the teardown metrics call still lands inside it on a loaded CI machine.

tests/scan/test_active_scan_cleanup.py
sunnyadn: Done. Hoisted to tests/helpers.py as temp_active_scans_store (composed from the existing temp_kvstore), and both tests now assert read_all() == {} against a throwaway store.

sunnyadn:
Thanks for tracing both. The set_batch_log_callback reset is included here since it's the same finally, happy to split it out if you'd rather keep this PR minimal. Opening a separate issue for the MP worker metrics drop so it doesn't get lost.

epatey (approved):
Thank you

PR #565, opened 2026-08-17
fix(examples/openclaw_telemetry_hal): accept 'explicit' session kind as an orchestrator surface

### LINKED ISSUE #560: openclaw_telemetry_hal example source: unrecognized session kind 'explicit' (gateway-fallback sessions) refuses import of otherwise-valid telemetry
*Filed by Claude Code on behalf of @alexandraabbas — Alexandra reviewed the contents below.*

We're using the `examples/sources/openclaw_telemetry_hal` worked example (vendored into our project per its stated copy-and-adapt purpose) to import OpenClaw telemetry from long-horizon research-agent runs.

One of our real transcripts fails to import:

```
ValueError: OpenClaw telemetry 'agent.start' event carries sessionKey
'agent:main:explicit:gateway-fallback-68ae5a27-cd80-4a49-974c-746166fc51a5'
with unrecognized session kind 'explicit' (known kinds: telegram, main,
dashboard, cron, subagent). This may be an OpenClaw surface the importer
does not yet support; refusing to import rather than silently dropping
this session's activity.
```

Characterisation of the failing file: ~18k agent-keyed events, of which 88 belong to 2 sessions with kind `explicit`, both keyed `agent:main:explicit:gateway-fallback-<uuid>` — this looks like OpenClaw's gateway-fallback path opening explicitly-created sessions on the main agent. The fail-fast is doing exactly what it says on the tin (and we appreciate that it refuses rather than silently drops); the ask is to teach the example about this surface — presumably classifying `explicit` on the orchestrator agent as an orchestrator kind, though it may deserve a look at what gateway-fallback sessions actually contain first.

We're happy to add this ourselves and PR it if you'd prefer not to — filing so it's tracked and in case the right classification is obvious to someone closer to OpenClaw's session model.

(Related, minor: we'll raise a separate PR with some patches we've come across while working with the ingestor — e.g. our vendored copy adds `cron` to `ORCH_KINDS` (CRUX runs are cron-scheduled) and maps `message.in` events to operator-sourced user messages. That's also why the error message above already lists `cron` — stock upstream would say `telegram, main, dashboard, subagent`.)

alexandraabbas:
*Opened by Claude Code on behalf of @alexandraabbas.*

Closes #560

## What

The `examples/sources/openclaw_telemetry_hal` importer's unrecognized-kind fail-fast refused telemetry containing sessionKey kind `explicit` (e.g. `agent:main:explicit:gateway-fallback-<uuid>`), failing the whole import. This PR classifies `explicit` as an orchestrator kind (`ORCH_KINDS`), following a characterisation of the real failing data (below). Genuinely unknown kinds keep failing loudly — the example's refuse-rather-than-silently-drop philosophy is unchanged.

## Characterisation of the failing data (structural fields only)

In the failing transcript (~18k agent-keyed events), 88 events belong to 2 sessions with kind `explicit`, both keyed `agent:main:explicit:gateway-fallback-<uuid>`:

- Event types: 4 × `agent.start`, 42 × `tool.start`, 42 × `tool.end` (no `agent.end`).
- **All 88 carry `agentId: "main"`** — identical to the kind-`main` sessions' events.
- Tool vocabulary is ordinary main-agent tooling: `exec` ×28, `cron` ×5, `read` ×2, `subagents` ×2, `sessions_list` ×2, `edit`/`write`/`process` ×1 each; `tool.end` `success` 40 true / 2 false.
- The `agent.start` events carry a spawn-style `prompt` (`promptLength: 674`) and an **empty `messages[]`** — no assistant/user/toolResult roles recorded under the explicit keys in this capture.
- Cross-check against orchestrator `agent.*` snapshots: 5 of the 42 explicit `tool.start` calls also appear as identical `(name, arguments)` toolCalls inside non-explicit orchestrator snapshots; the rest exist only on the `tool.*` channel — i.e. these sessions look like the always-on timing channel of an orchestrator surface, not delegated sub-agent work.

Illustrative event shapes (structural fields only, content omitted):

```json
{"type": "agent.start", "agentId": "main", "sessionKey": "agent:main:explicit:gateway-fallback-<uuid>", "prompt": "<omitted>", "promptLength": 674, "messages": []}
{"type": "tool.start",  "agentId": "main", "sessionKey": "agent:main:explicit:gateway-fallback-<uuid>", "toolName": "exec", "params": {"...": "..."}}
{"type": "tool.end",    "agentId": "main", "sessionKey": "agent:main:explicit:gateway-fallback-<uuid>", "toolName": "exec", "durationMs": 1234, "success": true, "error": null}
```

**Conclusion**: these are main-agent (orchestrator) activity reaching OpenClaw via a gateway-fallback path that opens an explicitly-created session — so `explicit` on the orchestrator agent is an orchestrator kind. Any assistant turns recorded under an `explicit` key are attributed as orchestrator turns; like the other orchestrator surfaces, the `tool.*` channel on these sessions remains timing-only (not reconstructed into the message thread), consistent with how `main`/`telegram`/`dashboard` are treated today.

## Changes

- `detection.py`: add `"explicit"` to `ORCH_KINDS`, with a comment documenting the observed gateway-fallback shape.
- `parse.py`: module docstring updated to list the kind.
- `tests/test_telemetry_hal.py`: new `test_explicit_kind_is_orchestrator` — an `explicit`-kind session's assistant turns join `orchestrator_turns`, it is not reconstructed as a sub-agent, and its `tool.*` events are accepted. The existing parametrized unrecognized-kind test (`discord`, missing sessionKey) still passes unchanged, covering the still-refuses path.

`pytest examples/sources/openclaw_telemetry_hal/tests/`: 70 passed. Verified end-to-end against the real failing transcript: the `explicit` refusal is gone and the file imports.

## Note on a sibling PR

A separate PR from the `openclaw-hal-cron-operator` branch ports our vendored `cron`-kind + `message.in` patches to this same example. The two are intentionally independent (both based on `main`) but may conflict trivially on adjacent lines in `detection.py`/`parse.py` — whichever lands second needs a one-line rebase. (For full-fidelity import of the transcript behind #560 you need both: it contains `cron`-kind sessions too.)

### DIFF
--- a/examples/sources/openclaw_telemetry_hal/detection.py
+++ b/examples/sources/openclaw_telemetry_hal/detection.py
@@ -12,11 +12,15 @@ import json
 
 # Session-kind classification (sessionKey = agent:<...>:<kind>:<uuid>).
 # ``main``/``telegram``/``dashboard`` are orchestrator surfaces (terminal, the
-# Telegram channel, and the web dashboard respectively); ``subagent`` is
-# delegated work. Any other kind on a consumed event (a surface we have not
+# Telegram channel, and the web dashboard respectively); ``explicit`` is an
+# explicitly-created session on the orchestrator agent — observed as
+# ``agent:main:explicit:gateway-fallback-<uuid>`` when OpenClaw's
+# gateway-fallback path opens a session for the main agent (same ``agentId``
+# as the ``main`` sessions, ordinary main-agent tool activity); ``subagent``
+# is delegated work. Any other kind on a consumed event (a surface we have not
 # seen, e.g. another chat channel) fails the import loudly — see
 # ``_consolidate`` — rather than silently dropping that session's turns.
-ORCH_KINDS = ("telegram", "main", "dashboard")  # the orchestrator
+ORCH_KINDS = ("telegram", "main", "dashboard", "explicit")  # the orchestrator
 SUBAGENT_KIND = "subagent"
 
 _REDACTED = "[REMOVED]"
--- a/examples/sources/openclaw_telemetry_hal/parse.py
+++ b/examples/sources/openclaw_telemetry_hal/parse.py
@@ -20,9 +20,11 @@ JSONL, one event per line. Payload-bearing events:
 - ``tool.start`` / ``tool.end`` are individual tool calls (the only place
   schema-B sub-agent activity is recorded).
 - ``sessionKey`` is ``agent:<...>:<kind>:<uuid>`` with ``kind`` one of ``main``,
-  ``telegram``, ``dashboard`` (orchestrator surfaces) or ``subagent``
-  (delegated work). Any other kind on a consumed event fails the import
-  loudly rather than silently dropping that session's activity.
+  ``telegram``, ``dashboard``, ``explicit`` (orchestrator surfaces — ``explicit``
+  is an explicitly-created session on the orchestrator agent, e.g. OpenClaw's
+  gateway-fallback path) or ``subagent`` (delegated work). Any other kind on a
+  consumed event fails the import loudly rather than silently dropping that
+  session's activity.
 
 Intentionally NOT consumed: ``message.in`` / ``message.sending`` /
 ``message.out`` (channel I/O). A sub-agent's final report is auto-announced as a
--- a/examples/sources/openclaw_telemetry_hal/tests/test_telemetry_hal.py
+++ b/examples/sources/openclaw_telemetry_hal/tests/test_telemetry_hal.py
@@ -251,6 +251,52 @@ class TestParse:
         with pytest.raises(ValueError, match=f"kind {expected_kind!r}"):
             parse_telemetry([*self._orch_raw("agent:main:main:s1"), event])
 
+    def test_explicit_kind_is_orchestrator(self) -> None:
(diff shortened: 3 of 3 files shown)


craig-meridian (approved):
Thanks — this is a model example of how the fail-fast was meant to be used: refuse, characterise, then classify with evidence. The characterisation in the PR body is much appreciated. Approving; happy for this to be merged as-is.

A few entirely optional nits, fine as a follow-up (or never):

- A couple of accepted-kind enumerations elsewhere are now stale: `README.md` ("a `sessionKey` kind outside `main`/`telegram`/`dashboard`/`subagent`…") and the comment on `test_unrecognized_session_kind_fails_with_meaningful_error`. Saying "orchestrator kinds" instead of repeating the list would stop these drifting.
- `explicit` sessions' trailing token is now eligible for the first-id-wins `session_id` selection in `_consolidate`, so file order could pick `gateway-fallback-<uuid>` as the transcript identity in a mixed capture. Harmless, but worth a line in the README's "Transcript identity" section, which enumerates id semantics per surface.
- OpenClaw's docs also use `explicit` for model-run probe sessions (`agent:*:explicit:model-run-<uuid>`), which this classification would sweep into the orchestrator thread if they ever appear in a capture. Might be worth a sentence in the `detection.py` comment as a known unknown.

PR #566, opened 2026-08-17
feat: cron kind, operator provenance, reconstruction fidelity + spawn-less session placement in the openclaw telemetry-hal example

alexandraabbas:
*Opened by Claude Code on behalf of @alexandraabbas; reviewed by @alexandraabbas.*

Upstream ports for the `examples/sources/openclaw_telemetry_hal` worked example, from patches we made to our vendored copy (per the example's stated copy-and-adapt purpose) while importing real CRUX research-agent telemetry — cron-scheduled, long-horizon OpenClaw runs. This is the "separate PR with some patches we've come across" promised in #560's filing, now carrying the whole vendored delta except the `explicit` session kind (see the heads-up at the end). Three groups: session-kind/provenance additions, reconstruction-fidelity fixes, and one deliberate design change called out separately for scrutiny.

## a. Session kinds + operator provenance

**`cron` is an orchestrator session kind.** OpenClaw runs can be cron-scheduled; the orchestrator session of such a run carries sessionKey kind `cron` (e.g. `agent:main:cron:<uuid>`). The importer's strict session classification refused these files outright (`unrecognized session kind 'cron'`). `detection.ORCH_KINDS` now includes `"cron"`, so a cron-scheduled run's orchestrator turns classify as orchestrator — the same treatment as `main`/`telegram`/`dashboard`. Our CRUX captures are all cron-scheduled runs, so without this the importer refuses the entire file.

**`message.in` becomes operator-sourced user-message provenance.** `message.in` events are the inbound operator channel: the full text of a message a human sent the agent mid-run. Previously all `message.*` events were unconsumed. This PR consumes `message.in` (only — outbound `message.sending`/`message.out` stay unconsumed, with the existing sub-agent-report-attribution rationale intact):

- an inbound message whose text matches a user turn stamps **that** turn `ChatMessageUser(source="operator")` — no duplicate message is added;
- an inbound message that never re-entered the session thread (e.g. delivered while the agent was busy) becomes its own operator-sourced user message, placed by its timestamp.

Why: for long-horizon autonomous runs, mid-run operator interventions are analytically load-bearing — downstream human-intervention analysis needs to distinguish "a human steered the agent here" from runtime-context injections and scaffold-generated user turns. `source="operator"` is a valid `ChatMessage.source` in current inspect_ai, and both bundled fixtures already contain `message.in` events, which now surface this provenance (the sample fixture's three Telegram prompts are stamped `operator`; message counts are unchanged).

## b. Reconstruction fidelity

**Scaffold roll-up usage aggregates are dropped.** OpenClaw's turn-finalization bookkeeping appends an AGGREGATE assistant record after a block of per-call records: no `responseId`, stopReason `stop`, no toolCalls, and a `totalTokens` FROZEN at the previous record's value — its usage restates the block it closes, not a new API call. Counting it double-counts every usage field: on one real capture, 9 such records added ~23M phantom cache-read tokens (~10% inflation of the headline billable total). `parse._drop_rollup_aggregates` drops records matching the full conjunctive signature against the previously kept turn, with an info log; a `responseId`, a toolCall, a non-`stop` stopReason, or a moving `totalTokens` each disqualify, so genuine turns are never touched.

**`durationMs`-derived tool-call width on bare captures.** Bare (non-service-sink) captures carry no `ts` envelope on `tool.*` events, so every schema-B sub-agent call collapsed to zero duration. `durationMs` (from `tool.end`) now gives such a call a real start→end span (`completed = timestamp + durationMs`); recorded `ts` pairs still win when present, and only a call with neither collapses to zero. Downstream busy-time sums rely on these widths. (The related `success=false` → `failed`/`error` surfacing already landed with the original example in #490 — no change here, existing tests still lock it.)

**Task-line span naming.** A sub-agent span with no spawn `label` was named the generic `"subagent"`. It now falls back to the first task-bearing line of the sub-agent's own prompt, skipping OpenClaw's `[Subagent Context] You are running…` scaffold preamble — so spans stay identifiable in the viewer even when no spawn arguments exist (which is every spawn-less session in group c).

## c. Design change, flagged for scrutiny: spawn-less sub-agent sessions are placed, not dropped

This one **changes an upstream design choice** (drop-with-warning for sub-agents with no linkable spawn call), so calling it out separately rather than folding it quietly into (b).

The original rationale ("not observed in practice — every sub-agent links via its spawn result's `childSessionKey`") held for the captures the example was built against, but doesn't survive cron-scheduled runs: cron/system-spawned sub-agent sessions have **no `sessions_spawn` call in the telemetry at all**, and they are common — on one real cron-scheduled CRUX capture, **46 of 101 sub-agent sessions** have no linkable spawn call. Drop-with-warning silently loses roughly half the delegated activity on such a run.

The change: an unlinked session is **placed at its file-order anchor** — its agent span is emitted next to the latest orchestrator turn seen (in file order) before the session's first event — with an info log listing the placed sessions. **No spawn tool call is fabricated**: the span simply has no folded `sessions_spawn` child, and the README documents the placement as a file-order heuristic ("the orchestrator was here when this session started"), not a recorded parent link. Genuinely-unknown session kinds are still refused loudly; that check is untouched.

If you'd rather keep drop-with-warning as the upstream default, or want the placement behind an option, happy to split this commit's placement portion into its own PR for separate discussion — the (a)/(b) items stand on their own without it.

## Tests

Extended `tests/test_telemetry_hal.py` (81 pass, from 73):

- `cron` in the orchestrator-kinds parametrize; a new `TestOperatorChannel` class (matched `message.in` stamps without duplication, unmatched becomes a standalone operator-sourced message, fixture prompts carry `source="operator"`);
- a new `TestRollupAggregates` class: roll-up dropped and usage sums no longer double-counted, plus negative controls (moving `totalTokens`, `responseId`/toolCall/stopReason each disqualify);
- schema-B timing: missing-`ts` calls derive width from `durationMs`; only calls with neither collapse to zero;
- a new `TestSpawnlessSubagents` class: placement between the correct orchestrator turns (anchor asserted), no fabricated spawn child, span named from the prompt's task line (boilerplate skipped), anchor-at-run-start placement; the old drop test is now `test_unlinked_subagent_placed_not_dropped`;
- unrecognized-session-kind refusal unchanged and still covered.

`ruff check` / `ruff format` / `mypy` clean on the example; `pytest examples/sources/openclaw_telemetry_hal/tests/` → 81 passed (against inspect_ai@main per CONTRIBUTING). Verified end-to-end on a real 120MB cron-scheduled CRUX capture (previously refused at the `cron` kind check; now imports with all sub-agent sessions represented).

## Not included / heads-up

- **#565's `explicit` session kind landed separately (merged)** — this branch is rebased on top of it; `ORCH_KINDS` here carries the union (`cron` + `explicit`).

### DIFF
--- a/examples/sources/openclaw_telemetry_hal/events.py
+++ b/examples/sources/openclaw_telemetry_hal/events.py
@@ -14,8 +14,10 @@ Mapping (mirrors the Claude Code importer):
   linkable (e.g. cron/system-spawned sessions), placed at its file-order
   anchor. Sub-agent messages are excluded from the main thread.
 - Inbound operator messages (``message.in``) mark the matching user turns
-  ``source="operator"``; an inbound message with no matching user turn (it never
-  entered the session thread) becomes its own operator-sourced user message.
+  ``source="operator"`` (occurrence-for-occurrence, in time order); an inbound
+  message with no matching user turn (it never entered the session thread)
+  becomes its own operator-sourced user message in the thread — excluded from
+  later ``ModelEvent.input``, which reflects only what the model was shown.
 """
 
 from __future__ import annotations
@@ -49,6 +51,7 @@ from inspect_ai.tool import ToolResult as ToolResultContent
 
 from .extraction import (
     content_blocks,
+    content_to_text,
     rich_or_text,
     toolcalls_of,
     usage_to_inspect,
@@ -90,19 +93,16 @@ def _stop_and_error(turn: dict[str, Any]) -> tuple[StopReason, str | None]:
     return "unknown", (str(message) if message else None)
 
 
-def _plain_text(content: object) -> str:
-    """A message ``content`` as comparable plain text (str or rich blocks).
+def _match_text(content: object) -> str:
+    """A message ``content`` as comparable text for operator reconciliation.
 
-    Used to match an inbound ``message.in`` (always a plain string) against the
-    session's user turns (plain strings or content-block lists).
+    Both sides flatten through :func:`.extraction.content_to_text`: an inbound
+    ``message.in`` is a plain string, but its user turn may have re-entered
+    the thread as a content-block list, and the two must compare equal (same
+    joiner, same block filter) or the turn is left unstamped and the send
+    duplicated as a standalone message.
     """
-    if isinstance(content, str):
-        return content.strip()
-    if isinstance(content, list):
-        return " ".join(
-            str(b.get("text", "")) for b in content if isinstance(b, dict)
-        ).strip()
-    return str(content or "").strip()
+    return content_to_text(content).strip()
 
 
 def build_content(
@@ -111,16 +111,19 @@ def build_content(
     """Build the event stream and the main-thread messages in one pass.
 
     User turns, orchestrator assistant turns, and real compactions are
-    interleaved by timestamp. A single running conversation is threaded through
-    both outputs: each ``ModelEvent.input`` carries the conversation up to that
+    interleaved by timestamp. A running conversation is threaded through both
+    outputs: each ``ModelEvent.input`` carries the conversation up to that
     turn (so the events view shows user prompts and tool results, matching the
-    Claude Code importer), and the final running list is the message thread.
+    Claude Code importer), and the full running list is the message thread.
 
     Inbound operator messages (``message.in``) are reconciled with the user
-    turns by content: a match stamps the existing turn ``source="operator"``
-    (never a duplicate message); an unmatched inbound message — one that never
-    entered the session thread — is added as its own operator-sourced user
-    message so mid-run operator interventions are preserved.
+    turns by content, occurrence-for-occurrence in time order: each send
+    stamps the first unconsumed user turn carrying the same text at-or-after
+    it ``source="operator"`` (never a duplicate message). A send with no such
+    turn — it never entered the session thread — is added as its own
+    operator-sourced user message so mid-run operator interventions are
+    preserved; it appears in the message thread only, NOT in later
+    ``ModelEvent.input``, because the model never saw it.
 
     Each sub-agent agent span is inserted immediately after the orchestrator
     tool event that spawned it. A sub-agent whose spawn could not be linked to
@@ -154,20 +157,32 @@ def build_content(
             ", ".join(sa.session_key for sa in unlinked),
         )
 
-    # The inbound operator channel (``message.in``). An inbound message whose
-    # text matches a user turn marks THAT turn ``source="operator"`` (no
-    # duplicate message); one with no matching user turn (it never entered the
-    # session thread, e.g. delivered while the agent was busy) becomes its own
-    # operator-sourced user message, placed by its timestamp.
-    operator_contents = {_plain_text(m.get("content")) for m in parse.operator_messages}
-    matched = {
-        _plain_text(t.get("content")) for t in parse.user_turns
-    } & operator_contents
-    extra_operator = [
-        m
-        for m in parse.operator_messages
-        if _plain_text(m.get("content")) not in matched
-    ]
+    # The inbound operator channel (``message.in``). Each send pairs with at
+    # most ONE user turn carrying the same text — the first unconsumed one
+    # at-or-after the send, since a message can only enter the session thread
+    # after it arrives (observed lags on real captures run from 0 to minutes;
+    # user turns carry no idempotencyKey there, so text is the only join key).
+    # Set-of-texts matching would silently drop a repeated send once its first
+    # occurrence matched, and could stamp a twin turn that predates the send.
+    # A send with no matching turn (it never entered the session thread, e.g.
+    # delivered while the agent was busy) becomes its own operator-sourced
+    # user message, placed by its timestamp.
+    turns_by_text: dict[str, list[dict[str, Any]]] = {}
+    for turn in parse.user_turns:  # already timestamp-sorted
+        turns_by_text.setdefault(_match_text(turn.get("content")), []).append(turn)
+    operator_turn_ids: set[int] = set()
+    extra_operator: list[dict[str, Any]] = []
+    for m in sorted(
+        parse.operator_messages, key=lambda m: int(m.get("timestamp") or 0)
+    ):
+        queue = turns_by_text.get(_match_text(m.get("content")), [])
+        m_ts = int(m.get("timestamp") or 0)
+        turn = next((t for t in queue if int(t.get("timestamp") or 0) >= m_ts), None)
+        if turn is not None:
+            queue.remove(turn)
+            operator_turn_ids.add(id(turn))
+        else:
+            extra_operator.append(m)
 
     # Merge user turns + operator messages + assistant turns + real compactions,
     # ordered by timestamp.
@@ -183,7 +198,8 @@ def build_content(
     ordered.sort(key=lambda item: item[0])
 
     events: list[Event] = []
-    messages: list[ChatMessage] = []  # running conversation; also the final thread
+    thread: list[ChatMessage] = []  # the final message thread (all messages)
+    conversation: list[ChatMessage] = []  # what the model saw (ModelEvent.input)
     order = 0  # monotonic working_start ordinal (stable tie-break for the timeline)
     last_ts = _ts_to_datetime(ordered[0][0]) if ordered else None
     last_ts = last_ts or _EPOCH
@@ -196,17 +212,19 @@ def build_content(
             # An operator-delivered turn carries its true source; other user
             # turns (runtime context, inter-session messages) carry none.
             source: Literal["operator"] | None = (
-                "operator" if _plain_text(item.get("content")) in matched else None
+                "operator" if id(item) in operator_turn_ids else None
             )
-            messages.append(
-                ChatMessageUser(
-                    content=rich_or_text(item.get("content")), source=source
-                )
+            user_msg = ChatMessageUser(
+                content=rich_or_text(item.get("content")), source=source
             )
+            thread.append(user_msg)
+            conversation.append(user_msg)
             continue
 
         if kind == "operator":  # message.in not present in the session thread
-            messages.append(
+            # Thread-only: the model never saw this message, so it must not
+            # appear in later ModelEvents' input.
+            thread.append(
                 ChatMessageUser(
                     content=rich_or_text(item.get("content")), source="operator"
                 )
@@ -273,7 +291,7 @@ def build_content(
         events.append(
             ModelEvent(
                 model=turn_model,
-                input=list(messages),  # conversation up to (not including) this turn
+                input=list(conversation),  # what the model saw before this turn
                 tools=[],
                 tool_choice="auto",
                 config=GenerateConfig(),
@@ -285,7 +303,8 @@ def build_content(
             )
         )
         order += 1
-        messages.append(assistant_msg)
+        thread.append(assistant_msg)
+        conversation.append(assistant_msg)
 
         for tc in toolcalls:
             tc_id = str(tc.get("id") or "")
@@ -322,14 +341,14 @@ def build_content(
 
             # The tool result still belongs in the orchestrator message thread,
             # whether or not the tool spawned a sub-agent.
-            messages.append(
-                ChatMessageTool(
-                    content=result_content,
-                    tool_call_id=tc_id,
-                    function=function,
-                    error=error,
-                )
+            tool_msg = ChatMessageTool(
+                content=result_content,
+                tool_call_id=tc_id,
+                function=function,
+                error=error,
             )
+            thread.append(tool_msg)
+            conversation.append(tool_msg)
 
     # Any unlinked sub-agent spans anchored after the last orchestrator turn.
     for sa in unlinked:
@@ -337,7 +356,7 @@ def build_content(
             events, sa, last_ts, order, parse.result_by_callid, spawn_tool=None
         )
 
-    return events, messages
+    return events, thread
 
 
 def _emit_subagent_span(
--- a/examples/sources/openclaw_telemetry_hal/events.py
+++ b/examples/sources/openclaw_telemetry_hal/events.py
@@ -181,8 +181,10 @@ def build_content(
     # delivered while the agent was busy) becomes its own operator-sourced
     # user message, placed by its timestamp.
     turns_by_text: dict[str, list[dict[str, Any]]] = {}
-    for turn in parse.user_turns:  # already timestamp-sorted
-        turns_by_text.setdefault(_match_text(turn.get("content")), []).append(turn)
+    for user_turn in parse.user_turns:  # already timestamp-sorted
+        turns_by_text.setdefault(_match_text(user_turn.get("content")), []).append(
+            user_turn
+        )
     operator_turn_ids: set[int] = set()
     extra_operator: list[dict[str, Any]] = []
     for m in sorted(
@@ -190,7 +192,9 @@ def build_content(
     ):
         queue = turns_by_text.get(_match_text(m.get("content")), [])
         m_ts = int(m.get("timestamp") or 0)
-        turn = next((t for t in queue if int(t.get("timestamp") or 0) >= m_ts), None)
+        turn: dict[str, Any] | None = next(
+            (t for t in queue if int(t.get("timestamp") or 0) >= m_ts), None
+        )
         if turn is not None:
             queue.remove(turn)
             operator_turn_ids.add(id(turn))
--- a/examples/sources/openclaw_telemetry_hal/events.py
+++ b/examples/sources/openclaw_telemetry_hal/events.py
@@ -136,18 +136,31 @@ def build_content(
     """
     # Each spawned sub-agent links to the single orchestrator toolCall that
     # spawned it: a spawn result names exactly one ``childSessionKey``, so the
-    # mapping is 1:1. A sub-agent with no linkable spawn call (or, defensively,
-    # one whose call is already claimed) is placed by its file-order anchor
-    # instead — dropping it would silently lose that session's activity (46 of
-    # 101 sub-agent sessions on a real cron-scheduled CRUX capture).
+    # mapping is 1:1. A sub-agent with no linkable spawn call is placed by its
+    # file-order anchor instead — dropping it would silently lose that
+    # session's activity (46 of 101 sub-agent sessions on a real
+    # cron-scheduled CRUX capture).
     subagent_by_tool_call: dict[str, SubagentSpan] = {}
     unlinked: list[SubagentSpan] = []
     for sa in parse.subagents:
         tc_id = sa.spawn_tool_call_id
-        if tc_id is not None and tc_id not in subagent_by_tool_call:
-            subagent_by_tool_call[tc_id] = sa
-        else:
+        if tc_id is None:
             unlinked.append(sa)
+        elif tc_id in subagent_by_tool_call:
+            # A second claim on the same spawn call violates the documented
+            # 1:1 spawn->session mapping (corrupt/ambiguous linkage). Place it
+            # like a spawn-less session, but — unlike that normal cron
+            # pattern — surface the anomaly as a warning.
+            logger.warning(
+                "OpenClaw sub-agent %s claims spawn call %s already claimed "
+                "by %s; placing it by file-order anchor instead",
+                sa.session_key,
+                tc_id,
+                subagent_by_tool_call[tc_id].session_key,
+            )
+            unlinked.append(sa)
+        else:
+            subagent_by_tool_call[tc_id] = sa
     unlinked.sort(key=lambda sa: sa.anchor_ts)
     if unlinked:
         logger.info(
@@ -208,6 +221,25 @@ def build_content(
         ts = _ts_to_datetime(ts_ms) or last_ts
         last_ts = ts
 
+        # Emit any unlinked sub-agent spans whose file-order anchor has been
+        # passed, so each lands immediately AFTER the orchestrator turn the
+        # run was on when the session first appeared. Strictly-before (<): a
+        # span whose anchor EQUALS this item's ts belongs on the far side of
+        # the anchor turn — the turn had already been produced when the
+        # session appeared. Running the flush on every timeline item keeps
+        # the span as close to its anchor as possible (in particular, before
+        # an interleaved compaction rather than after it); the trailing loop
+        # below covers anchors at or past the last item.
+        while unlinked and unlinked[0].anchor_ts < ts_ms:
+            order = _emit_subagent_span(
+                events,
+                unlinked.pop(0),
+                ts,
+                order,
+                parse.result_by_callid,
+                spawn_tool=None,
+            )
+
         if kind == "user":
             # An operator-delivered turn carries its true source; other user
             # turns (runtime context, inter-session messages) carry none.
@@ -248,19 +280,6 @@ def build_content(
             )
             order += 1
             continue
-
-        # Emit any unlinked sub-agent spans whose file-order anchor has been
-        # reached, so each lands next to the orchestrator turn the run was on
-        # when the session first appeared.
-        while unlinked and unlinked[0].anchor_ts <= ts_ms:
-            order = _emit_subagent_span(
-                events,
-                unlinked.pop(0),
-                ts,
-                order,
-                parse.result_by_callid,
-                spawn_tool=None,
-            )
 
         # Assistant turn: emit a model event (carrying the conversation so far as
         # input), then its tool events.
@@ -546,7 +565,7 @@ def _emit_subagent_span(
         # call a real width — downstream busy-time sums rely on it — so only a
         # call with neither collapses to zero duration.
         call_start = _ts_to_datetime(call.get("startTs")) or timestamp
-        duration_ms = call.get("durationMs")
+        duration_ms = _duration_to_ms(call.get("durationMs"))
         call_end = _ts_to_datetime(call.get("endTs")) or (
             call_start + timedelta(milliseconds=duration_ms)
             if duration_ms
@@ -626,17 +645,47 @@ def _ts_to_datetime(value: Any) -> datetime | None:
     return datetime.fromtimestamp(ms / 1000.0, tz=timezone.utc)
 
 
+def _duration_to_ms(value: Any) -> int | None:
+    """Coerce a telemetry ``durationMs`` to positive-int milliseconds.
+
+    The value comes verbatim out of ``json.loads``, so guard it the way
+    ``_ts_to_datetime`` guards timestamps: a numeric string still yields a
+    width; a non-numeric or non-positive value yields ``None`` (zero width)
+    rather than aborting the import inside ``timedelta`` arithmetic — or
+    placing a call's completion before its start.
+    """
+    try:
+        ms = int(value)
+    except (TypeError, ValueError):
+        return None
+    return ms if ms > 0 else None
+
+
+_TASK_TAG = "[Subagent Task]"
+
+
 def _task_line(prompt: str | None) -> str | None:
     """The first task-bearing line of a sub-agent prompt, for span naming.
 
     A spawn-less (anchor-placed) session has no ``label``/``task`` arguments to
     name its span from, only its own prompt — which opens with OpenClaw's
-    scaffold preamble (e.g. ``[Subagent Context] You are running as a
-    subagent…``). Skip bracketed and boilerplate lines and return the first
-    line that carries the actual task, or ``None`` when there is none.
+    scaffold preamble (a timestamp tag and/or ``[Subagent Context] You are
+    running as a subagent…``) and usually announces the task with a
+    ``[Subagent Task]`` tag: inline (``[Subagent Task]: <task>``) or with the
+    task on the following line. The tag is authoritative when present.
+    Otherwise fall back to the first line that is neither bracketed nor the
+    subagent boilerplate — prompt BODIES also carry bracketed lines (template
+    placeholders), so an unrecognized bracket line never becomes a name.
+    Returns ``None`` when nothing task-bearing is found.
     """
-    for line in (prompt or "").splitlines():
-        line = line.strip()
+    lines = [line.strip() for line in (prompt or "").splitlines()]
+    for i, line in enumerate(lines):
+        if line.startswith(_TASK_TAG):
+            inline = line[len(_TASK_TAG) :].lstrip(" :").strip()
+            if inline:
+                return inline
+            return next((follow for follow in lines[i + 1 :] if follow), None)
+    for line in lines:
         if (
             line
             and not line.startswith("[")
--- a/examples/sources/openclaw_telemetry_hal/events.py
+++ b/examples/sources/openclaw_telemetry_hal/events.py
@@ -12,13 +12,16 @@ Mapping (mirrors the Claude Code importer):
   (``SpanBeginEvent`` / ``SpanEndEvent``, ``type="agent"``) anchored at the
   ``sessions_spawn`` tool call that created it. Sub-agent messages are excluded
   from the main thread.
+- Inbound operator messages (``message.in``) mark the matching user turns
+  ``source="operator"``; an inbound message with no matching user turn (it never
+  entered the session thread) becomes its own operator-sourced user message.
 """
 
 from __future__ import annotations
 
 from datetime import datetime, timezone
 from logging import getLogger
-from typing import Any, cast
+from typing import Any, Literal, cast
 
 from inspect_ai.event import (
     CompactionEvent,
@@ -86,6 +89,21 @@ def _stop_and_error(turn: dict[str, Any]) -> tuple[StopReason, str | None]:
     return "unknown", (str(message) if message else None)
 
 
+def _plain_text(content: object) -> str:
+    """A message ``content`` as comparable plain text (str or rich blocks).
+
+    Used to match an inbound ``message.in`` (always a plain string) against the
+    session's user turns (plain strings or content-block lists).
+    """
+    if isinstance(content, str):
+        return content.strip()
+    if isinstance(content, list):
+        return " ".join(
+            str(b.get("text", "")) for b in content if isinstance(b, dict)
+        ).strip()
+    return str(content or "").strip()
+
+
 def build_content(
     parse: OpenClawTelemetry,
 ) -> tuple[list[Event], list[ChatMessage]]:
@@ -96,6 +114,12 @@ def build_content(
     both outputs: each ``ModelEvent.input`` carries the conversation up to that
     turn (so the events view shows user prompts and tool results, matching the
     Claude Code importer), and the final running list is the message thread.
+
+    Inbound operator messages (``message.in``) are reconciled with the user
+    turns by content: a match stamps the existing turn ``source="operator"``
+    (never a duplicate message); an unmatched inbound message — one that never
+    entered the session thread — is added as its own operator-sourced user
+    message so mid-run operator interventions are preserved.
 
     Each sub-agent agent span is inserted immediately after the orchestrator
     tool event that spawned it. A sub-agent whose spawn could not be linked to a
@@ -123,9 +147,26 @@ def build_content(
             ", ".join(sa.session_key for sa in dropped),
         )
 
-    # Merge user turns + assistant turns + real compactions, ordered by timestamp.
+    # The inbound operator channel (``message.in``). An inbound message whose
+    # text matches a user turn marks THAT turn ``source="operator"`` (no
+    # duplicate message); one with no matching user turn (it never entered the
+    # session thread, e.g. delivered while the agent was busy) becomes its own
+    # operator-sourced user message, placed by its timestamp.
+    operator_contents = {_plain_text(m.get("content")) for m in parse.operator_messages}
+    matched = {
+        _plain_text(t.get("content")) for t in parse.user_turns
+    } & operator_contents
+    extra_operator = [
+        m
+        for m in parse.operator_messages
+        if _plain_text(m.get("content")) not in matched
+    ]
+
+    # Merge user turns + operator messages + assistant turns + real compactions,
+    # ordered by timestamp.
     ordered: list[tuple[int, str, dict[str, Any]]] = (
         [(int(t.get("timestamp") or 0), "user", t) for t in parse.user_turns]
+        + [(int(m.get("timestamp") or 0), "operator", m) for m in extra_operator]
         + [
             (int(t.get("timestamp") or 0), "assistant", t)
             for t in parse.orchestrator_turns
@@ -145,7 +186,24 @@ def build_content(
         last_ts = ts
 
         if kind == "user":
-            messages.append(ChatMessageUser(content=rich_or_text(item.get("content"))))
+            # An operator-delivered turn carries its true source; other user
+            # turns (runtime context, inter-session messages) carry none.
+            source: Literal["operator"] | None = (
+                "operator" if _plain_text(item.get("content")) in matched else None
+            )
+            messages.append(
+                ChatMessageUser(
+                    content=rich_or_text(item.get("content")), source=source
+                )
+            )
+            continue
+
+        if kind == "operator":  # message.in not present in the session thread
+            messages.append(
+                ChatMessageUser(
+                    content=rich_or_text(item.get("content")), source="operator"
+                )
+            )
             continue
 
         if kind == "compaction":
--- a/examples/sources/openclaw_telemetry_hal/events.py
+++ b/examples/sources/openclaw_telemetry_hal/events.py
@@ -10,8 +10,9 @@ Mapping (mirrors the Claude Code importer):
   main ``messages`` thread.
 - Each delegated sub-agent becomes a nested **agent span**
   (``SpanBeginEvent`` / ``SpanEndEvent``, ``type="agent"``) anchored at the
-  ``sessions_spawn`` tool call that created it. Sub-agent messages are excluded
-  from the main thread.
+  ``sessions_spawn`` tool call that created it — or, when no spawn call is
+  linkable (e.g. cron/system-spawned sessions), placed at its file-order
+  anchor. Sub-agent messages are excluded from the main thread.
 - Inbound operator messages (``message.in``) mark the matching user turns
   ``source="operator"``; an inbound message with no matching user turn (it never
   entered the session thread) becomes its own operator-sourced user message.
@@ -19,7 +20,7 @@ Mapping (mirrors the Claude Code importer):
 
 from __future__ import annotations
 
-from datetime import datetime, timezone
+from datetime import datetime, timedelta, timezone
 from logging import getLogger
 from typing import Any, Literal, cast
 
@@ -122,29 +123,35 @@ def build_content(
     message so mid-run operator interventions are preserved.
 
     Each sub-agent agent span is inserted immediately after the orchestrator
-    tool event that spawned it. A sub-agent whose spawn could not be linked to a
-    tool call is dropped with a warning (not observed in practice — every
-    sub-agent in the CRUX1 captures links via its spawn result's
-    ``childSessionKey``). Sub-agent messages are excluded from the thread.
+    tool event that spawned it. A sub-agent whose spawn could not be linked to
+    a tool call — cron/system-spawned sessions have no ``sessions_spawn`` call
+    at all, and are the majority on scheduled runs — is placed at its
+    file-order anchor (the latest orchestrator turn before the session's first
+    activity) with NO spawn tool call fabricated, so its delegated activity is
+    kept rather than silently lost. Sub-agent messages are excluded from the
+    thread.
     """
-    # Each sub-agent links to the single orchestrator toolCall that spawned it:
-    # a spawn result names exactly one ``childSessionKey``, so the mapping is
-    # 1:1. A sub-agent whose spawn call is missing (or, defensively, already
-    # claimed) cannot be placed in the timeline, so it is dropped with a warning
-    # rather than guessed at (see docstring).
+    # Each spawned sub-agent links to the single orchestrator toolCall that
+    # spawned it: a spawn result names exactly one ``childSessionKey``, so the
+    # mapping is 1:1. A sub-agent with no linkable spawn call (or, defensively,
+    # one whose call is already claimed) is placed by its file-order anchor
+    # instead — dropping it would silently lose that session's activity (46 of
+    # 101 sub-agent sessions on a real cron-scheduled CRUX capture).
     subagent_by_tool_call: dict[str, SubagentSpan] = {}
-    dropped: list[SubagentSpan] = []
+    unlinked: list[SubagentSpan] = []
     for sa in parse.subagents:
         tc_id = sa.spawn_tool_call_id
         if tc_id is not None and tc_id not in subagent_by_tool_call:
             subagent_by_tool_call[tc_id] = sa
         else:
-            dropped.append(sa)
-    if dropped:
-        logger.warning(
-            "Dropping %d OpenClaw sub-agent(s) with no linkable spawn call: %s",
-            len(dropped),
-            ", ".join(sa.session_key for sa in dropped),
+            unlinked.append(sa)
+    unlinked.sort(key=lambda sa: sa.anchor_ts)
+    if unlinked:
+        logger.info(
+            "Placing %d OpenClaw sub-agent(s) without a linkable spawn call "
+            "by file-order anchor: %s",
+            len(unlinked),
+            ", ".join(sa.session_key for sa in unlinked),
         )
 
     # The inbound operator channel (``message.in``). An inbound message whose
@@ -223,6 +230,19 @@ def build_content(
             )
             order += 1
             continue
+
+        # Emit any unlinked sub-agent spans whose file-order anchor has been
+        # reached, so each lands next to the orchestrator turn the run was on
+        # when the session first appeared.
+        while unlinked and unlinked[0].anchor_ts <= ts_ms:
+            order = _emit_subagent_span(
+                events,
+                unlinked.pop(0),
+                ts,
+                order,
+                parse.result_by_callid,
+                spawn_tool=None,
+            )
 
         # Assistant turn: emit a model event (carrying the conversation so far as
         # input), then its tool events.
@@ -311,6 +331,12 @@ def build_content(
                 )
             )
 
+    # Any unlinked sub-agent spans anchored after the last orchestrator turn.
+    for sa in unlinked:
+        order = _emit_subagent_span(
+            events, sa, last_ts, order, parse.result_by_callid, spawn_tool=None
+        )
+
     return events, messages
 
 
@@ -320,7 +346,7 @@ def _emit_subagent_span(
     timestamp: datetime,
     order: int,
     result_by_callid: dict[str, ToolResult],
-    spawn_tool: dict[str, Any],
+    spawn_tool: dict[str, Any] | None,
 ) -> int:
     """Emit a delegated sub-agent as a nested agent span and return next order.
 
@@ -329,7 +355,9 @@ def _emit_subagent_span(
     Claude Code importer, the orchestrator ``sessions_spawn`` call
     (``spawn_tool``) is emitted as the span's FIRST child, tagged with
     ``agent_span_id`` so the view folds it into the agent header rather than
-    drawing a standalone tool row.
+    drawing a standalone tool row. ``spawn_tool`` is ``None`` for an unlinked
+    (anchor-placed) session — no spawn call was recorded, so none is emitted
+    (fabricating one would forge telemetry that does not exist).
 
     - Schema A: each ``messages[]`` assistant turn becomes a ``ModelEvent``
     (with usage) followed by its ``ToolEvent``s (results matched by
@@ -352,7 +380,11 @@ def _emit_subagent_span(
     events.append(
         SpanBeginEvent(
             id=span_id,
-            name=sa.spawn_label or "subagent",
+            name=(
+                sa.spawn_label
+                or _short_description(_task_line(sa.prompt))
+                or "subagent"
+            ),
             type="agent",
             timestamp=timestamp,
             working_start=float(order),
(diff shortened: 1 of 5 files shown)


examples/sources/openclaw_telemetry_hal/parse.py
```diff
@@ -165,6 +187,50 @@ def _primary_model(orchestrator_turns: list[dict[str, Any]]) -> str | None:
...
+    def _is_rollup(turn: dict[str, Any], prev: dict[str, Any] | None) -> bool:
+        if prev is None or turn.get("responseId") is not None:
+            return False
+        if (turn.get("stopReason") or "stop") != "stop":
+            return False
+        if toolcalls_of(turn.get("content")):
+            return False
+        total = (turn.get("usage") or {}).get("totalTokens")
+        prev_total = (prev.get("usage") or {}).get("totalTokens")
+        return total is not None and total == prev_total
```
craig-meridian: **[P1] Genuine rid-less turn with equal `totalTokens` is dropped as a roll-up.**

On service-sink captures no turn carries a `responseId` (parse.py:456 and this PR's own test comment), so the conjunctive signature degrades to `stop` + no-toolCalls + equal-`totalTokens`. Usage here is per-call, not cumulative (README:115), so two consecutive genuine short text-only replies with identical per-call cost are a reachable state — and the second real turn is then silently deleted from both the event stream and the message thread (info-level count log only). Reproduced with two distinct-text turns sharing a total: only the first survived.

Suggested fix: require evidence unique to aggregate records (e.g. full-usage equality plus content restatement of the previous record), and add a negative test for distinct turns with equal totals — `test_moving_total_tokens_is_kept` only covers the moving-total escape.

examples/sources/openclaw_telemetry_hal/parse.py
```diff
@@ -165,6 +187,50 @@ def _primary_model(orchestrator_turns: list[dict[str, Any]]) -> str | None:
...
+
+    kept: list[dict[str, Any]] = []
+    n_rollups = 0
+    for turn in orchestrator_turns:
+        if _is_rollup(turn, kept[-1] if kept else None):
```
craig-meridian: **[P1] Cross-surface interleaving defeats roll-up detection.**

`orchestrator_turns` merges all orchestrator surfaces (`main`/`telegram`/`dashboard`/`cron`/`explicit`) into one timestamp-sorted list before `_drop_rollup_aggregates` runs, and `_is_rollup` compares only against `kept[-1]`. If a turn from another surface interleaves between a block's last per-call record and its roll-up (main per-call ts=1000 total=950; telegram turn ts=1200 total=500; main roll-up ts=1500 total=950), `prev` is the telegram turn, totals mismatch, and the roll-up is kept — reintroducing exactly the double-count this function exists to prevent. Multi-surface captures are real: detection.py documents `explicit` sessions coexisting with `main` in actual captures.

Suggested fix: run the filter per surface (group by `assistant_kind`) before merging.

examples/sources/openclaw_telemetry_hal/events.py
```diff
@@ -97,35 +116,64 @@ def build_content(
...
         )
 
-    # Merge user turns + assistant turns + real compactions, ordered by timestamp.
+    # The inbound operator channel (``message.in``). An inbound message whose
+    # text matches a user turn marks THAT turn ``source="operator"`` (no
+    # duplicate message); one with no matching user turn (it never entered the
+    # session thread, e.g. delivered while the agent was busy) becomes its own
+    # operator-sourced user message, placed by its timestamp.
+    operator_contents = {_plain_text(m.get("content")) for m in parse.operator_messages}
+    matched = {
```
craig-meridian: **[P1] Operator reconciliation by run-wide text set loses interventions and mis-stamps twins.**

The set intersection ignores occurrence count and chronology. Reproduced: a user turn `"status?"` at ts=1000 followed by an unmatched inbound `message.in` `"status?"` at ts=2500 yields ONE user message — the t=2500 intervention vanishes from the transcript and the earlier turn is retroactively stamped `source="operator"`. An operator repeating "continue" collapses the same way, and any non-operator user turn coincidentally sharing text with one operator message is falsely stamped (no temporal scoping).

Suggested fix: pair each inbound message with at most one user occurrence — prefer stable IDs or timestamps (`message.in` carries both `timestamp` and `metadata.messageId`, and the fixture's inbound timestamps equal their user turns' exactly) — falling back to chronological text matching, and surface remaining inbound occurrences independently.

examples/sources/openclaw_telemetry_hal/events.py
```diff
@@ -430,9 +522,17 @@ def _emit_subagent_span(
...
-        # captures), which collapses it to zero duration but never mis-orders it.
+        # captures), which never mis-orders it. Bare captures record no
+        # ``endTs`` either, but ``durationMs`` (from tool.end) still gives the
+        # call a real width — downstream busy-time sums rely on it — so only a
+        # call with neither collapses to zero duration.
         call_start = _ts_to_datetime(call.get("startTs")) or timestamp
-        call_end = _ts_to_datetime(call.get("endTs")) or call_start
+        duration_ms = call.get("durationMs")
+        call_end = _ts_to_datetime(call.get("endTs")) or (
+            call_start + timedelta(milliseconds=duration_ms)
```
craig-meridian: **[P1] Unguarded `timedelta(milliseconds=durationMs)` — a non-numeric value aborts the whole import.**

`durationMs` is copied verbatim from `json.loads` output (parse.py:585) with no coercion, and the truthiness guard passes strings: reproduced `durationMs: "250"` → `TypeError: unsupported type for timedelta milliseconds component: str` out of `build_content`, failing a file the pre-PR importer handled (durationMs was metadata-only then). Also reproduced: `durationMs: -500` yields `completed` 0.5s *before* `timestamp`. The adjacent `startTs`/`endTs` path is defensively hardened (`_ts_to_datetime` wraps `int()` in try/except); this new arithmetic path deserves the same guard.

Suggested fix: coerce via a small helper (try/except `int()`, reject `<= 0`) before the arithmetic, mirroring `_ts_to_datetime`.

examples/sources/openclaw_telemetry_hal/events.py
```diff
@@ -166,6 +231,19 @@ def build_content(
             order += 1
             continue
 
+        # Emit any unlinked sub-agent spans whose file-order anchor has been
+        # reached, so each lands next to the orchestrator turn the run was on
+        # when the session first appeared.
+        while unlinked and unlinked[0].anchor_ts <= ts_ms:
```
craig-meridian: **[P2] Off-by-one: unlinked span emitted *before* its anchor turn, not "immediately after".**

This `<=` flush runs before the current assistant turn is emitted, so a sub-agent anchored at A2 (anchor_ts == A2's ts) is placed between A1 and A2. Reproduced: emitted order is `ModelEvent(A1), SpanBegin, SpanEnd, ModelEvent(A2), ModelEvent(A3)` — the timeline shows delegated activity starting while the orchestrator was still at A1, contradicting the README/docstring ("emitted immediately after the latest orchestrator turn seen ... before the session's first event"). `test_placed_at_file_order_anchor` only asserts "between A1 and A3", so it passes either way.

Suggested fix: use strict `<` (the post-loop flush already covers the anchor==last-turn tail) and tighten the test to assert exact adjacency to A2. Deeper alternative: merge unlinked spans into `ordered` as a fourth kind keyed by `anchor_ts`, eliminating both special-case emission sites.

examples/sources/openclaw_telemetry_hal/parse.py
```diff
@@ -173,13 +239,15 @@ def parse_telemetry(raw_events: Iterable[dict[str, Any]]) -> OpenClawTelemetry:
...
     c = _consolidate(raw_events)
 
-    orchestrator_turns = sorted(
-        (
-            c.assistant_by_id[rid]
-            for rid, kind in c.assistant_kind.items()
-            if is_orchestrator(kind)
-        ),
-        key=lambda m: int(m.get("timestamp") or 0),
+    orchestrator_turns = _drop_rollup_aggregates(
```
craig-meridian: **[P2] Schema-A sub-agent turns bypass the roll-up dropper.**

`_drop_rollup_aggregates` runs only here, on orchestrator turns; `SubagentSpan.turns` never pass through it, yet their usage feeds sub-agent ModelEvents (events.py schema-A path) and transcript `total_tokens` (transcripts.py). Reproduced: a sub-agent whose `messages[]` contain a per-call record (totalTokens=950) followed by a roll-up-shaped record (rid-less, `stop`, no toolCalls, frozen total) produces two ModelEvents summing 1900 tokens against a true cost of 950. The roll-up is turn-finalization scaffold bookkeeping with no stated orchestrator-only trait, and sub-agent sessions are finalized by the same scaffold — no fixture exhibits it yet, but nothing precludes or handles it.

Suggested fix: apply the dropper to each `SubagentSpan.turns` as well.

examples/sources/openclaw_telemetry_hal/events.py
```diff
@@ -86,6 +90,21 @@ def _stop_and_error(turn: dict[str, Any]) -> tuple[StopReason, str | None]:
     return "unknown", (str(message) if message else None)
 
 
+def _plain_text(content: object) -> str:
```
craig-meridian: **[P2] `_plain_text` diverges from the existing `content_to_text` — a multi-block turn double-faults.**

This re-implements `extraction.content_to_text` with different semantics: space-join vs newline-join, no `type == "text"` filter, and bare-string blocks are skipped. Reproduced: `message.in` `"line1\nline2"` plus the same message re-entering the thread as two text blocks yields TWO user messages with byte-identical `.text` — the operator copy stamped, the real turn unstamped — violating this PR's own "never a duplicate message" invariant. Block-form canonical user turns are reachable (crux1 fixture).

Suggested fix: use `content_to_text(...).strip()` on both sides (its `\n` join makes the multi-line case match) and delete `_plain_text` — parse.py already uses `content_to_text` for exactly this kind of content matching.

examples/sources/openclaw_telemetry_hal/events.py
```diff
@@ -145,7 +193,24 @@ def build_content(
...
+            )
+            messages.append(
+                ChatMessageUser(
+                    content=rich_or_text(item.get("content")), source=source
+                )
+            )
+            continue
+
+        if kind == "operator":  # message.in not present in the session thread
+            messages.append(
```
craig-meridian: **[P2] Unmatched operator messages leak into subsequent `ModelEvent.input`.**

An unmatched `message.in` — by this PR's own definition one that "never entered the session thread" (the model never saw it) — is appended to the running `messages` list, which doubles as `ModelEvent.input` ("conversation up to this turn") for every later orchestrator turn. Reproduced: an unmatched `"secret aside"` at ts=2500 appears in the ts=3000 turn's input. Scanners analyzing what the model was actually shown will see a user message the model never received. The single-thread design already tolerates some divergence (post-compaction inputs are unpruned), but this adds a new fabricated-context case.

Suggested fix: exclude unmatched operator messages from the running input list (thread-only), or document the caveat explicitly for scanner authors.

examples/sources/openclaw_telemetry_hal/events.py
```diff
@@ -507,6 +607,26 @@ def _ts_to_datetime(value: Any) -> datetime | None:
...
+def _task_line(prompt: str | None) -> str | None:
+    """The first task-bearing line of a sub-agent prompt, for span naming.
+
+    A spawn-less (anchor-placed) session has no ``label``/``task`` arguments to
+    name its span from, only its own prompt — which opens with OpenClaw's
+    scaffold preamble (e.g. ``[Subagent Context] You are running as a
+    subagent…``). Skip bracketed and boilerplate lines and return the first
+    line that carries the actual task, or ``None`` when there is none.
+    """
+    for line in (prompt or "").splitlines():
```
craig-meridian: **[P2] `startswith("[")` rejects the bundled crux1 fixture's standard inline task shape.**

crux1-sample-telemetry.jsonl prompts inline the task as `[Subagent Task]: Check for unread emails …`, which this filter skips — reproduced against the actual fixture prompt: `_task_line` returns the later `"Report back with any unread emails …"` boilerplate as the span name, and `None` (generic `subagent`) when no unbracketed line follows. Either way the actual task text is never chosen, defeating the naming improvement on the very telemetry the example ships with (the new test's prompt happens to put its task on an unbracketed line, so it doesn't catch this).

Suggested fix: handle the `[Subagent Task]:`/`[Subagent Task]` prefix explicitly (strip the bracket tag and keep the remainder, or take the next paragraph), skip only recognized context/timestamp boilerplate, and add a test with a fixture-shaped prompt.

examples/sources/openclaw_telemetry_hal/events.py
```diff
@@ -97,35 +116,64 @@ def build_content(
...
-            dropped.append(sa)
-    if dropped:
-        logger.warning(
-            "Dropping %d OpenClaw sub-agent(s) with no linkable spawn call: %s",
-            len(dropped),
-            ", ".join(sa.session_key for sa in dropped),
+            unlinked.append(sa)
+    unlinked.sort(key=lambda sa: sa.anchor_ts)
+    if unlinked:
+        logger.info(
```
craig-meridian: **[P3] The defensive "spawn call already claimed" anomaly lost its warning.**

The old WARNING covered two distinct cases: no linkable spawn call, and — defensively — a second sub-agent claiming an already-claimed `spawn_tool_call_id` (a violation of the documented 1:1 mapping, i.e. corrupt/ambiguous linkage). Both now funnel into this one info-level "placing by file-order anchor" message, so the genuinely anomalous duplicate-claim case is indistinguishable from the normal cron pattern and invisible at default log levels; the second claimant is quietly anchor-placed, possibly far from its true spawn point.

Suggested fix: keep info for the `tc_id is None` branch, but log a warning when `tc_id` was already claimed.

examples/sources/openclaw_telemetry_hal/parse.py
craig-meridian: **[P1] Genuine rid-less turn with equal `totalTokens` is dropped as a roll-up.**

On service-sink captures no turn carries a `responseId` (parse.py:456 and this PR's own test comment), so the conjunctive signature degrades to `stop` + no-toolCalls + equal-`totalTokens`. Usage here is per-call, not cumulative (README:115), so two consecutive genuine short text-only replies with identical per-call cost are a reachable state — and the second real turn is then silently deleted from both the event stream and the message thread (info-level count log only). Reproduced with two distinct-text turns sharing a total: only the first survived.

Suggested fix: require evidence unique to aggregate records (e.g. full-usage equality plus content restatement of the previous record), and add a negative test for distinct turns with equal totals — `test_moving_total_tokens_is_kept` only covers the moving-total escape.

examples/sources/openclaw_telemetry_hal/parse.py
craig-meridian, line 262: **[P1] Cross-surface interleaving defeats roll-up detection.**

`orchestrator_turns` merges all orchestrator surfaces (`main`/`telegram`/`dashboard`/`cron`/`explicit`) into one timestamp-sorted list before `_drop_rollup_aggregates` runs, and `_is_rollup` compares only against `kept[-1]`. If a turn from another surface interleaves between a block's last per-call record and its roll-up (main per-call ts=1000 total=950; telegram turn ts=1200 total=500; main roll-up ts=1500 total=950), `prev` is the telegram turn, totals mismatch, and the roll-up is kept — reintroducing exactly the double-count this function exists to prevent. Multi-surface captures are real: detection.py documents `explicit` sessions coexisting with `main` in actual captures.

Suggested fix: run the filter per surface (group by `assistant_kind`) before merging.

examples/sources/openclaw_telemetry_hal/events.py
craig-meridian: **[P1] Operator reconciliation by run-wide text set loses interventions and mis-stamps twins.**

The set intersection ignores occurrence count and chronology. Reproduced: a user turn `"status?"` at ts=1000 followed by an unmatched inbound `message.in` `"status?"` at ts=2500 yields ONE user message — the t=2500 intervention vanishes from the transcript and the earlier turn is retroactively stamped `source="operator"`. An operator repeating "continue" collapses the same way, and any non-operator user turn coincidentally sharing text with one operator message is falsely stamped (no temporal scoping).

Suggested fix: pair each inbound message with at most one user occurrence — prefer stable IDs or timestamps (`message.in` carries both `timestamp` and `metadata.messageId`, and in sample-telemetry.jsonl every inbound timestamp equals its user turn's exactly) — falling back to chronological text matching, and surface remaining inbound occurrences independently.

examples/sources/openclaw_telemetry_hal/events.py
craig-meridian, line 601: **[P1] Unguarded `timedelta(milliseconds=durationMs)` — a non-numeric value aborts the whole import.**

`durationMs` is copied verbatim from `json.loads` output (parse.py:585) with no coercion, and the truthiness guard passes strings: reproduced `durationMs: "250"` → `TypeError: unsupported type for timedelta milliseconds component: str` out of `build_content`, failing a file the pre-PR importer handled (durationMs was metadata-only then). Also reproduced: `durationMs: -500` yields `completed` 0.5s *before* `timestamp`. The adjacent `startTs`/`endTs` path is defensively hardened (`_ts_to_datetime` wraps `int()` in try/except); this new arithmetic path deserves the same guard.

Suggested fix: coerce via a small helper (try/except `int()`, reject `<= 0`) before the arithmetic, mirroring `_ts_to_datetime`.

examples/sources/openclaw_telemetry_hal/events.py
craig-meridian: **[P2] Off-by-one: unlinked span emitted *before* its anchor turn, not "immediately after".**

This `<=` flush runs before the current assistant turn is emitted, so a sub-agent anchored at A2 (anchor_ts == A2's ts) is placed between A1 and A2. Reproduced: emitted order is `ModelEvent(A1), SpanBegin, SpanEnd, ModelEvent(A2), ModelEvent(A3)` — the timeline shows delegated activity starting while the orchestrator was still at A1, contradicting the README/docstring ("emitted immediately after the latest orchestrator turn seen ... before the session's first event"). `test_placed_at_file_order_anchor` only asserts "between A1 and A3", so it passes either way.

Suggested fix: use strict `<` (the post-loop flush already covers the anchor==last-turn tail) and tighten the test to assert exact adjacency to A2. Deeper alternative: merge unlinked spans into `ordered` as a fourth kind keyed by `anchor_ts`, eliminating both special-case emission sites.

examples/sources/openclaw_telemetry_hal/parse.py
craig-meridian: **[P2] Schema-A sub-agent turns bypass the roll-up dropper.**

`_drop_rollup_aggregates` runs only here, on orchestrator turns; `SubagentSpan.turns` never pass through it, yet their usage feeds sub-agent ModelEvents (events.py schema-A path) and transcript `total_tokens` (transcripts.py). Reproduced: a sub-agent whose `messages[]` contain a per-call record (totalTokens=950) followed by a roll-up-shaped record (rid-less, `stop`, no toolCalls, frozen total) produces two ModelEvents summing 1900 tokens against a true cost of 950. The roll-up is turn-finalization scaffold bookkeeping with no stated orchestrator-only trait, and sub-agent sessions are finalized by the same scaffold — no fixture exhibits it yet, but nothing precludes or handles it.

Suggested fix: apply the dropper to each `SubagentSpan.turns` as well.

examples/sources/openclaw_telemetry_hal/events.py
craig-meridian: **[P2] `_plain_text` diverges from the existing `content_to_text` — a multi-block turn double-faults.**

This re-implements `extraction.content_to_text` with different semantics: space-join vs newline-join, no `type == "text"` filter, and bare-string blocks are skipped. Reproduced: `message.in` `"line1\nline2"` plus the same message re-entering the thread as two text blocks yields TWO user messages with byte-identical `.text` — the operator copy stamped, the real turn unstamped — violating this PR's own "never a duplicate message" invariant. Block-form canonical user turns are reachable (crux1 fixture).

Suggested fix: use `content_to_text(...).strip()` on both sides (its `\n` join makes the multi-line case match) and delete `_plain_text` — parse.py already uses `content_to_text` for exactly this kind of content matching.

examples/sources/openclaw_telemetry_hal/events.py
craig-meridian: **[P2] Unmatched operator messages leak into subsequent `ModelEvent.input`.**

An unmatched `message.in` — by this PR's own definition one that "never entered the session thread" (the model never saw it) — is appended to the running `messages` list, which doubles as `ModelEvent.input` ("conversation up to this turn") for every later orchestrator turn. Reproduced: an unmatched `"secret aside"` at ts=2500 appears in the ts=3000 turn's input. Scanners analyzing what the model was actually shown will see a user message the model never received. The single-thread design already tolerates some divergence (post-compaction inputs are unpruned), but this adds a new fabricated-context case.

Suggested fix: exclude unmatched operator messages from the running input list (thread-only), or document the caveat explicitly for scanner authors.

examples/sources/openclaw_telemetry_hal/events.py
craig-meridian: **[P2] `startswith("[")` rejects the bundled crux1 fixture's standard inline task shape.**

crux1-sample-telemetry.jsonl prompts inline the task as `[Subagent Task]: Check for unread emails …`, which this filter skips — reproduced against the actual fixture prompt: `_task_line` returns the later `"Report back with any unread emails …"` boilerplate as the span name, and `None` (generic `subagent`) when no unbracketed line follows. Either way the actual task text is never chosen, defeating the naming improvement on the very telemetry the example ships with (the new test's prompt happens to put its task on an unbracketed line, so it doesn't catch this).

Suggested fix: handle the `[Subagent Task]:`/`[Subagent Task]` prefix explicitly (strip the bracket tag and keep the remainder, or take the next paragraph), skip only recognized context/timestamp boilerplate, and add a test with a fixture-shaped prompt.

examples/sources/openclaw_telemetry_hal/events.py
craig-meridian, line 170: **[P3] The defensive "spawn call already claimed" anomaly lost its warning.**

The old WARNING covered two distinct cases: no linkable spawn call, and — defensively — a second sub-agent claiming an already-claimed `spawn_tool_call_id` (a violation of the documented 1:1 mapping, i.e. corrupt/ambiguous linkage). Both now funnel into this one info-level "placing by file-order anchor" message, so the genuinely anomalous duplicate-claim case is indistinguishable from the normal cron pattern and invisible at default log levels; the second claimant is quietly anchor-placed, possibly far from its true spawn point.

Suggested fix: keep info for the `tc_id is None` branch, but log a warning when `tc_id` was already claimed.

craig-meridian (changes requested):
Thanks Alex! In principle I think the changes are sound at a high-level. I had Fable 5 do an in-depth code review. The comments are inline. Feel free to push back against any you disagree with.

---

High-effort review of the head `d132ad2a`. 10 findings, each verified by running a concrete reproduction against this branch (in the project venv, importing the example modules directly). The severe end clusters around the two new heuristics: the roll-up aggregate dropper (can both delete a genuine turn and miss a real roll-up) and the operator-channel text-set join (loses repeated interventions, mis-stamps text twins). Details and suggested fixes inline.

examples/sources/openclaw_telemetry_hal/parse.py
alexandraabbas: Fixed in fe847910, though with a different fifth condition than either suggested one — both fail against the real aggregates. Measured on the two real captures (12 dropped aggregates total): **full-usage equality with the previous record holds for 0/12** — an aggregate's components are block *sums*, several-fold larger than the closing record's, with only `totalTokens` frozen — and **text restatement holds for only 7/12** (the other 5 close blocks whose final per-call record has empty text). Either as a hard AND would disable the real detection and bring the phantom usage back.

What does separate them cleanly: a genuine per-call record satisfies `input + output + cacheRead + cacheWrite == totalTokens` (the identity already documented at `usage_to_inspect`; re-verified here across all 10,067 kept turns of both captures — 0 violations), while **all 12 real aggregates violate it** — their components sum the block against a stale total. That identity-break is now the fifth conjunctive condition. Your repro (two distinct, self-consistent turns with equal totals) is kept, and it also fixes a latent false positive on adjacent zero-usage failure placeholders (`0 == 0+0+0+0`). Negative tests added for both (`test_genuine_turns_with_equal_totals_are_kept`, `test_adjacent_zero_usage_placeholders_are_kept`), and the synthetic roll-up in `test_rollup_dropped_and_usage_not_double_counted` is re-authored to the real block-sum shape (it was an exact-copy shape, which real aggregates never take).

Known residual: an aggregate that exactly copies a single self-consistent record (a one-call block) would now be kept, double-counting that one call. Not observed in the captures, and bounded to one per-call usage — versus the reachable real-turn deletion the identity condition prevents.

examples/sources/openclaw_telemetry_hal/parse.py
alexandraabbas: Fixed in fe847910 as suggested: turns group per session kind before the filter runs, and the merged output stays globally timestamp-sorted. The hazard is real — one capture mixes three orchestrator surfaces with ~100 cross-kind adjacencies in merged order — though on both available captures the per-surface pass happens to drop the identical record set (no roll-up block was interleaved in practice, so this is pure hazard removal, no behavior change on real data). `test_rollup_detected_across_interleaved_surfaces` pins your repro shape, including the other surface's genuine turn surviving.

examples/sources/openclaw_telemetry_hal/events.py
alexandraabbas: Fixed in d414b23f with occurrence-based chronological pairing: each send consumes the first unconsumed user turn carrying the same text at-or-after its timestamp; leftover sends surface standalone. Your repro now yields the stamped turn plus the t=2500 standalone (`test_repeated_operator_send_not_lost`), and a twin turn that predates the send is no longer stamped (`test_twin_text_matches_turn_at_or_after_send`).

On the suggested stable-ID/timestamp join — it isn't available in the real data. Across both real captures (64 `message.in` events), `message.in` does carry `metadata.messageId`, but none of the 689 user turns carries an `idempotencyKey` to join against, and timestamp equality only holds on one capture: on the other, the matched turn lags its send by 379ms to ~32 minutes. An equality join would zero-match that capture. At-or-after chronology is the invariant that holds for all 50 real matched pairs, so that's the tiebreak. Verified the new pairing reproduces the old code's outcomes exactly on both captures (44 stamped + 13 standalone, and 6 + 1).

examples/sources/openclaw_telemetry_hal/events.py
alexandraabbas: Fixed in 27dae0da: `_duration_to_ms` coerces via the same try/`int` pattern as `_ts_to_datetime` and rejects non-positive values, so a malformed `tool.end` degrades that one call to zero width instead of aborting the import, and a negative duration can no longer place `completed` before `timestamp`. Both repros are pinned in `test_schema_b_malformed_duration_ms_survives` (the numeric string "250" still yields its 0.25s width).

examples/sources/openclaw_telemetry_hal/events.py
alexandraabbas: Fixed in 27dae0da: strict `<`, and the flush moved to the top of the item loop so it runs on every timeline item — a span now lands immediately after its anchor turn (the documented side), and also before an interleaved compaction rather than being swept past it (the related nit from the first review pass). `test_placed_at_file_order_anchor` now pins exact adjacency and `test_anchor_flush_precedes_interleaved_compaction` pins the compaction side.

On the deeper alternative (unlinked spans as a fourth kind in `ordered`): considered, but anchor_ts=0 spans (sessions first seen before any orchestrator turn) would need special-cased emission timestamps anyway, so it removes one special case and adds another. The top-of-loop flush reaches the same emission points with less machinery; happy to revisit in a follow-up if you prefer that shape.

examples/sources/openclaw_telemetry_hal/parse.py
alexandraabbas: Applied in fe847910: each session's schema-A turns pass through the same filter (with the tightened signature), and `n_assistant_turns` reflects the filtered count. To the question: on the real captures, roll-up-shaped records appear only on orchestrator surfaces — 0 occurrences across the 111 schema-A sub-agent sessions of the richest capture. Filtered defensively anyway, as suggested; `test_subagent_schema_a_rollup_dropped` pins it.

examples/sources/openclaw_telemetry_hal/events.py
alexandraabbas: Fixed in d414b23f: both comparison sides flatten through `content_to_text(...).strip()`; `_plain_text` is deleted (a thin `_match_text` wrapper remains to document why the two sides must share the flattening). Your multi-block repro is pinned in `test_multiline_send_matches_block_content_turn` — one stamped turn, no duplicate.

examples/sources/openclaw_telemetry_hal/events.py
alexandraabbas: Fixed in d414b23f, taking the exclusion option: the running list is split into the thread (returned as `messages`, keeps everything) and the conversation the model was actually shown (feeds `ModelEvent.input`) — unmatched operator messages are thread-only. `test_unmatched_send_not_in_later_model_input` pins your repro, and the README's operator-channel paragraph documents the distinction for scanner authors.

examples/sources/openclaw_telemetry_hal/events.py
alexandraabbas: Fixed in 27dae0da: `_task_line` handles the `[Subagent Task]` tag explicitly — inline remainder wins, else the next non-empty line — before any bracket skipping, so fixture-shaped prompts name from the actual task (`test_span_named_from_inline_subagent_task_tag`, `test_span_named_from_line_after_bare_task_tag`).

One deliberate divergence from "skip only recognized context/timestamp boilerplate": the no-tag fallback keeps the blanket bracket-skip rather than an allowlist, because real prompt bodies carry bracketed template-placeholder lines (fill-in markers, table stubs) that must never become span names — an allowlist would let every unrecognized one through. With the tag handled first, the fallback only sees tag-less prompts. On the real captures the tag path renames 60 of 166 sub-agent spans and none degrade to the generic fallback.

examples/sources/openclaw_telemetry_hal/events.py
alexandraabbas: Fixed in 27dae0da: a second claim on an already-claimed spawn call is anchor-placed with a WARNING naming both session keys and the contested call id; the normal spawn-less pattern stays at info. `test_duplicate_spawn_claim_warns_and_places_by_anchor` pins it.

examples/sources/openclaw_telemetry_hal/events.py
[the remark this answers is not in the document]
alexandraabbas: Fixed in 27dae0da — strict `<`, flush on every timeline item (also settles the compaction note here), exact-adjacency test. Details on the corresponding thread of the follow-up review.

examples/sources/openclaw_telemetry_hal/events.py
```diff
@@ -97,35 +116,64 @@ def build_content(
...
+    # The inbound operator channel (``message.in``). An inbound message whose
+    # text matches a user turn marks THAT turn ``source="operator"`` (no
+    # duplicate message); one with no matching user turn (it never entered the
+    # session thread, e.g. delivered while the agent was busy) becomes its own
+    # operator-sourced user message, placed by its timestamp.
+    operator_contents = {_plain_text(m.get("content")) for m in parse.operator_messages}
+    matched = {
+        _plain_text(t.get("content")) for t in parse.user_turns
+    } & operator_contents
+    extra_operator = [
```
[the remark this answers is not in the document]
alexandraabbas: Fixed in d414b23f — occurrence-based chronological pairing; repeated sends surface standalone. Details on the corresponding follow-up-review thread.

examples/sources/openclaw_telemetry_hal/events.py
```diff
@@ -430,9 +522,17 @@ def _emit_subagent_span(
...
         # real start->end span; fall back to the spawn time when absent (bare
-        # captures), which collapses it to zero duration but never mis-orders it.
+        # captures), which never mis-orders it. Bare captures record no
+        # ``endTs`` either, but ``durationMs`` (from tool.end) still gives the
+        # call a real width — downstream busy-time sums rely on it — so only a
+        # call with neither collapses to zero duration.
         call_start = _ts_to_datetime(call.get("startTs")) or timestamp
-        call_end = _ts_to_datetime(call.get("endTs")) or call_start
+        duration_ms = call.get("durationMs")
+        call_end = _ts_to_datetime(call.get("endTs")) or (
```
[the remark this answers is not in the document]
alexandraabbas: Fixed in 27dae0da — `_duration_to_ms` guard mirroring `_ts_to_datetime`.

examples/sources/openclaw_telemetry_hal/parse.py
```diff
@@ -165,6 +187,50 @@ def _primary_model(orchestrator_turns: list[dict[str, Any]]) -> str | None:
     return winner[0][0] if winner else None
```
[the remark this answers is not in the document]
alexandraabbas: Applied in fe847910 — the filter now runs over schema-A sub-agent turns too. On the real captures roll-ups appear only on orchestrator surfaces (0 in 111 schema-A sessions), so this is defensive parity.

alexandraabbas:
All 10 findings addressed on the branch (fe847910, d414b23f, 27dae0da, + a style commit): suite grows 82 → 95 (every repro from the review is now a pinned test), `ruff check` / `ruff format --check` / `mypy` clean.

Two places the fix deliberately diverges from the suggested mechanism, both evidence-driven — pushback welcome:

- **Roll-up signature**: neither full-usage equality (0/12 real aggregates) nor content restatement (7/12) survives contact with the real captures. The added fifth condition is instead a usage self-consistency check — genuine per-call records satisfy `input+output+cacheRead+cacheWrite == totalTokens` (0 violations across 10,067 kept turns of the two real captures), all 12 real aggregates break it. Details on the thread.
- **Operator reconciliation**: occurrence-based chronological text pairing rather than an ID/timestamp-equality join — real captures carry no `idempotencyKey` on user turns, and send→turn lag runs up to ~32 minutes on one capture, so equality joins would zero-match it. Details on the thread.

End-to-end regression against both real CRUX captures with the final branch: kept-turn sets and billable totals byte-identical to pre-review behavior, operator stamped/standalone outcomes identical; the only intended behavior changes on real data are the span names (60 of 166 sub-agent spans now named from the `[Subagent Task]` tag instead of downstream boilerplate) plus the documented placement side of anchor-placed spans.

*Drafted by Claude Code on behalf of @alexandraabbas.*

examples/sources/openclaw_telemetry_hal/parse.py
```diff
@@ -173,11 +253,21 @@ def parse_telemetry(raw_events: Iterable[dict[str, Any]]) -> OpenClawTelemetry:
...
     c = _consolidate(raw_events)
 
+    # The roll-up filter runs per surface: an aggregate closes a block on ITS
+    # surface, so a turn interleaved from another orchestrator surface (real
+    # captures mix e.g. main/telegram/cron) must not sit between a block's
+    # closing record and its roll-up and mask the frozen-total comparison.
+    turns_by_surface: dict[str, list[dict[str, Any]]] = {}
+    for rid, kind in c.assistant_kind.items():
+        if is_orchestrator(kind):
+            turns_by_surface.setdefault(kind, []).append(c.assistant_by_id[rid])
```
craig-meridian: **[P1] Grouping by session *kind* still lets two concurrent same-kind sessions interleave.**

Two telegram chats (or two cron runs) share one kind, so chat-2's turn can land between chat-1's closing record and its roll-up and mask the frozen-total comparison — reproduced: kept turns `['C1-A','C2-A','C1-A']`, the roll-up surviving. The `_drop_rollup_aggregates` docstring requires "a SINGLE surface's stream", which this caller doesn't establish (telegram session ids are chat ids; multi-chat captures are real).

Fix: group by full `sessionKey` rather than kind.

examples/sources/openclaw_telemetry_hal/parse.py
```diff
@@ -165,6 +188,63 @@ def _primary_model(orchestrator_turns: list[dict[str, Any]]) -> str | None:
...
+            return False
+        if (turn.get("stopReason") or "stop") != "stop":
+            return False
+        if toolcalls_of(turn.get("content")):
+            return False
+        total = (turn.get("usage") or {}).get("totalTokens")
+        prev_total = (prev.get("usage") or {}).get("totalTokens")
+        if total is None or total != prev_total:
+            return False
+        return bool(tokens_from_usage(turn.get("usage")) != total)
```
craig-meridian: **[P2] The identity rescue misfires on atypical usage shapes — genuine turns still dropped.**

Two reproduced cases: (a) `totalTokens` as a numeric string — `"950" == "950"` passes the frozen-total gate, but `tokens_from_usage` returns int, `950 != "950"` breaks the identity, second genuine turn deleted (same raw-JSON hazard `_duration_to_ms` guards against); (b) usage carrying only `{"totalTokens": N}` — components sum to 0 ≠ N, same deletion.

Fix: `int()`-coerce the totals, and treat component-less usage as unclassifiable (keep) rather than identity-broken.

examples/sources/openclaw_telemetry_hal/events.py
```diff
@@ -86,46 +93,119 @@ def _stop_and_error(turn: dict[str, Any]) -> tuple[StopReason, str | None]:
...
+    # after it arrives (observed lags on real captures run from 0 to minutes;
+    # user turns carry no idempotencyKey there, so text is the only join key).
+    # Set-of-texts matching would silently drop a repeated send once its first
+    # occurrence matched, and could stamp a twin turn that predates the send.
+    # A send with no matching turn (it never entered the session thread, e.g.
+    # delivered while the agent was busy) becomes its own operator-sourced
+    # user message, placed by its timestamp.
+    turns_by_text: dict[str, list[dict[str, Any]]] = {}
+    for user_turn in parse.user_turns:  # already timestamp-sorted
+        turns_by_text.setdefault(_match_text(user_turn.get("content")), []).append(
```
craig-meridian: **[P2] Empty text participates in the pairing join.**

A content-less `message.in` (media-only send) flattens to `""` — as does every image-only user turn — so the send stamps the first text-less turn at-or-after it `source="operator"` (reproduced). Fix: skip empty-text sends and turns in the join; an empty send surfaces standalone (or is dropped).

examples/sources/openclaw_telemetry_hal/events.py
```diff
@@ -507,6 +649,56 @@ def _ts_to_datetime(value: Any) -> datetime | None:
...
+    ``[Subagent Task]`` tag: inline (``[Subagent Task]: <task>``) or with the
+    task on the following line. The tag is authoritative when present.
+    Otherwise fall back to the first line that is neither bracketed nor the
+    subagent boilerplate — prompt BODIES also carry bracketed lines (template
+    placeholders), so an unrecognized bracket line never becomes a name.
+    Returns ``None`` when nothing task-bearing is found.
+    """
+    lines = [line.strip() for line in (prompt or "").splitlines()]
+    for i, line in enumerate(lines):
+        if line.startswith(_TASK_TAG):
```
craig-meridian: **[P2] Tag detection is prefix-only, so a mid-line `[Subagent Task]` is missed.**

`"[Fri … PST] [Subagent Task]: Check the unread queue"` — the tag isn't detected, the bracket fallback then skips the whole line, and the span is named from a later body line (reproduced). The fixtures happen to put the timestamp tag on the Context line only, but the shape is documented. Fix: locate the tag with `find(_TASK_TAG)` instead of `startswith`.

examples/sources/openclaw_telemetry_hal/events.py
```diff
@@ -507,6 +649,56 @@ def _ts_to_datetime(value: Any) -> datetime | None:
...
+            inline = line[len(_TASK_TAG) :].lstrip(" :").strip()
+            if inline:
+                return inline
+            return next((follow for follow in lines[i + 1 :] if follow), None)
```
craig-meridian: **[P2] The bare-tag lookahead takes the next non-empty line unfiltered.**

`"[Subagent Task]\n[workspace: /tmp/run-42]\nRun the smaller grids first."` names the span `"[workspace: /tmp/run-42]"` (reproduced) — contradicting the docstring's "an unrecognized bracket line never becomes a name". Fix: apply the same bracket/boilerplate skip to the lookahead as the fallback loop.

examples/sources/openclaw_telemetry_hal/events.py
```diff
@@ -86,46 +93,119 @@ def _stop_and_error(turn: dict[str, Any]) -> tuple[StopReason, str | None]:
...
+        )
+    operator_turn_ids: set[int] = set()
+    extra_operator: list[dict[str, Any]] = []
+    for m in sorted(
+        parse.operator_messages, key=lambda m: int(m.get("timestamp") or 0)
+    ):
+        queue = turns_by_text.get(_match_text(m.get("content")), [])
+        m_ts = int(m.get("timestamp") or 0)
+        turn: dict[str, Any] | None = next(
+            (t for t in queue if int(t.get("timestamp") or 0) >= m_ts), None
```
craig-meridian: **[non-blocking] A timestamp-less user turn can never match any send.**

`int(t.get("timestamp") or 0) >= m_ts` coerces a missing turn timestamp to 0, so the turn stays unstamped and the send duplicates as a standalone operator message (reproduced — pre-fix set-matching handled this shape). Fine as a follow-up: treat a missing turn-ts as always-eligible, or fall back to text-only pairing when the turn carries no timestamp.

examples/sources/openclaw_telemetry_hal/parse.py
```diff
@@ -173,11 +253,21 @@ def parse_telemetry(raw_events: Iterable[dict[str, Any]]) -> OpenClawTelemetry:
...
 
+    # The roll-up filter runs per surface: an aggregate closes a block on ITS
+    # surface, so a turn interleaved from another orchestrator surface (real
+    # captures mix e.g. main/telegram/cron) must not sit between a block's
+    # closing record and its roll-up and mask the frozen-total comparison.
+    turns_by_surface: dict[str, list[dict[str, Any]]] = {}
+    for rid, kind in c.assistant_kind.items():
+        if is_orchestrator(kind):
+            turns_by_surface.setdefault(kind, []).append(c.assistant_by_id[rid])
     orchestrator_turns = sorted(
```
craig-meridian: **[non-blocking] Equal-timestamp cross-surface ordering changed by the regroup/re-merge.**

Ties now order by surface-first-seen instead of file order (reproduced: `['M1','M2','T1']` vs pre-fix `['M1','T1','M2']`), which can also flip `_primary_model`'s first-seen tie-break. One-line determinism restore: carry a file-order index as a secondary sort key through the regroup.

examples/sources/openclaw_telemetry_hal/parse.py
craig-meridian: **[P1] Grouping by session *kind* still lets two concurrent same-kind sessions interleave.**

Two telegram chats (or two cron runs) share one kind, so chat-2's turn can land between chat-1's closing record and its roll-up and mask the frozen-total comparison — reproduced: kept turns `['C1-A','C2-A','C1-A']`, the roll-up surviving. The `_drop_rollup_aggregates` docstring requires "a SINGLE surface's stream", which this caller doesn't establish (telegram session ids are chat ids; multi-chat captures are real).

Fix: group by full `sessionKey` rather than kind. (See also the dedup-identity comment below — the keyless dedup collapses across sessions before this grouping runs, so the two fixes belong together.)

examples/sources/openclaw_telemetry_hal/parse.py
```diff
@@ -387,6 +494,12 @@ def _consolidate(raw_events: Iterable[dict[str, Any]]) -> _Consolidated:
                     if rid not in out.assistant_by_id:
```
craig-meridian, line 543: **[P1, pairs with the grouping fix] The rid-less dedup key is session-blind, so per-session grouping alone isn't sufficient.**

The synthetic key `("a", timestamp, masked_content)` (just above) collapses identical-content same-millisecond turns ACROSS sessions — reproduced: two genuine turns from two telegram sessions retained as one. Snapshot re-dumps (the reason this dedup exists) always share their sessionKey, so adding session identity to the key is safe for its purpose.

Fix: include the sessionKey in the synthetic key, and add a cross-session collision test. Relevant precisely because service-sink captures are rid-less and multiple same-kind sessions are the motivating case.

examples/sources/openclaw_telemetry_hal/events.py
```diff
@@ -144,8 +225,46 @@ def build_content(
...
+        # Emit any unlinked sub-agent spans whose file-order anchor has been
+        # passed, so each lands immediately AFTER the orchestrator turn the
+        # run was on when the session first appeared. Strictly-before (<): a
+        # span whose anchor EQUALS this item's ts belongs on the far side of
+        # the anchor turn — the turn had already been produced when the
+        # session appeared. Running the flush on every timeline item keeps
+        # the span as close to its anchor as possible (in particular, before
+        # an interleaved compaction rather than after it); the trailing loop
+        # below covers anchors at or past the last item.
+        while unlinked and unlinked[0].anchor_ts < ts_ms:
```
craig-meridian, line 258: **[P1] Anchor-placed spans inherit the NEXT timeline item's timestamp, not their anchor's.**

The strict `<` fixed the event ORDER, but the flush passes the current item's `ts` to `_emit_subagent_span`: with A1@1000, A2@2000 (anchor), A3@3000, the span begin and its schema-B tool are stamped 3000 — reproduced; and a schema-A child event at 2100 now predates its own SpanBeginEvent. Timing shifts systematically for every spawn-less session (46/101 on the real capture).

Fix: keep the insertion point, but derive the emitted timestamp from `sa.anchor_ts` (falling back to the current ts only for a zero/invalid anchor), and extend the placement test to assert span/tool timestamps, not just adjacency.

examples/sources/openclaw_telemetry_hal/parse.py
craig-meridian: **[P2] The identity rescue misfires on atypical usage shapes — genuine turns still dropped.**

Two reproduced cases: (a) `totalTokens` as a numeric string — `"950" == "950"` passes the frozen-total gate, but `tokens_from_usage` returns int, `950 != "950"` breaks the identity, second genuine turn deleted (same raw-JSON hazard `_duration_to_ms` guards against); (b) usage carrying only `{"totalTokens": N}` — components sum to 0 ≠ N, same deletion.

Fix: `int()`-coerce the totals, and treat component-less usage as unclassifiable (keep) rather than identity-broken.

examples/sources/openclaw_telemetry_hal/events.py
craig-meridian: **[P2] Empty text participates in the pairing join.**

A content-less `message.in` (media-only send) flattens to `""` — as does every image-only user turn — so the send stamps the first text-less turn at-or-after it `source="operator"` (reproduced). Fix: skip empty-text sends and turns in the join; an empty send surfaces standalone (or is dropped).

examples/sources/openclaw_telemetry_hal/events.py
craig-meridian: **[P2] Tag detection is prefix-only, so a mid-line `[Subagent Task]` is missed.**

`"[Fri … PST] [Subagent Task]: Check the unread queue"` — the tag isn't detected, the bracket fallback then skips the whole line, and the span is named from a later body line (reproduced). The fixtures happen to put the timestamp tag on the Context line only, but the shape is documented. Fix: locate the tag with `find(_TASK_TAG)` instead of `startswith`.

examples/sources/openclaw_telemetry_hal/events.py
craig-meridian: **[P2] The bare-tag lookahead takes the next non-empty line unfiltered.**

`"[Subagent Task]\n[workspace: /tmp/run-42]\nRun the smaller grids first."` names the span `"[workspace: /tmp/run-42]"` (reproduced) — contradicting the docstring's "an unrecognized bracket line never becomes a name". Fix: apply the same bracket/boilerplate skip to the lookahead as the fallback loop.

examples/sources/openclaw_telemetry_hal/events.py
craig-meridian: **[non-blocking] A timestamp-less user turn can never match any send.**

`int(t.get("timestamp") or 0) >= m_ts` coerces a missing turn timestamp to 0, so the turn stays unstamped and the send duplicates as a standalone operator message (reproduced — pre-fix set-matching handled this shape). Fine as a follow-up: treat a missing turn-ts as always-eligible, or fall back to text-only pairing when the turn carries no timestamp.

examples/sources/openclaw_telemetry_hal/parse.py
craig-meridian, line 300: **[non-blocking] Equal-timestamp cross-surface ordering changed by the regroup/re-merge.**

Ties now order by surface-first-seen instead of file order (reproduced: `['M1','M2','T1']` vs pre-fix `['M1','T1','M2']`), which can also flip `_primary_model`'s first-seen tie-break. One-line determinism restore: carry a file-order index as a secondary sort key through the regroup.

craig-meridian (changes requested):
Thanks for the fixes Alex! I've done another review round with Fable 5 + Sol 5.6 - but I don't intend this to be a never-ending cycle. Most findings relate to the code just pushed and the severity of items is generally dropping.

---

Second pass over the fix commits (`fe847910`..`96959a35`). All ten first-round findings verified fixed — repros re-run against the new head, threads resolved. Nice work, especially the real-capture measurements behind the usage-identity condition and the at-or-after pairing.

Note on scope: these comments are (with one partial exception) about code **introduced by the fix commits**, not misses from the first round — and they're narrower than round one: mostly shapes unobserved in the bundled captures. Two of the nine (anchor timestamps, dedup identity) travel together with the interleaving fix, so tackling those three as one unit should avoid another round. Only the interleaving/anchor pair is what I'd fix before merge; the rest are one-to-few-line tightenings, and the ones marked non-blocking can be follow-ups. Each repro shape is stated compactly — your agent can reconstruct the full case from it.

examples/sources/openclaw_telemetry_hal/parse.py
alexandraabbas: Fixed in c145a8c9: the filter groups by full `sessionKey`. The shape is very reachable — the PFN capture carries **419 distinct orchestrator sessionKeys** (mostly per-run cron sessions of 3–9 records each), so kind-grouping was effectively merging hundreds of streams. Drop sets on both real captures are unchanged under session grouping (identical 9/3, same records — no roll-up block was interleaved in practice), so this is hazard removal with no real-data behavior change. Your two-chat repro is pinned in `test_rollup_detected_across_same_kind_sessions`, and the `_drop_rollup_aggregates` docstring now states the per-session contract its caller actually establishes.

examples/sources/openclaw_telemetry_hal/parse.py
alexandraabbas: Fixed in c145a8c9, together with the grouping change: the orchestrator keyless key is now `(session, timestamp, masked content)`, mirroring the sub-agent branch's key (which already carried its sessionKey). Measured before changing: **0** keyless `(ts, content)` keys appear under more than one sessionKey across both real captures, so the session term un-collapses nothing real while removing the cross-session merge. Collision test added (`test_keyless_turns_do_not_collapse_across_sessions`), and the module docstring's dedup-key description updated.

examples/sources/openclaw_telemetry_hal/events.py
alexandraabbas: Fixed in c145a8c9: both emission sites (the per-item flush and the trailing loop) stamp the span at `_ts_to_datetime(sa.anchor_ts)`, falling back to the current item's ts only for a zero/invalid anchor — so a schema-A child event can no longer predate its own SpanBeginEvent, and spawn-less sessions stop reporting systematically late starts. `test_placed_at_file_order_anchor` now asserts the span begin AND its ts-less schema-B call are stamped at the anchor time (the schema-B fallback inherits the span time, so both moved together).

examples/sources/openclaw_telemetry_hal/parse.py
alexandraabbas: Fixed in e50ccc0a: totals are coerced through the same try/`int` pattern as `_duration_to_ms` (`_int_or_none`; a non-numeric total is unclassifiable → keep), and a usage with no component fields is treated as unclassifiable rather than identity-broken — the filter drops only on positive evidence of restated usage. Both repros are pinned: `test_string_total_tokens_coerced_not_identity_broken` (which also pins that a block-sum roll-up with string totals is STILL dropped) and `test_component_less_usage_is_unclassifiable_and_kept`. On the real captures neither shape occurs (0 string totals, 0 component-less usages), so no behavior change there.

examples/sources/openclaw_telemetry_hal/events.py
alexandraabbas: Fixed in e50ccc0a: empty text never enters the join on either side — empty-text user turns are excluded from the queues, and an empty send goes straight to standalone. Kept standalone rather than dropped: it still records that an operator intervened at that time, with the text simply not captured. `test_empty_send_does_not_stamp_media_only_turn` pins your media-only repro.

examples/sources/openclaw_telemetry_hal/events.py
alexandraabbas: Fixed in e50ccc0a with `find()` as suggested (`test_mid_line_task_tag_detected`). Extending the tests surfaced an adjacent shape the round-1 code also mishandled: real prompts REPEAT the tag — a bare `[Subagent Task]` heading first, then a `[Subagent Task]: <task>` line with the task inline (25 + 34 prompts across the two real captures; the dominant tag shape). A bare tag's lookahead therefore runs only up to the next tag line and defers to it (`test_bare_tag_defers_to_later_inline_tag`). Net effect on the real captures: those 59 spans are now named from the actual inline task text — the round-1 code had been returning the raw tag-prefixed line for them.

examples/sources/openclaw_telemetry_hal/events.py
alexandraabbas: Fixed in e50ccc0a: the lookahead applies the same bracket/boilerplate skip as the no-tag fallback (shared `_is_task_bearing` predicate), so a workspace tag or template placeholder can never become a span name — restoring the docstring's invariant. Your repro is pinned in `test_bare_tag_lookahead_skips_bracketed_lines`.

examples/sources/openclaw_telemetry_hal/events.py
alexandraabbas: Took this now rather than as a follow-up (e50ccc0a): a turn with no timestamp is always-eligible in the pairing — the pre-pairing set semantics matched it, and a standalone duplicate is worse than the loose match. `test_ts_less_turn_still_matched_by_text` pins it.

examples/sources/openclaw_telemetry_hal/parse.py
alexandraabbas: Took this now rather than as a follow-up (c145a8c9): a file-order index (insertion order of the consolidated turn map, which is first-occurrence file order) rides through the regroup as the secondary sort key, so equal-timestamp merges — and `_primary_model`'s first-seen tie-break — are deterministic in file order again. Your M1/T1/M2 shape is pinned in `test_equal_ts_ties_keep_file_order`.

alexandraabbas:
All nine round-2 findings addressed in three commits (c145a8c9 — the interleaving/dedup/anchor unit you asked to travel together, plus the tie-order determinism; e50ccc0a — the P2 tightenings and both non-blocking items, taken now rather than as follow-ups; ed24220d — see below). Suite grows 95 → 109; `ruff` / `ruff format` / `mypy` clean; end-to-end regression against both real captures is byte-identical on kept turns, billable totals, and operator outcomes.

Two things beyond the letter of the findings, in the same spirit:

- **Session-scoping applied to user turns too** (ed24220d): your dedup-identity argument for assistant turns applies verbatim to the keyless user-turn key and the transient-twin text drop twenty lines below — an adversarial pass we ran over our own fix commits flagged leaving them session-blind as incoherent. Both are now session-scoped (behavior-neutral on the real captures: no keyed user turns, no cross-session collisions), and the same pass produced two more corrections: the ts-less operator turn is now a fallback rather than always-preferred (greedy-first could manufacture the very duplicate the rule exists to avoid), and the trailing-flush half of the anchor-time stamp gained the test it was missing.
- **Tag-shape evidence** (e50ccc0a): extending the `find()` tests surfaced that real prompts repeat the tag — bare `[Subagent Task]` heading first, `[Subagent Task]: <task>` after (25 + 34 prompts across the two captures) — so the bare tag's lookahead defers to the next tag line; those 59 spans now name from the actual inline task text (the round-1 code had been returning the raw tag-prefixed line for them).

Understood on not wanting a never-ending cycle — nothing here was deferred to a follow-up.

*Drafted by Claude Code on behalf of @alexandraabbas.*

craig-meridian (approved):
Thanks very for addressing all the points and then some. Have done a final review pass with 5.6 Sol and I'm happy.
I'll leave this for you to merge once main is merged in - lmk if you need me to merge though.

PR #568, opened 2026-08-18
fix: resolve attachments in pooled event messages

Rogan-Inglis:
## Summary

- retain attachments while streaming whenever events are requested
- resolve attachment references that appear only in later `events_data.messages` pools
- add an end-to-end regression test for eval-log field ordering

## Root cause

Eval-log JSON stores `attachments` before `events_data`. The streaming loader selectively retained only attachment IDs already observed, but references inside the later event message pool were not known until after the attachment section had passed. Pool expansion therefore produced unresolved `attachment://...` content.

## Validation

- `tests/scanner/test_load_filtered.py`: 32 passed, 1 skipped
- Ruff lint and format checks pass for all touched files
- validated against a real affected eval log: both transcript messages and the first agent `ModelEvent.input` now contain resolved prompt text rather than attachment URIs

### DIFF
--- a/src/inspect_scout/_transcript/json/load_filtered.py
+++ b/src/inspect_scout/_transcript/json/load_filtered.py
@@ -300,7 +300,10 @@ async def _parse_and_filter(
     )
     events_coro = event_item_coroutine(state, events_config) if events_config else None
     timelines_coro = timeline_item_coroutine(state)
-    attachments_coro = attachments_coroutine(state)
+    # Event message pools are stored after attachments in eval-log JSON. Their
+    # attachment references therefore are not known while the attachment section
+    # is streaming past, so retain every attachment whenever events are requested.
+    attachments_coro = attachments_coroutine(state, retain_all=events_coro is not None)
     metadata_coro = metadata_coroutine(state)
     target_coro = target_coroutine(state)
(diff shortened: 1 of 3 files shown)


epatey:
Thanks for this, and welcome. I want to make sure we land this at the right layer before reviewing.

In theory the bug is real: if the pooled events_data streams after the attachments section, refs that appear only in those pooled messages aren't known when attachments stream past, so they'd be dropped. But the loader's invariant runs the other way. load_filtered_transcript's docstring says the attachments section "follows events," so with a well-formed log every event ref is already known by the time attachments stream, and the selective retention is correct. So in practice this shouldn't arise.

You mention validating against a real affected log. Could you share it, or a minimal sample from it? If that log really does put events_data after attachments, that reads as a producer emitting sections out of order, and the fix belongs there rather than a retain-all in the loader (which gives back the streaming memory optimization on every request that includes events). If the ordering is actually fine and refs still drop, then the cause is something else and worth pinning down.

Either way the regression test and the writeup are helpful. Once we know which case the real log is, this should be quick.

nix-aisi:
My impression is that this is fixed by https://github.com/meridianlabs-ai/inspect_scout/pull/633

PR #572, opened 2026-08-21
fix: deliver the final metrics update synchronously at strategy teardown

### LINKED ISSUE #569: MP worker subprocesses drop their final metrics update
Found by @epatey while reviewing #564 (https://github.com/meridianlabs-ai/inspect_scout/pull/564#issuecomment-5332612335).

Worker subprocesses in the multiprocess strategy never call `set_background_task_group`, so the throttle's deferred fire falls back to a bare `asyncio.create_task`. The worker sends `WorkerComplete` and its loop tears down, anyio cancels the still-sleeping task, and the final zeroed metrics never reach the parent. Cosmetic impact, the display shows stale per-worker metrics at worker exit.

Related single-process observation from testing #564, verified with an instrumented run. The single-process strategy's finally issues its final zeroed metrics update through the same throttled callback, so it is only delivered when that call lands outside the throttle window of the previous fire. On a short completed scan with the real 1s window the zeroed update was dropped 3 of 3 runs (the last store write still had process_count=1), and with throttling removed it was delivered 3 of 3. If there is appetite for a structural fix, delivering the final update synchronously in the finally (the store is still open at that point) and then disarming the callback would cover both paths at the source.

### LINKED ISSUE #578: Ctrl-C during a scan exits with AssertionError instead of an interrupted status
Ctrl-C during a scan does not produce the interrupted exit the code intends. Instead it prints about 80 lines of traceback, 2 of them chained, through click, anyio and asyncio internals, ending in

```
AssertionError: scan async did not return a result.
```

To reproduce, start a scan long enough to interrupt and press Ctrl-C once after the progress panel appears. On current main (`bbacfbb2`) it happens every time, 3 of 3 runs with `--max-processes 1` and 2 of 2 on the default multiprocess path, exit code 1. A programmatic `scan()` call fails the same way.

An instrumented run shows the mechanism. The interrupt cancels the scan's task group and the cancellation handler at `_scan.py:818` is entered, but `handle_scan_interrupted` awaits `recorder.sync()` inside the already cancelled scope, so its first await re-raises the cancellation immediately. The "Aborted!" message it was called with never reaches the display. `result` is never assigned, `_scan_async` swallows the cancellation at `_scan.py:499` on the assumption that the inner function handled it, and execution trips the `assert result is not None` at `_scan.py:505`.

So the intended aborted-status path exists but cannot complete under a real interrupt. Shielding the final sync from cancellation would be one way to let it finish. Happy to turn this into a PR if that direction sounds right.

I ran into this while probing the cancellation path for #558 (context in #564).

sunnyadn:
Fixes #569.
Fixes #578.

fix: complete the interrupted-status sync under Ctrl-C (#578)

**What happens**

The final zeroed metrics update goes out through the same `@throttle(1)` callback as every other update, so when it lands inside the throttle window it becomes a pending trailing-edge fire that has to survive teardown. It does not. In a worker subprocess there is no background task group, so the fire is a bare `asyncio.create_task` that dies when the worker's loop exits after `WorkerComplete`, and the parent goes on counting that worker's last non zero snapshot in its sum for the rest of the run. Scanning 4 transcripts with 2 workers where one takes much longer, the first worker finished about 2 seconds in and the display kept reporting 2 processes and 1 transcript parsing for the remaining 12 seconds, 3 of 3 runs. The single process path drops the update too, invisibly, since the store entry is deleted right after.

**The fix**

The `finally` now delivers the final update by calling the callback directly instead of through the throttle, then sets a flag so a pending fire cannot deliver a duplicate afterwards. One choke point covers both paths.

Calling `set_background_task_group` in the worker instead would give the fire somewhere to live, but it makes the worker task group wait out the throttle window on every exit and leaves the single process path unfixed.

**Verification**

Two tests through the public `scan()` API, red 3 of 3 on main and green 3 of 3 each. The multiprocess one runs the real throttle window, so its red direction depends on teardown landing inside that window. The green direction does not, because the delivery is synchronous and ordered ahead of `WorkerComplete`. #564's tests still pass, and so does the full suite.

The diff also lifts the store spying helper that #564's test had inline into `tests/helpers.py`, so both use one copy.

cc @epatey

### DIFF
--- a/src/inspect_scout/_concurrency/single_process.py
+++ b/src/inspect_scout/_concurrency/single_process.py
@@ -136,14 +136,22 @@ def single_process_strategy(
         def _scanner_job_info(item: ScannerJob) -> str:
             return f"{item.union_transcript.transcript_id, item.scanner_name}"
 
-        @throttle(1)
-        def _update_metrics() -> None:
-            if update_metrics:
+        # the finally at the end of the_func delivers the final zeroed update
+        # itself, and sets this so a trailing-edge fire still pending in the
+        # throttle does not deliver a duplicate of it after teardown
+        final_metrics_delivered = False
+
+        def _update_metrics_now() -> None:
+            if update_metrics and not final_metrics_delivered:
                 # USS - Unique Set Size
                 metrics.memory_usage = process.memory_full_info().uss
                 # print(f"{diag_prefix} CPU {metrics.cpu_use}")
                 metrics.buffered_scanner_jobs = len(scanner_job_deque)
                 update_metrics(metrics)
+
+        @throttle(1)
+        def _update_metrics() -> None:
+            _update_metrics_now()
 
         def _choose_next_action() -> Literal["parse", "scan", "wait"]:
             """Decide what action this worker should take: 'parse', 'scan', or 'wait'."""
@@ -347,6 +355,10 @@ def single_process_strategy(
             metrics.tasks_parsing = 0
             metrics.tasks_scanning = 0
             metrics.tasks_idle = 0
-            _update_metrics()
+            # deliver synchronously, since a throttled call would defer to a
+            # trailing-edge fire that may not survive teardown (#569). nothing
+            # awaits between the delivery and the flag, so no fire interleaves
+            _update_metrics_now()
+            final_metrics_delivered = True
 
     return the_func
--- a/src/inspect_scout/_concurrency/single_process.py
+++ b/src/inspect_scout/_concurrency/single_process.py
@@ -138,11 +138,13 @@ def single_process_strategy(
 
         # the finally at the end of the_func delivers the final zeroed update
         # itself, and sets this so a trailing-edge fire still pending in the
-        # throttle does not deliver a duplicate of it after teardown
+        # throttle does not deliver a duplicate of it after teardown. the guard
+        # lives in the throttled wrapper, so the suppression holds regardless of
+        # the order of the delivery and the flag in that finally
         final_metrics_delivered = False
 
         def _update_metrics_now() -> None:
-            if update_metrics and not final_metrics_delivered:
+            if update_metrics:
                 # USS - Unique Set Size
                 metrics.memory_usage = process.memory_full_info().uss
                 # print(f"{diag_prefix} CPU {metrics.cpu_use}")
@@ -151,7 +153,8 @@ def single_process_strategy(
 
         @throttle(1)
         def _update_metrics() -> None:
-            _update_metrics_now()
+            if not final_metrics_delivered:
+                _update_metrics_now()
 
         def _choose_next_action() -> Literal["parse", "scan", "wait"]:
             """Decide what action this worker should take: 'parse', 'scan', or 'wait'."""
@@ -356,8 +359,8 @@ def single_process_strategy(
             metrics.tasks_scanning = 0
             metrics.tasks_idle = 0
             # deliver synchronously, since a throttled call would defer to a
-            # trailing-edge fire that may not survive teardown (#569). nothing
-            # awaits between the delivery and the flag, so no fire interleaves
+            # trailing-edge fire that may not survive teardown (#569). the flag
+            # suppresses any such fire that is still pending
             _update_metrics_now()
             final_metrics_delivered = True
 
--- a/src/inspect_scout/_scan.py
+++ b/src/inspect_scout/_scan.py
@@ -892,7 +892,12 @@ def init_scan_model_context(
 async def handle_scan_interrupted(
     message_or_exc: str | Exception, spec: ScanSpec, recorder: ScanRecorder
 ) -> Status:
-    scan_status = await recorder.sync(await recorder.location(), complete=False)
+    # on interrupt this runs inside an already-cancelled scope, where an
+    # unshielded await re-raises the cancellation before the sync can
+    # complete (#578) -- shield it so the interrupted status gets written
+    # and reported rather than tripping the not-None assert upstream
+    with anyio.CancelScope(shield=True):
+        scan_status = await recorder.sync(await recorder.location(), complete=False)
     display().scan_interrupted(message_or_exc, scan_status)
     return scan_status
 
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -3,10 +3,14 @@
 import secrets
 from contextlib import contextmanager
 from pathlib import Path
-from typing import Iterator
+from typing import Any, Callable, ContextManager, Iterator
 from unittest.mock import patch
 
 from inspect_ai._util.kvstore import inspect_kvstore
+from inspect_scout._recorder.active_scans_store import (
+    ActiveScansStore,
+    active_scans_store,
+)
 
 
 @contextmanager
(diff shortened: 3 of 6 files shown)


epatey:
@auto

i-am-marvin:
**Claude finished @epatey's task in 4m 12s** —— [View job](https://github.com/meridianlabs-ai/inspect_scout/actions/runs/33098801003)

---
### Review

- [x] Gather context (diff, related code, issue #569 background)
- [x] Review the fix in `single_process.py`
- [x] Review the tests
- [x] Post review feedback

**Verdict: the fix is correct, well-scoped, and lands the structural fix proposed in #569 at the right choke point.** I traced the delivery path end to end in both strategies and found no correctness issues — only a couple of optional nits below.

### Why the fix holds

- **Synchronous delivery is sound.** `_update_metrics_now()` (`src/inspect_scout/_concurrency/single_process.py:364`) and `final_metrics_delivered = True` run back-to-back with no `await` between them, so on a single-threaded event loop no pending trailing-edge fire can interleave. Any fire that runs *before* the `finally` delivers a superseded intermediate snapshot; any fire after is suppressed by the flag check inside the throttled wrapper (`single_process.py:156`). Both orderings are safe.
- **Flag scoping is correct.** `final_metrics_delivered` is defined and assigned in `the_func`'s own scope, so no `nonlocal` is needed, and each strategy invocation gets a fresh flag.
- **MP ordering is guaranteed.** The worker's `finally` runs inside `_work_task` before it returns, so the final `MetricsItem` is enqueued before `WorkerComplete` (`_mp_subprocess.py:183-185`) on the same FIFO queue. The parent's collector therefore sums the zeroed snapshot before counting the worker finished — exactly what the test's green direction relies on.
- **Bonus not mentioned in the PR description:** the synchronous call is also cancellation-safe. On the interrupted path the `finally` runs during unwind, where an `await` would be hazardous; a plain sync call delivers reliably (the store is still open — `scan_finished` isn't set until `_scan.py:808`).
- **Batch callback can't re-fire after teardown** — `set_batch_status_callback(None)` precedes the final delivery in the same `finally`.
- The single-process test's monkeypatch of `single_process.throttle` works because the decorator is applied at `the_func` execution time, and the MP test correctly leaves workers on the real 1s window (spawned subprocesses don't inherit the patch anyway).

The test-helper lift (`active_scans_store_spy`) is a clean consolidation, and reading delivered state back out of the store via `read_all()` rather than trusting the callback argument is the right observability choice — a write swallowed by a closed store can't masquerade as delivered.

### Minor observations (optional)

1. **Comment drift in `_put_upstream_or_drop`** (`_mp_subprocess.py:209-221`): it still explains that the *final* metrics snapshot can be lost with no successor to supersede it. That's now only true for the queue-full drop (maxsize 100 — rare at teardown); the throttle-swallowed drop this comment implicitly covered is fixed here. A one-line touch-up would keep it accurate. [Fix this →](https://claude.ai/code?q=On%20branch%20fix%2F569-final-metrics-update%2C%20update%20the%20comment%20in%20src%2Finspect_scout%2F_concurrency%2F_mp_subprocess.py%20lines%20209-221%20(_put_upstream_or_drop)%3A%20the%20final%20metrics%20snapshot%20is%20now%20delivered%20synchronously%20at%20strategy%20teardown%20(PR%20%23572%2C%20issue%20%23569)%2C%20so%20the%20only%20remaining%20loss%20path%20is%20a%20full%20upstream%20queue%20at%20put_nowait%20time%2C%20not%20the%20throttle.%20Keep%20the%20warning%20about%20not%20making%20the%20put%20blocking.&repo=meridianlabs-ai/inspect_scout)

2. **`typing.ContextManager` in `tests/helpers.py:6`** is a deprecated alias; `contextlib.AbstractContextManager` is the modern spelling and fits the repo's modern-syntax typing rule. Ruff evidently doesn't flag it, so purely a nit.

3. The refactored cleanup test now records `deleted_at` *after* `delete_current()` runs (the spy calls `on_call` post-method) where the inline version recorded it before. That marginally tightens the `max(fire_times) > deleted_at[0]` precondition, and is arguably the more accurate semantics — just noting the behavior change was intentional-looking and benign.

### Residual limitation (acknowledged, not a defect)

A worker's final snapshot can still be dropped if the upstream queue is full at teardown (`put_nowait` + drop). That's pre-existing, deliberately documented behavior, and the new MP test's docstring correctly scopes its guarantee to "the upstream queue has room."

### CI

All lint/typecheck/build/schema checks pass on the head commit; the two `py-test` jobs were still running at review time.

---

epatey:
Assess if this is closely related to #569. If so, let's combine its fix into this pr.

epatey:
@i-am-marvin, Assess if this is closely related to https://github.com/meridianlabs-ai/inspect_scout/issues/569. If so, let's combine its fix into this pr.

i-am-marvin:
**Claude finished @epatey's task in 19m 52s** —— [View job](https://github.com/meridianlabs-ai/inspect_scout/actions/runs/33100218003)

---
### Assessment: yes — closely related, and the fix is ready to fold in

- [x] Gather context (issue #569, candidate issues, code)
- [x] Determine the referent and assess relatedness
- [x] Implement the fix + test, verify red/green, run full suite
- [x] Push the commit (to a base-repo branch — see delivery note below)

### Which issue "this" is

The comment links #569, but that's the issue this PR already fixes — so I take the question to be about **#578** (*Ctrl-C during a scan exits with AssertionError instead of an interrupted status*). It's the natural referent: filed by @sunnyadn out of the same #558/#564 investigation that produced #569, and its author offered to turn it into a PR. I also considered #577 (per-result `model_usage` inflation), but that's a usage-accounting bug in the result-recording loop — different mechanism, different subsystem, not a fit for this PR.

### Why #578 is closely related to #569

Both are the same failure class at adjacent layers: **end-of-scan finalization work handed to a context that cannot complete it.**

- #569: the strategy's final metrics update becomes a throttle trailing-edge task that dies when the loop tears down. This PR fixes it by delivering synchronously in the strategy's `finally`.
- #578: the scan's interrupted-status sync (`handle_scan_interrupted`, `src/inspect_scout/_scan.py:892`) runs inside the already-cancelled scope after Ctrl-C, so its first `await` re-raises the cancellation before `recorder.sync()` completes — `result` is never assigned and the scan dies on `assert result is not None, "scan async did not return a result."` (`_scan.py:511`).

Together they make teardown coherent end to end: under Ctrl-C, the strategy's `finally` now delivers the final zeroed metrics synchronously (this PR), and the shielded sync then writes and reports the interrupted status (#578's fix). Combining them in one PR is defensible; they're one story — "scan teardown survives cancellation."

### What I implemented

Commit `fdce838b`, exactly the direction #578 proposed:

- **`src/inspect_scout/_scan.py`** — wrap the sync in `handle_scan_interrupted` in `anyio.CancelScope(shield=True)` so it completes inside the cancelled scope. This is the parent-process handler, so it covers both the single-process and multiprocess paths reported in #578.
- **`tests/scan/test_scan_interrupted.py`** — cancels a genuinely-running scan through the public `scan_async` API and asserts an interrupted (`complete=False`) status comes back. **Red** on the unshielded code with exactly the issue's error (`AssertionError: scan async did not return a result.`); **green 3/3** with the fix. Full suite: 3194 passed, 107 skipped; `mypy` and `ruff` clean (against inspect_ai git main, matching CI).

### Delivery note — one command needed from @sunnyadn

This PR's head branch lives on @sunnyadn's fork, which I can't push to. The commit is on [`fix/578-interrupted-status-sync`](https://github.com/meridianlabs-ai/inspect_scout/tree/fix/578-interrupted-status-sync) in this repo. To combine it into this PR:

```bash
git pull https://github.com/meridianlabs-ai/inspect_scout.git fix/578-interrupted-status-sync
git push
```

Then add `Fixes #578` to the PR description (and, since we squash-merge, consider a body line like `fix: complete the interrupted-status sync under Ctrl-C (#578)` so Release Please records both fixes). If I've guessed the referent wrong and you meant a different issue, the branch can simply be deleted — say the word and I'll reassess.
 · [Fix branch](https://github.com/meridianlabs-ai/inspect_scout/tree/fix/578-interrupted-status-sync)

epatey (approved):
Thank you

PR #579, opened 2026-08-23
feat: typed parsed answers on scanner results

kaifronsdal:
## What

`Result` gains a `parsed` field carrying the typed parsed answer, and `generate_answer()` / `parse_answer()` type it based on the answer spec you pass:

| answer spec | return type | `.parsed` |
|---|---|---|
| `AnswerStructured(Detection)` | `Result[Detection]` | `Detection \| None` |
| `AnswerStructured(list[Detection])` | `Result[list[Detection]]` | `list[Detection] \| None` |
| `"boolean"` / `"numeric"` / `"string"` | `Result[bool/float/str]` | `bool/float/str \| None` |
| `list[str]` (single label) | `Result[str]` | chosen label text |
| `AnswerMultiLabel` | `Result[list[str]]` | chosen labels |

`parsed` is the pre-`value_to_float` answer and is `None` whenever the response could not be parsed. For `AnswerStructured`, instances are of the exact declared class (not the internal explanation-augmented subclass), so they are picklable for multiprocess scanning. `AnswerStructured` is now generic in its answer type, which also makes invalid specs like `AnswerStructured(int)` or `AnswerStructured(list[str])` type-check errors instead of runtime errors.

## Backward compatibility

- `Result`'s type parameter defaults to `Any` (PEP 696 via `typing_extensions.TypeVar`), so bare `Result` annotations everywhere — including user scanners — are unchanged under strict mypy.
- `parsed` is `Field(exclude=True)` + `SkipJsonSchema`: it never appears in `model_dump`/JSON/stored results, and `openapi.json` is byte-identical (no TS regeneration needed).
- All existing parse behavior is preserved exactly, including failure values (`value=False` for boolean/numeric failures, `value=None` elsewhere). Verified with a differential battery of 18 parse cases producing byte-identical serialized output vs `main`.

## Testing

- New `tests/llm_scanner/test_parsed.py`: parsed values for every answer kind, `None` on parse failures, pre-`value_to_float` semantics, declared-type instances + pickling, resultsets, and serialization/schema exclusion.
- Full suite green locally (3,197 passed / 0 failed), strict mypy clean on `src`/`examples`/`tests`, ruff clean.

### DIFF
--- a/docs/llm_scanner.qmd
+++ b/docs/llm_scanner.qmd
@@ -257,8 +257,10 @@ Scanners produce results which ultimately carry a specific `value`. There are tw
     class ToolErrors(BaseModel):
         tool_errors: int = Field(alias="value", description="The number of tool errors encountered.")
 
-        causes: str = Field(description="What were the most common causes of tool errors.") 
+        causes: str = Field(description="What were the most common causes of tool errors.")
     ```
+
+In addition to the serialized `value`, results carry an in-memory `parsed` field with the validated instance of your `BaseModel` type — useful when working with `generate_answer()` directly in a custom scanner (see [Scanner Tools](scanner_tools.qmd#answer-generation)). Note that `parsed` is not stored in the results database.
 
 ### Labels and Explanations
 
--- a/docs/llm_scanner.qmd
+++ b/docs/llm_scanner.qmd
@@ -260,8 +260,6 @@ Scanners produce results which ultimately carry a specific `value`. There are tw
         causes: str = Field(description="What were the most common causes of tool errors.")
     ```
 
-In addition to the serialized `value`, results carry an in-memory `parsed` field with the validated instance of your `BaseModel` type — useful when working with `generate_answer()` directly in a custom scanner (see [Scanner Tools](scanner_tools.qmd#answer-generation)). Note that `parsed` is not stored in the results database.
-
 ### Labels and Explanations
 
 We've noted the special `alias="value"` field annotation that promotes one of your `BaseModel` fields to be the main `value` for the returned result. In addition, there are two other special field annotations:
--- a/docs/scanner_tools.qmd
+++ b/docs/scanner_tools.qmd
@@ -209,7 +209,7 @@ result = await generate_answer(
 
 : {tbl-colwidths=\[25,75\]}
 
-In addition to the serialized `value`, the returned `Result` carries an in-memory `parsed` field with the typed parsed answer — `bool` for `"boolean"`, `float` for `"numeric"`, `str` for `"string"` and single-label answers, `list[str]` for multi-label answers, and validated instances of your Pydantic type for `AnswerStructured` (where the return type is inferred from the answer specification, e.g. `Result[Detection]` from `AnswerStructured(Detection)`). It is `None` whenever the response could not be parsed, and is excluded from serialization, so it does not survive storage of results.
+In addition to the serialized `value`, the returned `Result` carries an in-memory `parsed` field with the typed parsed answer — `bool` for `"boolean"`, `float` for `"numeric"`, `str` for `"string"` and single-label answers, `list[str]` for multi-label answers, and validated instances of your Pydantic type for `AnswerStructured` (where the return type is inferred from the answer specification, e.g. `Result[Detection]` from `AnswerStructured(Detection)`). It is `None` whenever the response could not be parsed, and is excluded from serialization and pickling, so it does not survive storage of results or crossing process boundaries.
 
 ## Answer Parsing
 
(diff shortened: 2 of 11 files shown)


kaifronsdal:
Addressed all review findings in bf70daee:

1. **Serialization round trip (correctness):** agreed and fixed as suggested — `as_declared_type` now builds the declared instance from the augmented instance's already-validated field values (`dict(obj)` + drop the injected `explanation`) instead of `model_dump()` → `model_validate()`. Both reported shapes reproduced before the fix and are now covered by regression tests (`test_parsed_structured_with_non_roundtripping_serializer`, `test_parsed_structured_with_required_excluded_field`).
2. **Resultset double work:** fixed — the resultset now reuses the per-item `parsed` instances already computed by `create_result_from_parsed`.
3. **Docstring precision:** both points taken. The `parsed` docstring now says it is unaffected by `value_to_float` (which converts `value` only — sidestepping the single-label letter/text asymmetry) and notes that aggregated/reduced multi-segment results do not carry it. I chose documenting over propagating `parsed` through the reducers: for value-reducers (mean/mode/etc.) there is no principled `parsed` for the combined result, so propagation only cleanly applies to the resultset-aggregation path and feels like a separate change if wanted.

Also added the suggested `docs/scanner_tools.qmd` coverage (Answer Generation + Answer Parsing sections), and renamed `StructuredT` → `StructuredAnswerT`.

kaifronsdal:
Ran a second, deeper review with four parallel agents (correctness/edge-cases, static typing, downstream integration, tests/docs), each required to reproduce findings with running code. Results, all addressed in 977848ea:

**Critical (integration):** with `parsed` holding an instance of a class pickle cannot resolve by reference — the *common* case of an answer model defined in a script's `__main__` or inside a scanner factory — the multiprocess scan queues' feeder threads raised `PicklingError` and silently dropped every result: the scan completed "successfully" with nothing recorded (multiprocess is the default execution mode). Fixed with `Result.__getstate__` dropping `parsed` during pickling, consistent with its documented in-memory-only semantics; results now always cross process boundaries (verified end-to-end through a spawn-context queue), and nothing on the main-process side ever read `parsed`.

**Major (correctness):** the previous `as_declared_type` fix eliminated the *serializer* round trip but still re-ran field *validators* on already-validated values — crashing `mode="before"` validators that parse wire formats and `Json[...]` fields, double-applying non-idempotent validators (so `parsed` disagreed with `value`), and running `model_post_init` twice. `structured_result` now validates the wire-format data exactly once, directly into the declared type (lifting the injected `explanation` key out first), so validators and `model_post_init` behave exactly as on `main`. Regression tests added for all these shapes.

**Minor (typing):** `AnswerStructured(tuple[Model, ...])` was statically admitted by the `Sequence` bound (main rejected it) but failed at runtime with an obscure `AttributeError` — `structured_answer_type` now raises a clear `ValueError`. Also dropped the bound on `ParsedT`: it hardcoded llm_scanner's answer vocabulary into the core `Result`, blocking custom scanners from e.g. `Result[dict[str, float]]`, and bought no safety since the overloads fix their type arguments themselves.

**Tests/docs:** `parsed` now asserted on `generate_answer` results (including the max-attempts failure path), multi-label failure rows added, the table test pins types (`bool` vs `1.0`), new `test_parsed_typing.py` guards the overload inference statically via `assert_type` (verified under both mypy and pyright), and docs now note that `parse_answer` raises rather than returning `parsed=None` for unparseable structured output, plus a `parsed` mention in `llm_scanner.qmd`.

Everything re-verified: strict mypy clean (458 files), full suite green (3,217 passed / 0 failed), `openapi.json` still byte-identical.

kaifronsdal:
Restored `ValidationError` semantics for the injected explanation field in 26090210.

The round-2 fix for single-pass validation replaced validation against the augmented type with hand-rolled `ValueError` checks for the injected `explanation`. Since models violate the answer schema often in practice, that path is hot, and the change cost callers two things: `except ValidationError` handlers no longer caught these failures, and recorded scan errors lost pydantic's loc/input detail.

`parse_answer_data` now validates the lifted explanation against a small static model (`_InjectedExplanation`), so missing, non-string, and non-object shapes all raise `ValidationError` exactly as on `main`, while the declared type's fields are still validated exactly once on wire-format data (the `mode="before"` validator / `model_post_init` guarantees are untouched — no user validators run on the explanation). One residual difference from `main`: a completion that is invalid in both the explanation and a declared field reports the explanation error first rather than one combined `ValidationError`.

Added a parametrized regression test covering all three shapes. Full suite green (3,227 passed / 0 failed), strict mypy clean (460 files), ruff clean.

PR #580, opened 2026-08-24
feat: add antigravity_cli() source

anthonyduong9:
fix: ATIF event working times not calculated from first event clock time

For `atif()` imports: `working_start` is now the offset from the first event rather than the monotonic-clock default, and `completed` is set equal to `timestamp` on synthesized `ModelEvent`s.

## This PR contains:
- [x] New features
- [ ] Docs
- [x] Bug fixes
- [ ] Code refactor

### What is the current behavior? (You can also link to an open issue here)

Users can't import Antigravity CLI transcripts.

### What is the new behavior?

Users can use `antigravity()` to import transcripts, specifying a conversation ID, time range, or limit.

### Does this PR introduce a breaking change? (What changes might users need to make in their application due to this PR?)

No.

### Other information:

In `step_to_messages()`, if `Step.type` doesn't match any values in our code, the function logs and returns a `ChatMessageSystem` if `Step.content` is truthy (and otherwise returns an empty list) because the alternative is for `Step.type` to be a `Literal` and if there are unknown values in future versions of the Antigravity CLI, instantiating `Step` with one of these values results in a validation error.

If `Step.type` is `"CHECKPOINT"`, `_convert_steps()` calls `checkpoint_index()` to get the number/index and appends a `ChatMessage` to the list used to set `Transcript.messages` and a `CompactionEvent` to the list used to set `Transcript.events`, only if the number/index is greater than 0. This is because `CHECKPOINT 0` means the start of a conversation, and only `CHECKPOINT 1`, `CHECKPOINT 2`, etc. mean compaction.

`_create_subagent_span_events()` logs and returns an empty list if it can't find a conversation for a given subagent, meaning there's no span for this subagent in the transcript. We can still tell that Antigravity CLI spawned a subagent because `Transcript.messages` contains a `ChatMessageAssistant` where `tool_calls` contains a `ToolCall` where `function` is `"invoke_subagent"` and a `ChatMessageTool` where `content` is the result of the `invoke_subagent` tool call, and `Transcript.metadata` has a `"subagent_conversation_ids"` key.

When `Step.type` is `"USER_INPUT"`, only the content between `<USER_REQUEST>` and `</USER_REQUEST>` is used to set `ChatMessageUser.content`. Content between `<ADDITIONAL_METADATA>` and `</ADDITIONAL_METADATA>`, which contains the current local time, is not used anywhere, and the content between `<USER_SETTINGS_CHANGE>` and `</USER_SETTINGS_CHANGE>` is used as one way of getting the model name.

I ran
```
python examples/sources/antigravity/populate_db.py
```
.

Then I ran
```
scout view -T examples/sources/antigravity/transcripts
```
and looked at the transcripts and messages.

I effectively ran those but without the `path` argument in `examples/sources/antigravity/populate_db.py`, to import my own conversations.

I ran `cd docs` and then `quarto preview`.


dragonstyle (changes requested):
Thanks — the layout follows atif/claude_code, the protobuf walker held up under fuzzing (5k mutations, no escapes), and the suite passes locally against inspect_ai main. Several things need to change before this lands; each item says what I observed and what would resolve it.

**Blockers**
- [ ] Tool results with a typed step are not recognized. Only `type == "GENERIC"` is treated as a tool result (and the sub-agent spawn check keys on it too). Three other parsers of `transcript_full.jsonl` — specstory's format doc, hapi's `agyTranscriptTypes.ts`, and agent-sessions (verified against agy 1.1.14 and 1.1.22 sessions) — document typed results: `RUN_COMMAND`, `VIEW_FILE`, `CODE_ACTION`, `GREP_SEARCH`, `LIST_DIRECTORY`, `SEARCH_WEB`, `READ_URL_CONTENT`, `GENERATE_IMAGE`, `INVOKE_SUBAGENT`, `ASK_QUESTION`, with `GENERIC` only for tools lacking a dedicated type. Retyping the fixtures that way, every tool result becomes a `ChatMessageSystem`, no `ChatMessageTool` is produced, the pairer never pops (so later results would pair with stale calls), and the `INVOKE_SUBAGENT` result triggers no inlining — the child is yielded standalone with one "Unknown step type" warning per result. Please check your corpus for these types; either way, treat any step with `source == "MODEL"` and `type != "PLANNER_RESPONSE"` as a tool result (and run the spawn-marker check on the same set), and add a fixture with typed results.
- [ ] Sub-agent events aren't nested under their agent span. `_create_subagent_span_events` never sets `span_id` on the child's events or `parent_id` on nested spans, so `event_tree()` shows the "Test researcher" span with 0 children and the child's model calls at the root. Copy the re-parenting block from `_atif/transcripts.py` (`span_id = agent_span_id` on non-span events, `parent_id` on `SpanBeginEvent`s with none). Once you do, a spawn cycle (a child whose spawn result names an ancestor) produces a span whose `parent_id` is its own `id` and `timeline_build` recurses until `RecursionError` — the depth-5 cap doesn't catch it — so also thread the set of conversation ids currently being inlined through `_convert_steps` and skip a child already in it. Add an assertion that `event_tree(transcript.events)` places the two child ModelEvents under the span; `test_subagent_inlined_as_agent_span` only checks list order today.
- [ ] One undecodable file aborts the whole import. `_read_steps` catches `OSError` but `open(..., encoding="utf-8")` raises `UnicodeDecodeError` on a transcript with non-UTF8 bytes; it fires in the child-id pre-scan, so `db.insert(antigravity())` stores nothing. Catch `(OSError, ValueError)` there (and `RecursionError` alongside `JSONDecodeError` in `read_jsonl_steps` — a deeply nested line aborts the same way).
- [ ] Time-windowed imports drop sub-agents. `records_by_id` is built from the mtime-filtered records, so with `from_time` between a child's and its parent's mtime (the usual case — the parent keeps writing after the child finishes) the parent is yielded with no span and a "not found on disk" warning, while `metadata.subagent_conversation_ids` still lists the child. Build the lookup map from an unfiltered `discover_conversations(path)` and apply `from_time`/`to_time` only to what gets yielded.
- [ ] The description doesn't say what changes for ATIF. The body's `fix: ATIF event working times…` line becomes a release-notes entry, so add a sentence on the actual change for `atif()` imports: `working_start` is now the offset from the first event rather than the monotonic-clock default, and `completed` is set equal to `timestamp` on synthesized ModelEvents. (Verified: nothing else in the ATIF output changes — `total_time`, tokens, dates, and event structure are identical on the fixtures.)

**Take-or-leave**
- [ ] A `CHECKPOINT` step whose content has no `{{ CHECKPOINT N }}` marker is dropped without a `CompactionEvent`, message, or warning; unknown step *types* warn, so an unmarked checkpoint should at least warn too.
- [ ] An unreadable conversation directory (or `brain/` itself) raises `PermissionError` out of `discover_conversations` and aborts the import; wrapping the per-directory work in `try/except OSError` with a warning would match how unreadable files are handled.
- [ ] `read_generation_metadata`'s docstring cites "antigravity-usage (MIT)" as corroboration for the field paths, but that project tracks quota via Google's API and never reads `gen_metadata`; kenn-io/agentsview (`antigravity_proto.go`) and ccusage's antigravity adapter are the ones that decode the same blobs. agentsview also reads the daemon's own `conversations/<id>.trajectory.json` sidecar (`generatorMetadata[].stepIndices` + `chatModel.usage`, plaintext), which would replace the ordinal mapping — worth a tracked issue rather than a change here.
- [ ] `db_importing.qmd` says token usage is best-effort; one line in the `antigravity()` docstring that `model`/`total_tokens` may be `None` when the conversation store can't be read would keep the generated reference page self-contained.
- [ ] `_claude_code` builds its events with the same default-factory `working_start` this PR fixes for ATIF; if you'd rather not widen the PR, a tracked issue pointing at `apply_working_start` would do.

anthonyduong9:
Thanks.

I've checked my own transcripts (for Antigravity CLI 1.1.14 - 1.1.19) and didn't see values except `"GENERIC"`. Regardless, I've replaced all checks that `Step.type` is `"GENERIC"` with the check that `Step.source` is `"MODEL"` and `Step.type` isn't `"PLANNER_RESPONSE"` by extracting `is_tool_result_step()` and `_spawn_result_content()`. I've also added
`fixtures/root/brain/eeeeeeee-0000-0000-0000-000000000005/.system_generated/logs/transcript_full.jsonl`, which contains objects for some of the other values for `Step.type`.

I've copied the block from `_create_subagent_span_events()` for ATIF to `_create_subagent_span_events()` for Antigravity. I've added a `frozenset[str]` param to `_convert_steps()` and `_create_subagent_span_events()`. `_create_subagent_span_events()` will log and return early if `child_id` is in that set. I've also updated `test_subagent_inlined_as_agent_span()` to call `event_tree()` and assert that for the `EventTreeSpan` in its output, `EventTreeSpan.children` has the two `ModelEvent` objects from the subagent in `Transcript.events` (in the same order).

I've:
- changed `_read_steps()` to also handle `ValueError` and `read_jsonl_steps()` to also handle `RecursionError`
- changed `antigravity()` to call `discover_conversations()` without any `from_time` and `to_time` args and moved the `from_time` and `to_time` checks out of the latter function and into the former
- updated the PR description with the sentence you wrote for `atif()`
- changed `_convert_steps()` to log when `Step.type` is `"CHECKPOINT"` but `Step.content` doesn't match `_CHECKPOINT_RE`
- updated `discover_conversations()` to handle `OSError` from `Path.is_dir()`, `Path.iterdir()`, and `_find_transcript()`, and also fixed a bug in the same class, in `read_title()`, by moving a block with `Path.is_file()` under `try`
- replaced `antigravity-usage` with `kenn-io/agentsview` and `ccusage` in `read_generation_metadata()` docstring
- updated the `antigravity()` docstring to say `Transcript.total_tokens` can be `None`
- created #629

I didn't create an issue to use the sidecar file, because in my transcripts I found that there was always an object where `"type"` was `"PLANNER_RESPONSE"`, even if the model generation failed, so I can't see the matching between objects and rows of generation metadata not working. Plus, the sidecar file comes from `agy-reader`, not Antigravity CLI.

dragonstyle (changes requested):
The earlier fixes are covered, and the Antigravity/ATIF tests pass. Before merge:

- [ ] Correct the model context at compaction boundaries. With the existing compaction fixture, `span_messages(..., compaction="all")` repeats the pre-checkpoint messages, while `compaction="last"` still includes them. Track the current model context separately from complete transcript history and test both extraction modes.
- [ ] Prevent interrupted calls from consuming later turns' tool results. A `view_file` call without a result followed by a new `run_command` call currently attaches the command result to `view_file`. Handle abandoned pending calls and subagent roles at the appropriate boundaries, with a regression test.
- [ ] Handle undecodable annotation files in `read_title()`. A non-UTF-8 `.pbtxt` currently aborts the whole import during discovery; omit the title and continue importing the valid conversations, with a regression test.

anthonyduong9:
I've also
- renamed `antigravity()` to `antigravity_cli`
- changed `"antigravity"` to `"antigravity-cli"` for `Transcript.agent`
- changed `"antigravity"` to `"antigravity_cli"` for `Transcript.source_type` and `CompactionEvent.source`

since the feature is for transcripts from Antigravity CLI, not Antigravity SDK, Antigravity IDE, etc.

Given that `inspect-swe` has `antigravity()` for the Antigravity SDK, and it looks like https://github.com/meridianlabs-ai/inspect_swe/pull/153 will add `antigravity_cli()` for Antigravity CLI, I thought it made sense to match the latter in `inspect-scout`.

If you prefer the name `antigravity`, I can revert f0c0d91 and change the PR title back to what it was.

dragonstyle:
Thanks for the rename; matching `antigravity_cli()` in inspect_swe makes sense.

Before I re-review, can you confirm which of the three "before merge" items from my last review afa22a9 addresses? Specifically the compaction-boundary model context (both `compaction="all"` and `"last"`), abandoned pending tool calls consuming later results, and undecodable `.pbtxt` titles. If any are still open, say so and I'll wait for the push rather than reviewing a partial state.

anthonyduong9:
afa22a9 addresses all the checkboxes in https://github.com/meridianlabs-ai/inspect_scout/pull/580#pullrequestreview-5159107928.

It added the `context` variable in `_convert_steps()`. Until `CHECKPOINT 1`, `context` is the same as `messages`. From `CHECKPOINT 1` onwards, it's the `ChatMessageSystem`, then every `ChatMessage` until `CHECKPOINT 2`, and so on. It also passes `context` to `to_model_event()`, to set `ModelEvent.input`. It added tests `test_compaction_model_context_all()` and `test_compaction_model_context_last()`.

It changed `_convert_steps()` so when `Step.type` is `"PLANNER_RESPONSE"`, the function effectively calls the `clear()` method on `ToolCallPairer._pending`, then `extend()`. This means that if `ToolCallPairer._pending` has a `ToolCall` with no tool result in the transcript, `step_to_messages()` doesn't use that `ToolCall` to create a `ChatMessageTool`. It made analogous changes for `pending_roles`, and added tests `test_abandoned_call_does_not_claim_later_result()` and `test_abandoned_spawn_does_not_claim_later_role()`.

It changed `read_title()` so when `Path.read_text()` raises `ValueError`, `read_title()` handles it by returning `None`. This means `discover_conversations()` sets `ConversationRecord.title` to `None` and that there's no `"title"` key in `Transcript.metadata`. It also added the test `test_undecodable_annotation_omits_title()`.

PR #586, opened 2026-08-28
fix: report per-item model usage for custom loaders

### LINKED ISSUE #577: Per-result model_usage is a cumulative snapshot for loader-multiplexed scanners, inflating usage ~N/2x
*Filed by Claude Code on behalf of @alexandraabbas, who reviewed this issue before filing.*

## Summary

When a scanner has a custom loader that yields N items per transcript, each `ResultReport` records the scanner's *accumulated* model usage at the time the result landed, not the usage of that item's own calls. Result k therefore reports the sum of calls 1..k+1, and anything that sums per-result usage — the scan display's per-scanner token column, `scan_total_tokens` / `scan_model_usage` in results dataframes — overcounts by roughly N/2×.

## Observed

inspect-scout 0.4.45. A classification scanner (llm_scanner-based, one small call per item) over a loader yielding 64 items from one transcript. Each call is genuinely ~1.2k tokens, but per-result `scan_model_usage` grows in exact multiples of the first call:

```
result 0:  input 21    cache_write 1,134   output 86      (1x — correct)
result 1:  input 42    cache_write 2,262   output 167     (2x)
result 2:  input 63    cache_write 3,406   output 250     (3x)
result 31: input 672   (= 32 x 21)         output 2,747   (32x)
result 63: input 1,344 (= 64 x 21)         output 5,494   (64x)
```

The scan display attributed **2,628,422 tokens** to this scanner (the sum of 64 running totals); true spend was **~81k** (the last cumulative value). The real per-call deltas are uniform and correct.

## Locus

`_scan.py`: `init_model_usage(initial_usage={})` runs once per scan job, then inside `async for loader_result in loader(...)` each report records the running counter:

```python
results.append(
    ResultReport(
        ...,
        model_usage=model_usage(),   # accumulated since job start, not per item
    )
)
```

With a single result per transcript (no loader, or one item) the snapshot equals the delta, which is presumably why this hasn't surfaced.

## Suggested fix

Snapshot the counter at the top of each loader iteration and record the difference (or re-init per iteration if no cross-item usage semantics are intended). Downstream consumers already sum per-result usage, so per-item deltas appear to be the intended meaning.

## Repro sketch (no API key)

```python
# scanner with a loader yielding N tiny items; model=mockllm/model with
# scripted outputs; run scout scan; read scan_model_usage per result ->
# values grow linearly instead of staying flat.
```

Happy to turn the sketch into a full runnable repro or a PR if useful.

YusefSyed:
## Summary

- reset Inspect model-usage tracking after each loader result is captured
- attribute loader, scanner, and validation calls exactly once to the yielded item
- add a public-API regression proving per-result and scanner-summary token totals

Closes #577.

## Evidence

With the new regression copied onto clean upstream `HEAD`, the three per-result token
counts are `[4, 8, 12]` and the scanner summary is 24. With this patch, the result
counts are `[4, 4, 4]` and the scanner summary is 12.

## Verification

- Ruff check and format: passed
- Mypy across `src`, `examples`, and `tests`: passed (458 source files)
- Focused regression: passed
- `tests/scan/test_scan_e2e.py`: 8 passed
- Full pytest: 3,197 passed, 107 skipped; two unrelated Rich terminal-capture tests
  failed on macOS and fail identically on a detached clean upstream `HEAD`

### Agent review

- Pass 1: GPT-5.6 Sol via Codex architect role, fresh context. Found four issues:
  loader-side calls were initially excluded, a manual type annotation broke mypy,
  initializing that variable changed an unrelated error path, and fieldwise usage
  subtraction plus duplicate tests were brittle. All four were fixed.
- Pass 2: GPT-5.6 Terra via Codex scout role, fresh context. No blocking findings;
  confirmed loader/scanner/validation/error-path accounting, summary aggregation,
  typing, two-file scope, and regression coverage.
- Dismissed as out of scope: the pre-existing summary helper omits `total_cost` when
  aggregating `ModelUsage`; this patch changes per-result isolation and token
  overcounting, not that separate aggregation behavior.

### DIFF
--- a/src/inspect_scout/_scan.py
+++ b/src/inspect_scout/_scan.py
@@ -1054,6 +1054,7 @@ async def _scan_one(
         try:
             type_and_ids = get_input_type_and_ids(loader_result)
             if type_and_ids is None:
+                init_model_usage(initial_usage={})
                 continue
 
             # do scan
@@ -1104,6 +1105,7 @@ async def _scan_one(
                     model_usage=model_usage(),
                 )
             )
+        init_model_usage(initial_usage={})
 
     return results
 
--- a/tests/scan/test_scan_e2e.py
+++ b/tests/scan/test_scan_e2e.py
@@ -1,12 +1,22 @@
 import json
 import shutil
 import tempfile
+from collections.abc import AsyncIterator
 from pathlib import Path
 
 import pytest
-from inspect_ai.model import ModelOutput
+from inspect_ai.model import ModelOutput, get_model
 from inspect_ai.model._chat_message import ChatMessageUser
-from inspect_scout import Result, Scanner, llm_scanner, scan, scanner, transcripts_db
+from inspect_scout import (
+    Loader,
+    Result,
+    Scanner,
+    llm_scanner,
+    loader,
+    scan,
+    scanner,
+    transcripts_db,
+)
 from inspect_scout._scanresults import scan_results_df
 from inspect_scout._transcript.factory import transcripts_from
 from inspect_scout._transcript.types import Transcript
@@ -47,6 +57,32 @@ def simple_scanner_factory() -> Scanner[Transcript]:
 def llm_scanner_factory() -> Scanner[Transcript]:
     """LLM scanner that uses mockllm for testing."""
     return llm_scanner(question="Is this conversation helpful?", answer="boolean")
+
+
+@loader(name="multi_item_model_usage_loader", messages=["user"])
+def multi_item_model_usage_loader_factory() -> Loader[ChatMessageUser]:
+    """Create a loader that makes one model call before yielding each message."""
+
+    async def load(transcript: Transcript) -> AsyncIterator[ChatMessageUser]:
+        for message in transcript.messages:
+            if isinstance(message, ChatMessageUser):
+                await get_model().generate([message])
+                yield message
+
+    return load
+
+
+@scanner(
+    name="multi_item_model_usage_scanner",
+    loader=multi_item_model_usage_loader_factory(),
+)
+def multi_item_model_usage_scanner_factory() -> Scanner[ChatMessageUser]:
+    """Create a scanner whose model usage occurs only inside its loader."""
+
+    async def scan_message(message: ChatMessageUser) -> Result:
+        return Result(value=bool(message.text))
+
+    return scan_message
 
 
 @scanner(name="llm_dynamic_question_scanner", messages="all")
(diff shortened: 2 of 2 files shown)


epatey:
Intake notes ahead of a review. These orient whoever reviews deeply; they are not the review.

**Recommendation.** Worth a deep review. Thanks for the thorough writeup and the regression. Two things to look at first: (1) the loader-skip branch now discards that iteration's usage rather than carrying it forward, so any model calls a loader makes for a filtered item vanish from accounting, please confirm that's intended; (2) CI hasn't run yet on this fork PR, so we'll want the workflows kicked before merge.

**What it does.** `_scan_one` calls `init_model_usage(initial_usage={})` once before the loader loop, then records `model_usage=model_usage()` on every `ResultReport` (`_scan.py:1104`). Because `model_usage()` is a running counter and the loader makes its call before each `yield`, result k was capturing the cumulative total of calls 1..k. The patch adds `init_model_usage(initial_usage={})` in two spots: at the end of each loop iteration (after the append) and in the `type_and_ids is None` skip branch. Since the loader is only resumed at the top of the next iteration, resetting at the tail of the body means each result captures exactly its own item's delta. Premises verified against main: loader/scanner/validation calls are attributed once to the yielded item, and the scan summary sums per-result usage (`_recorder/summary.py:135-141`), so per-result deltas are the intended meaning. Evidence in the diff (`[4,8,12] -> [4,4,4]`, summary `24 -> 12`) matches the mechanism; no meaningful body/diff drift.

**Why.** Closes #577 (filed by a maintainer with a precise diagnosis and observed N-times inflation). Provenance is solid: linked issue plus a demonstrated regression in the diff.

**Beyond the diff.** Safety fact (proven): the reset counter is read in exactly one place, the per-result snapshot at `_scan.py:1104`; the scan-level summary sums those per-result values, never the global counter after the loop, so resetting per iteration cannot break a grand-total consumer. Verified that `init_model_usage(initial_usage={})` installs a fresh dict, so previously-captured per-result dicts are not mutated. Real risk (minor): the skip-branch reset discards that iteration's usage entirely; previously it leaked forward into the next result, now it's counted nowhere. Contract risk: attribution assumes loaders call the model before `yield`; a loader that calls after yielding would attribute to the next item.

**Open questions.**
- Discard-on-skip: intended, or should that usage carry to the next appended result?
- CI has not run and the branch is BEHIND main; kick the fork PR's workflows before merge (`maintainerCanModify` is true).

epatey:
@review

epatey (approved):
Approving. The per-result usage delta fix is correct, and the scan summary sums per-result values so grand totals are unaffected. Thanks for the regression test.

YusefSyed:
Updated the branch to current `main` (`b1c12dc`) with merge commit `ba312a0`.

The six upstream commits do not touch `_scan.py` or `tests/scan/test_scan_e2e.py`; the PR diff remains the same focused two-file fix and regression.

Local verification on the updated branch:

- `tests/scan/test_scan_e2e.py`: 8 passed
- `make check`: Ruff, formatting, mypy (461 source files), and suppression-ledger check passed
- full suite: 3,293 passed, 107 skipped, with only the two existing macOS Rich terminal-capture failures; both fail identically on a clean worktree at current `main`

The `Suppressions comment` check passed. The five `pull_request` workflows are currently marked `action_required` for this fork PR, so they still need maintainer approval to run when convenient.

PR #585, opened 2026-08-27
docs: expand AGENTS.md with async guardrails, submodule guard, and PR conventions

epatey:
Expands AGENTS.md using inspect_ai's AGENTS.md as the model, keeping all existing content. Every command, path, and claim was verified against the repo.

**Additions**
- Submodule gitlink guard: never bump the ts-mono pointer incidentally; reset recipe + pointer to the `land-ts-mono` skill
- Async and Concurrency section: `tg_collect` over `asyncio.gather` (existing `gather` sites tracked in #584), FastAPI don't-block-the-loop rule (distilled from `design/fastapi-async.md`), job vs infrastructure exception boundary (from `design/exception_handling.md`), `s3://`-capable path handling via `filesystem()`
- Testing: `@pytest.mark.asyncio` required (strict mode); `slow`/`api`/`flaky` markers and their `--run*` flags; xdist on by default, `-n 0` for debugging
- Pull Requests: outcome-not-mechanism title rule; watch CI after opening; Agent review disclosure section
- Style: typed returns (NamedTuple over bare tuple), call-site comment and comment-length guidance (adopted from inspect_ai)
- Setup pointer to CONTRIBUTING.md; `fastapi-async.md` added to the design-doc list; Git LFS pointer-stub pitfall

**Corrections to pre-existing content** (found while verifying)
- `pnpm watch` is defined per app, not at the ts-mono root
- `pnpm build` does not regenerate types; the Type Sharing steps now use `pnpm types:generate` in `apps/scout` with real artifact paths
- LFS note: LFS tracks `src/inspect_scout/_view/dist/**` in this repo; the submodule has no LFS-tracked files

Note: CONTRIBUTING.md and the submodule guide repeat the "`pnpm build` regenerates types" defect — left untouched here (AGENTS.md-only PR); can follow up if wanted.

### Agent review
- Reviewer: fresh-context Claude agent (same session model), 1 pass, fact-checking every command/path/claim against the repo
- Findings: 5 — 5 fixed (unmarked async tests fail rather than skip on pytest ≥8.4; `pnpm watch` not a root script; `pnpm build` doesn't regenerate types; `types:generate` not available at ts-mono root; LFS misattributed to the submodule)

PR #587, opened 2026-08-29
refactor: collect concurrent scan reads with tg_collect

### LINKED ISSUE #584: Replace remaining asyncio.gather calls with tg_collect
AGENTS.md now specifies `tg_collect()` from `inspect_ai._util._async` instead of `asyncio.gather()` for concurrent async tasks (structured concurrency / proper cancellation). Three call sites still use `asyncio.gather`:

- `src/inspect_scout/_scanjobs/refresh.py:151`
- `src/inspect_scout/_scanjobs/refresh.py:218`
- `src/inspect_scout/_recorder/file.py:734`

Convert them to `tg_collect`. Note `tg_collect` takes a list of no-arg coroutine functions (use `functools.partial`) rather than awaitables, and raises on first error via the task group—check each site's error-handling expectations when converting.

antnewman:
Closes #584.

Converts the three remaining `asyncio.gather` call sites to `tg_collect`, per the async guardrails added in #585.

## The wrinkle

`tg_collect` has no `return_exceptions` equivalent — it always raises — and two of the three sites depend on per-scan failure isolation. A direct swap would change behaviour in ways CI would not catch:

- **`refresh_index`**: a scan directory with no `_scan.json` yet always raises `FileNotFoundError`, so a partially-created scan would abort the whole refresh instead of being skipped, and `store.upsert`/`store.delete` would never run.
- **`FileRecorder.list`**: a scan deleted between listing and read is routine on a remote store; `scout scan list` would raise instead of returning the other scans.
- **Worse, and not obvious**: an anyio task group drops a child `CancelledError` without recording it, then cancels its own scope — so `tg_collect` returns a **short list and raises nothing**. In `async_listing_to_scans` a short listing is not merely incomplete: `compute_delta` treats an absent `scan_id` as deleted, so `refresh_index` would delete indexed scans. Verified directly:

```
tg_collect([ok, raises_CancelledError, ok])  ->  returns ['ok', 'ok'], raises nothing
```

So each read captures its own failure and returns it as a value. Results stay aligned with the scans they came from, and the existing skip/propagate handling is untouched. The capture covers `Exception` and cancellation but deliberately not `KeyboardInterrupt`/`SystemExit`, which are left to unwind as before.

**If you would rather these sites were genuinely fail-fast**, that is a smaller diff than this one and I am happy to make it — the current behaviour is documented in comments at both sites, so I preserved it rather than assume.

## Behaviour differences I did not eliminate

Stating these rather than claiming "no change":

- `async_listing_to_scans` no longer short-circuits. It previously raised as soon as one listing failed, leaving siblings running detached; it now waits for all of them, so a failure surfaces after the remaining reads finish and the exception raised is the first in listing order rather than the first chronologically. Only `BaseException` can escape that path (`_listing_from_scan_dir` handles everything else internally), so this is rare — but it is a real difference, and it removes orphan tasks that previously outlived the `AsyncFilesystem` they were using.
- Structured concurrency costs more scheduling overhead than `gather` on the local-filesystem path, where `AsyncFilesystem.info` has no await point. I measured roughly 4x on a synthetic 2000-item fan-out with an empty body; real listings are dominated by I/O, so the practical effect is much smaller, but it is not zero and `async_listing_to_scans` runs on every `SqliteScanJobsView.connect()`.

I also left `asyncio.Semaphore` alone. Switching it to `anyio.Semaphore` would match `_transcript/messages.py`, but it is out of scope here, measurably slower on this path, and there is no trio backend in the repo for it to protect against.

## Tests

Three new tests, all **characterisation** tests — they pass before and after this change. Their value is that they fail against a direct swap, which I verified by making one:

| Test | Catches |
| --- | --- |
| `test_async_listing_to_scans_propagates_cancelled_metadata_error` | the silent-truncation path above |
| `test_list_skips_scan_that_vanished_after_listing` | a missing scan aborting the listing |
| `test_list_returns_every_readable_scan_when_one_is_missing` | one failure cancelling its siblings, and result/scan alignment |

`FileRecorder.list` previously had no error-path coverage at all. The first two of the three existing `refresh_index` tests already pinned the equivalent behaviour there and stay green.

## Verification

`ruff check`, `ruff format --check` and `mypy` clean on the changed files; `pytest tests/scanjobs tests/recorder` gives 100 passed. Three failures in that run (`test_duckdb_security` ×2, `test_list_preserves_file_uri_locations`) reproduce identically on a clean checkout of `main` — they are pre-existing and Windows-specific.

Full disclosure on the environment: `pip install -e ".[dev]"` fails on my machine because `litellm` needs a Rust toolchain, so I installed runtime dependencies plus pytest/ruff/mypy and the two stub packages rather than the complete dev extras. I ran the two affected test directories, not the entire suite.

### Agent review
- Reviewer: Claude Opus 5, two independent passes, each in a fresh context with no knowledge of the other — one on semantics preservation, one on repo conventions and test quality
- Findings: 12 — 8 fixed, 4 disclosed above rather than fixed
  - Fixed: hand-wrapped code that `ruff format` would rewrite (would have failed `py-ruff`); three `# noqa: B036` suppressing a rule ruff does not implement; `_status_or_error` duplicated verbatim in both modules, now defined once and imported; a call-site comment restating what the helper does; over-long docstrings that documented their callers; catching `BaseException` where it changed `KeyboardInterrupt` handling; an `asyncio.Semaphore` change that was out of scope and slower; a third test that passed under a direct swap and so pinned nothing, replaced with the sibling-isolation and alignment test above
  - Disclosed, not fixed: the loss of short-circuiting, the scheduling cost, exception-selection order at that site, and that an empty input now yields one extra checkpoint

### DIFF
--- a/src/inspect_scout/_recorder/file.py
+++ b/src/inspect_scout/_recorder/file.py
@@ -1,4 +1,3 @@
-import asyncio
 import functools
 import io
 import json
@@ -18,6 +17,7 @@ import pyarrow.compute as pc
 import pyarrow.dataset as ds
 import pyarrow.fs as pafs
 import pyarrow.parquet as pq
+from inspect_ai._util._async import tg_collect
 from inspect_ai._util.asyncfiles import AsyncFilesystem
 from inspect_ai._util.file import file, filesystem
 from inspect_ai._util.json import to_json_str_safe
@@ -729,10 +729,9 @@ class FileRecorder(ScanRecorder):
                 scan_dirs.append(entry.as_posix())
 
         # Fetch in parallel; skip scans that no longer exist and propagate
-        # any other error. return_exceptions=True keeps one failure from
-        # cancelling its siblings.
-        results = await asyncio.gather(
-            *(FileRecorder.status(d) for d in scan_dirs), return_exceptions=True
+        # any other error.
+        results = await tg_collect(
+            [functools.partial(_status_or_error, d) for d in scan_dirs]
         )
         scans: list[Status] = []
         for r in results:
@@ -757,6 +756,20 @@ class FileRecorder(ScanRecorder):
 # cleared when a run starts (`init`/`resume`/`attach`) and when a sync
 # completes the scan.
 _prior_parquet_cache: dict[str, dict[str, str | None]] = {}
+
+
+async def _status_or_error(scan_dir: str) -> Status | BaseException:
+    """Read one scan's status, returning any failure as a value.
+
+    Concurrent reads run in a task group, which cancels the remaining reads as
+    soon as one raises. Cancellation is captured too — the caller re-raises it,
+    and anyio re-delivers a genuine cancellation on scope exit — but
+    KeyboardInterrupt and SystemExit are left to unwind as they always did.
+    """
+    try:
+        return await FileRecorder.status(scan_dir)
+    except (Exception, anyio.get_cancelled_exc_class()) as ex:
+        return ex
 
 
 def _clear_prior_parquet_cache(scan_location: str) -> None:
--- a/src/inspect_scout/_recorder/file.py
+++ b/src/inspect_scout/_recorder/file.py
@@ -42,6 +42,7 @@ from .._recorder.buffer import (
 from .._scanner.result import Error, ResultReport
 from .._scanspec import ScanSpec, ScanTranscripts
 from .._transcript.types import TranscriptInfo
+from .._util._async import as_value
 from .._util.duckdb import create_parquet_view, restrict_external_access
 from .recorder import (
     ScanRecorder,
@@ -731,7 +732,10 @@ class FileRecorder(ScanRecorder):
         # Fetch in parallel; skip scans that no longer exist and propagate
         # any other error.
         results = await tg_collect(
-            [functools.partial(_status_or_error, d) for d in scan_dirs]
+            [
+                functools.partial(as_value, functools.partial(FileRecorder.status, d))
+                for d in scan_dirs
+            ]
         )
         scans: list[Status] = []
         for r in results:
@@ -756,20 +760,6 @@ class FileRecorder(ScanRecorder):
 # cleared when a run starts (`init`/`resume`/`attach`) and when a sync
 # completes the scan.
 _prior_parquet_cache: dict[str, dict[str, str | None]] = {}
-
-
-async def _status_or_error(scan_dir: str) -> Status | BaseException:
-    """Read one scan's status, returning any failure as a value.
-
-    Concurrent reads run in a task group, which cancels the remaining reads as
-    soon as one raises. Cancellation is captured too — the caller re-raises it,
-    and anyio re-delivers a genuine cancellation on scope exit — but
-    KeyboardInterrupt and SystemExit are left to unwind as they always did.
-    """
-    try:
-        return await FileRecorder.status(scan_dir)
-    except (Exception, anyio.get_cancelled_exc_class()) as ex:
-        return ex
 
 
 def _clear_prior_parquet_cache(scan_location: str) -> None:
--- a/src/inspect_scout/_scanjobs/refresh.py
+++ b/src/inspect_scout/_scanjobs/refresh.py
@@ -13,7 +13,6 @@ from functools import partial
 from logging import getLogger
 from typing import Protocol
 
-import anyio
 from inspect_ai._util._async import tg_collect
 from inspect_ai._util.asyncfiles import AsyncFilesystem
 from inspect_ai._util.file import FileInfo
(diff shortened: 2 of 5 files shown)


epatey:
Intake notes ahead of a review. These orient whoever reviews deeply; they are not the review.

**Recommendation.** Deep review. This is a strong, unusually well-documented PR. Two things to look at first: (1) the `refresh_index` status consumer, which checks `isinstance(status, Exception)` while `_status_or_error` can hand it a captured cancellation (a `BaseException`, not an `Exception`); (2) a quick sanity-check of the disclosed ~4x scheduling cost on the `SqliteScanJobsView.connect()` hot path. Everything else here is careful and honest.

**What it does.** Converts the three remaining `asyncio.gather(..., return_exceptions=True)` sites to `tg_collect` per #584. Since `tg_collect` has no `return_exceptions`, the PR wraps each read in a helper (`_status_or_error`, `_listing_or_error`) that catches `(Exception, cancelled_exc_class)` and returns the failure as a value; consumers then skip or re-raise. Premises verified against the code: `tg_collect` takes no-arg async callables and returns results in input order (so alignment holds), and the hazard is real (verified by repro: a child raising a bare `CancelledError` inside a task group is dropped, and `compute_delta` would then delete indexed scans on a silently shortened listing). This is a correctness fix, not cosmetics.

**Why.** Provenance is solid: issue #584 is maintainer-authored, names all three sites, and flags the error-handling swap as the thing to get right. The PR matches that scope and adds coverage where `FileRecorder.list` had none.

**Beyond the diff.** Safety fact (proven): alignment and no-silent-truncation hold because the wrappers let nothing escape `run_task`, so `tg_collect` never self-cancels siblings and captured failures come back as ordered values. Real risk (low-likelihood): the `refresh_index` consumer at `refresh.py:227` gates on `isinstance(status, Exception)`, but `_status_or_error` can return a captured cancellation (`BaseException`, not `Exception`), which would fall through and be treated as a `Status`. Asymmetric with `FileRecorder.list`, which handles `BaseException`. Looks unreachable today, but the asymmetry is worth a deliberate decision; the `isinstance(status, ExceptionGroup)` unwrap just below is now dead. Disclosed behavior changes (all real): failure now surfaces the first exception in listing order rather than chronologically, orphan tasks are removed, and the ~4x scheduling overhead on the empty-body local fan-out.

**Open questions.**
- Should the `refresh_index` consumer handle `BaseException` for parity with `list`, or is a captured cancellation reaching that loop genuinely impossible (a one-line assert/comment would settle it)?
- CI has not run and the branch is BEHIND main; kick the workflows before deep review so the mypy/ruff/pytest claims are confirmed in CI, not only the author's partial local env.

epatey:
@review

epatey (approved):
Approving. A careful, semantics-preserving conversion: the wrappers keep tg_collect from silently dropping failures, and result alignment holds. Thanks.

antnewman:
Thanks for the intake notes — both points answered below, and I've taken the inline suggestion.

**1. The `refresh_index` status consumer and captured cancellation.** This one is handled, and deliberately so. The consumer checks in this order:

```python
if isinstance(status, ExceptionGroup):
    status = inner_exception(status)
if isinstance(status, Exception):      # skip, with a warning unless FileNotFoundError
    ...
    continue
if isinstance(status, BaseException):  # a captured cancellation lands here
    raise status
```

A captured `CancelledError` is not an `Exception`, so it falls past the skip branch to `raise status` — which is exactly what `gather(return_exceptions=True)` produced before, since it also stored a child-raised cancellation as a result value. `tests/scanjobs/test_refresh.py::test_refresh_index_propagates_cancelled_status_error` pins precisely this and passes; `FileRecorder.list` has the same two-tier shape (`FileNotFoundError` → skip, any other `BaseException` → raise). It was worth flagging though — the ordering is load-bearing and easy to get wrong.

**2. The disclosed ~4x.** Fair challenge, and my number was the pessimistic end. That figure came from a synthetic fan-out with an empty task body, which measures pure scheduling overhead and nothing else. Re-measured with a real `os.stat` per item — the shape `AsyncFilesystem.info` takes on a local path — over 500 scan dirs at the production concurrency of 50, best of 5:

```
old (gather)     :  28.18 ms
new (tg_collect) :  41.48 ms
delta            : +13.30 ms  (1.47x, +26.6 us per scan dir)
```

So ~1.5x and about 27µs per scan dir on the local path, not 4x. On a remote listing, where each `info` is a network round trip, it disappears into the noise. I've left the original figure in the description as the worst case but this is the more representative one — happy to fold it into the body if you'd prefer that.

**3. Took the inline suggestion** (`24552722`). The two local wrappers are now one `as_value` helper in `_util/_async.py`, composed with `partial` at all three call sites. That also removes the `_scanjobs` → `_recorder.file` private import the reviewer objected to — `refresh.py` is back to importing only the public `FileRecorder` — and the cancellation-capture rationale now lives in exactly one docstring instead of two. mypy clean, and the affected suites still pass (100 passed; the three failures in that run reproduce on a clean `main` and are Windows-specific, as noted in the description).

epatey (approved):
Re-approving after the branch update. The fix is unchanged and the suppression ledger is untouched. Good to merge.

PR #589, opened 2026-08-29
feat: optionally fail a scan when validation cases match no scanned input

kaifronsdal:
## What

Adds an opt-in `strict` flag to `ValidationSet` (and to the `validation_set()` factory). When it is set, a scan that runs to completion without every case in the set having matched a scanned input raises `ValidationCoverageError`, listing the unmatched case ids per scanner:

```
Strict validation: 2 validation case(s) matched no scanned input.
  ctf_environment (2): Fg3KBpgFr6RSsEWmHBUqeo, VFkCH7gXWpJYUYonvfHxrG
Scan results were written to scans/2025-11-14T09-33-12+00-00_ctf-environment_bMLcBnvVzX.
Check that these case ids match the ids of the inputs that were scanned, or set 'strict' to False on the validation set to ignore unmatched cases.
```

The results are written first and the completed scan's `Status` is carried on the error (`ex.status`), so nothing is lost — the scan itself succeeded, the label coverage didn't. `scout scan` prints the message and exits non-zero. The default is `False`, so behavior is unchanged unless you opt in.

## Why

`_validate_scan()` matches cases to scanner inputs by id, and a case that matches nothing is dropped without a trace: no result row, no error, no warning. The accuracy/precision/recall reported in the scan summary and in Scout View can therefore be computed from a fraction of the labels while looking entirely healthy. Ways to end up there without noticing:

- transcript ids drift (re-imported or re-run logs), so a validation file written last week matches half of this week's scan
- the scan reads a different or stale transcripts database than the one the cases were created against
- a `.where()` filter, `limit`, or a `worklist` excludes labelled transcripts
- `transcripts.for_validation()` doesn't protect against this either — it filters `sample_id IN (...)`, so ids that aren't present just yield fewer transcripts

This is exactly the failure you want to be loud in CI, where a validation set is the regression gate for scanner prompt changes.

## API choice

Two candidate surfaces: a field on the validation set, or a `scan()`-level flag next to `fail_on_error`. I went with the field on `ValidationSet` because:

- **It's a property of the label set, not of a run.** "This file is the complete ground truth for this scanner" is a statement about the labels. It also gives per-scanner granularity for free — one scanner in a scan can be strict while another is spot-checked with a partial set.
- **It rides along the paths that already carry validation sets.** `scan()`, `scan_async()`, `ScanJob(validation=...)`, and a `@scanjob` that builds its sets with `validation_set(..., strict=True)` all work with no new plumbing.
- **It is recorded in `ScanSpec`, so `scan_resume()` re-checks on the same terms.** A runtime-only `scan()` argument would silently lose strictness on resume, or would need a new `ScanOptions` field — which touches `openapi.json` anyway, so the schema cost isn't avoided by the alternative design.

Placement of the check: coverage is evaluated when the scan completes, not up front, because case ids can address messages, events, or groups of them (custom loaders), none of which are known before transcripts are read. It reads the validation entries recorded in the scan summary, which means cases validated by an earlier attempt of a resumed scan still count. Only a scan that reached `complete` is checked — an interrupted or errored scan already reports itself as such, and its cases genuinely haven't all had their chance yet.

`ValidationCoverageError` subclasses `PrerequisiteError` (inspect_ai's user-facing "fix your configuration" error, which prints without a traceback and exits non-zero) and adds `.status`.

## Testing

New tests in `tests/test_validation_e2e.py` (Strict Coverage section):

| Case | Expectation |
|---|---|
| strict, all cases match | scan completes, both cases validated |
| strict, one unmatched case | `ValidationCoverageError` naming the id and scanner but not the matched ids; `ex.status.complete` and results readable at `ex.status.location` |
| default (lenient), one unmatched case | scan completes, unmatched case ignored (pre-existing behavior) |
| two scanners, only one strict | only the strict scanner is reported |
| composite (multi-id) cases | matched pair passes, unmatched pair reported as `"id1,id2"` |
| 26 unmatched cases | list truncated at 25 with `(1 more)` |

`tests/scan/test_strict_validation_resume.py` covers the resume path: a case validated before an interruption counts toward coverage on `scan_resume()`.

Checks: `ruff check`, `ruff format`, and `mypy src examples tests` clean; full `pytest` suite 3234 passed / 107 skipped.

## Generated types

`openapi.json` is regenerated (adds `strict` to `ValidationSet`). `apps/scout/src/types/generated.ts` in the ts-mono submodule is **not** — that needs the coordinated submodule flow, which is why this is a draft (the `check-schema-and-types` failure is exactly this: the only drift is the two `strict: boolean` additions to the generated `ValidationSet` types). Note that the schema generator marks non-nullable fields required by design ("required is determined by nullability, not defaults" in `CustomJsonSchemaGenerator`), so the TS type gains a required `strict: boolean` and any `ValidationSet` object literal in `apps/scout` will need it. This matches the existing convention rather than creating a new case: `openapi.json` already has 16 required boolean-with-default fields (e.g. `Summary.complete`, `MessageFormatOptions.exclude_system`), so keeping `strict` non-nullable and following the coordinated regeneration flow seems right; happy to hand that off.

## Interaction with #588 (score predicates)

Reconciled — the fix is included in #588: it now records a `ValidationEntry` for every validated case (scored cases get `valid: None` plus a `score` field, and the confusion-matrix metrics skip them), so this PR's coverage check sees scored cases as covered and the two PRs are order-independent. Verified empirically on a local merge of both branches: a strict validation set whose cases were all matched by a score-returning predicate falsely raised `ValidationCoverageError` before the reconciliation and completes cleanly with it, while a genuinely unmatched case still raises.

## Things I deliberately left alone

Found while working on this; per AGENTS.md I'm flagging rather than fixing. Happy to fix any of them here or in follow-ups:

1. **A single-element list id can never match.** `_validate_scan()` normalizes the scanned ids (`["abc"]` → `"abc"`) but not `case.id`, and `_parse_id()` produces `["abc"]` from a validation file cell containing `["abc"]`. Such a case is silently skipped today; with `strict` it now fails the scan (correctly reporting that it wasn't validated, though the real fix belongs in the matcher).
2. **Duplicate case ids.** `_validate_scan()` uses `next(...)`, so only the first case with a given id is evaluated; the coverage check compares against the set of validated ids, so duplicates count as matched. A duplicated row in a CSV still loses a label silently. A pre-flight duplicate-id check (independent of `strict`) would catch it.
3. **`scan_complete()` bypasses the check** — force-completing an errored scan marks it complete without verifying coverage.
4. **`_scan_one()` can reuse a stale `validation_result`.** It is initialized outside the loader loop, so an iteration that raises before validation attaches the previous iteration's `ResultValidation` to the new report; that records a validation entry for an input that was never validated. One-line fix (reset inside the loop).
5. **No CLI or config way to enable `strict`** — `-V scanner:file.csv:predicate` has no slot for it, and the scan job / project config forms take a file path rather than a `ValidationSet`. It can be set from a `@scanjob` in Python today; a CLI or config form could follow.

### Agent review

- Reviewer: Claude Opus 5 via a general review agent (fresh context, different model from the author, which was Claude Fable 5), 1 pass; second pass by Claude Sonnet 4.6 (fresh context) on the revised diff.
- Pass 1 findings: 13 — 5 fixed (list-valued ids were being swallowed as rich markup in the message; the error now carries the scan status instead of discarding it; misleading "confirm the ids correspond to transcripts" wording; test isolation via `temp_active_scans_store()`; a resume-test assertion that depended on summary double-counting), 4 addressed by new tests (composite ids, per-scanner strictness, truncation, results readable after the failure), 4 disclosed above as flagged-not-fixed. One design objection dismissed with disagreement: the reviewer wanted the scan marked incomplete on disk; the scan's work *is* complete, and marking it otherwise would misreport the run and invite a pointless resume, so the coverage failure is surfaced as an exception carrying the status instead.
- Pass 2 findings: 3 — 2 fixed (`BaseException` captured `(message, status)` in `args`, so `str(ex)` dumped the entire `Status`; the `status` attribute docstring moved into an `Attributes:` section), 1 dismissed (display ambiguity when a case id itself contains a comma — the CSV format can't express that either).
- Independent adversarial pass: Claude (Fable 5), fresh context, after the draft opened. Re-verified rather than trusted the claims above: local `ruff`/`mypy`/full pytest at branch HEAD (3234 passed); no race between the coverage check and result recording (the summary is updated synchronously under a lock in `RecorderBuffer.record`, and `recorder.sync` runs after the strategy completes); the `str(ex)` fix (`BaseException.__new__` does capture both constructor args, and `args=(message, status)` also keeps the exception picklable); the CI failure hides nothing beyond the expected `generated.ts` drift (regenerating `openapi.json` locally is a no-op); and the `scan_complete()` bypass flag is accurate (`scan_complete_async` calls `recorder.sync(complete=True)` directly — it is an explicit "complete despite errors" override, so the bypass is arguably coherent, but worth a maintainer decision). New findings: the semantic interaction with #588 (section above), and a note that `repr(ex)` still embeds the full `Status` via `args` (no default display path uses `repr`, so left as-is). No code changes were needed on this branch.

### DIFF
--- a/docs/reference/results.qmd
+++ b/docs/reference/results.qmd
@@ -22,6 +22,7 @@ reference: inspect_scout
 ### validation_set
 ### ValidationSet
 ### ValidationCase
+### ValidationCoverageError
 ### PredicateType
 ### PredicateFn
 ### ValidationPredicate
--- a/docs/validation.qmd
+++ b/docs/validation.qmd
@@ -103,6 +103,50 @@ For YAML/JSON files, use a labels key instead of target:
 
 ---->
 
+## Strict Coverage {#strict-coverage}
+
+Validation cases are matched to scanner inputs by id, and by default a case whose id matches nothing that was scanned is ignored. This keeps partial validation sets convenient (see [Validation Splits](#validation-splits)), but it also means a scan can report validation metrics for a fraction of your cases without saying so — for example when transcript ids have changed, when the scan reads a different transcripts database, or when a transcript filter excludes labelled transcripts.
+
+Set `strict` on the validation set to require that every case matches something that was scanned:
+
+``` python
+from inspect_scout import scan, transcripts_from, validation_set
+
+scan(
+    scanners=[ctf_environment()],
+    transcripts=transcripts_from("./logs"),
+    validation=validation_set("ctf-environment.csv", strict=True)
+)
+```
+
+If any case matched no scanned input, the scan writes its results as usual and then raises `ValidationCoverageError`, listing the unmatched case ids per scanner:
+
+``` default
+Strict validation: 2 validation case(s) matched no scanned input.
+  ctf_environment (2): Fg3KBpgFr6RSsEWmHBUqeo, VFkCH7gXWpJYUYonvfHxrG
+Scan results were written to scans/2025-11-14T09-33-12+00-00_ctf-environment_bMLcBnvVzX.
+Check that these case ids match the ids of the inputs that were scanned, or set 'strict' to False on the validation set to ignore unmatched cases.
+```
+
+The error carries the completed scan's `status`, so results are still available programmatically:
+
+``` python
+from inspect_scout import ValidationCoverageError, scan
+
+try:
+    status = scan(...)
+except ValidationCoverageError as ex:
+    status = ex.status  # results were written; coverage was incomplete
+```
+
+`scout scan` prints the message and exits non-zero.
+
+Coverage is checked when the scan completes rather than before it starts, since case ids can also address messages or events, which aren't known until the transcripts are read. Scans that are interrupted or that end with errors are reported as usual and aren't checked for coverage.
+
+Because coverage is measured against what was actually scanned, anything that narrows the scan can leave cases unmatched: a transcript filter, `limit`, or a `worklist`. Narrowing the scan *to* the validation set with `transcripts.for_validation()` doesn't guarantee coverage either — case ids that aren't in the transcripts simply match nothing — so `strict` is useful there too.
+
+`strict` is a property of the validation set rather than of a single scan, so it applies per scanner (each scanner in a scan can have its own setting) and is recorded in the scan spec — a scan resumed with `scan_resume()` checks coverage on the same terms as the original scan, counting the cases that were validated before the interruption.
+
 ## Predicates
 
 By default, validation compares scanner results to targets using equality (`eq`). You can specify different comparison predicates either per-case or as a default for all cases.
--- a/src/inspect_scout/__init__.py
+++ b/src/inspect_scout/__init__.py
@@ -25,6 +25,7 @@ from ._recorder.recorder import (
 )
 from ._recorder.summary import Summary
 from ._scan import (
+    ValidationCoverageError,
     scan,
     scan_complete,
     scan_resume,
@@ -203,6 +204,7 @@ __all__ = [
     "PredicateType",
     "PredicateFn",
     "validation_set",
+    "ValidationCoverageError",
     # version
     "__version__",
 ]
--- a/src/inspect_scout/_scan.py
+++ b/src/inspect_scout/_scan.py
@@ -37,6 +37,7 @@ from inspect_ai.util import span
 from inspect_ai.util._anyio import inner_exception
 from pydantic import JsonValue, TypeAdapter
 from rich import box
+from rich.markup import escape
 from rich.table import Column, Table
 from typing_extensions import Unpack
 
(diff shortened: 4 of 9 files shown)


kaifronsdal:
CI status: everything passes except `check-schema-and-types`, which fails exactly as expected for this change — the submodule's `apps/scout/src/types/generated.ts` is missing the new `strict` property, and its diff in the job log is only the two `strict: boolean` additions. Regenerating it needs the coordinated ts-mono flow, so I left the submodule pointer untouched. Let me know if you would rather I open the ts-mono PR, or reshape the field (e.g. nullable) so the TS type stays optional.

PR #178, opened 2026-01-13
Fix ENG-432: Scan viewer showing 'No input available' despite data being retrieved

sjawhar:
## Summary

This PR fixes ENG-432 where the scan viewer showed "No input available" despite input data being retrieved successfully.

### Bug Fix
- **Root cause 1**: `setLoadingData` in store.ts was incorrectly modifying `state.loading` instead of `state.loadingData`
- **Root cause 2**: React Query loading state wasn't being passed to `ResultBody` component

### Improvements
- Added `inputLoading` prop to `ResultBody` and `InputRenderer` components
- Added detection and warning for duplicate scanner entries in worklist (previously silently dropped)
- Added explicit type annotations to `ScanDisplayLog` and `TextProgressLog` classes

### Tests
- Added frontend tests for `ResultBody` loading state behavior
- All existing display module tests pass (23 tests)

## Test plan
- [x] Frontend tests pass (52 tests including 3 new ones)
- [x] Display module tests pass (23 tests)
- [x] Type checker shows no new errors
- [x] Manually verified test fails on main, passes with fix

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #592, opened 2026-08-30
docs: codify panel-ownership rule for async query lifecycle

sjawhar:
## Summary

Adds a short `### Architecture` subsection to AGENTS.md's TypeScript section stating the panel-ownership rule that is already enforced in review and implemented in the codebase: a panel that initiates an async query owns its `AsyncData` lifecycle (loading/error rendering); components that only render once data is ready take required data props instead of reading query/loading state themselves.

## Why

The rule is already the enforced convention, but it is not written anywhere an agent or new contributor would find it:

- #178 fixed a bug caused by loading state not being threaded down to a leaf component (`inputLoading` missing on `ResultBody`, producing a false "No input available" state).
- #180 (merged) is the direct follow-through, and its own description states the rule this PR documents: "Lift loading state handling up to top level panel components as much as possible... Pass `inputData` as prop to `ResultPanel` instead of fetching internally."
- #194 review applied the same boundary discipline to grid-API access.

AGENTS.md was recently expanded with async guardrails and PR conventions (#585), but those cover the Python side; the frontend design docs (`react-query.md`, `front-end-testing.md`) don't state this rule either. This just writes down what contributors are already held to, in the file agents read.

### Agent review
- Reviewer: Claude (this PR was drafted and reviewed by Claude sessions acting for Sami; same-model, not fresh-context)
- Passes: 3 — an initial line-by-line duplication check of the new text against AGENTS.md, CONTRIBUTING.md, and design/*.md at drafting time; a second pass six days later re-verifying against current main after #585's AGENTS.md expansion (no overlap: #585's async guardrails are Python-side) and re-checking `react-query.md`/`front-end-testing.md` in the ts-mono submodule; and a thermonuclear-deep-review pass that re-verified the rule against current code (`ScannerResultPanel.tsx:180-181,195-200,383-384,449-452`; `ResultBody.tsx:24-26`'s required `inputData: ScannerInput` prop) and found the rule text itself sound, but flagged the `(see #178, #180)` parenthetical on the AGENTS.md line as doubly wrong: #178 is closed-unmerged, and even citing it alongside #180 invites a reader to treat both as equally authoritative evidence.
- Findings: the third pass's finding is fixed in this PR — the `(see #178, #180)` parenthetical was removed from the AGENTS.md line (that provenance belongs in this PR body, which already correctly identifies #180 as the merged follow-through, not in the evergreen rule text). No other change; the rule text and placement were otherwise sound. Placement was adjusted once before this (originally drafted for CONTRIBUTING.md, moved to AGENTS.md since that is the file agents read — the same reasoning as #585).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -156,6 +156,9 @@ pnpm install
 - Strict mode enabled; no `any`, no type assertions
 - ESLint + Prettier
 
+### Architecture
+- Panels that initiate an async query own its `AsyncData` lifecycle and render loading/error states at that boundary; components rendered only after data is ready take required data props rather than reading query/loading state themselves
+
 ### Common Pitfalls
 - Use pnpm, not npm—this project uses pnpm exclusively
 - Hook tests don't need JSX—use `.test.ts` not `.test.tsx`; see `useMapAsyncData.test.ts`


sjawhar:
Pushed a correction after a thermonuclear docs-review pass (head moved to `6ea87f60`):

- The rule itself checked out against current code (`ScannerResultPanel.tsx:180-181,195-200,383-384,449-452` owns `resultLoading`/`detailLoading` and gates `ResultPanel` on `inputData` being ready; `ResultBody.tsx:24-26` declares `inputData: ScannerInput` as a required prop) — no change needed there.
- Removed the `(see #178, #180)` parenthetical from the AGENTS.md line. #178 is closed-unmerged and #180 is the merged follow-through that actually implements the rule, so citing them side-by-side on the evergreen line was misleading; nothing is lost since this PR body already correctly identifies #180 as the merged source.

PR description's `### Agent review` block updated to disclose this third pass and its finding.

epatey:
Thanks for writing this up, and for trimming the #178/#180 citations after the earlier review.

I'm going to close this one. The rule as written frames it as a panel "owning its AsyncData lifecycle," and that does not match the model. AsyncData is a value derived from a react-query result (useAsyncDataFromQuery), and react-query shares and refcounts a single cache entry across every observer by design, so nothing owns it. The real preference here is narrower: panel-level components consume the AsyncData hooks and gate on them, not mounting their children until the data is ready, so the children stay simple and never have to deal with loading. That is a gating rule, not ownership, and prop-drilling resolved data is only one way to satisfy it (AsyncGate gates and lets children read the resolved value instead).

Two other reasons to hold off. The code does not follow this uniformly yet, so documenting it as an established convention is a bit ahead of where things are. And it belongs in ts-mono, not here: AsyncData, the gate helpers, and the scout viewer components all live in ts-mono, which inspect_scout only embeds through the _view/ts-mono submodule, so AGENTS.md guidance about them should live there.

If it is worth writing down, I would reframe it around gating rather than ownership and put it in ts-mono. Appreciate you raising it.

PR #595, opened 2026-08-31
fix: offload remaining blocking I/O in scans endpoints

WatchTree-19:
Partially closes #363.

Checked all 15 endpoints named in the issue against `main`. Most are already fixed: every endpoint in `_api_v2_validations.py` is already plain `def`, and in `_api_v2_config.py` `config` and `get_project_config` are already `def` while `put_project_config` already offloads correctly before awaiting. The three in `_api_v2_scans.py` are the only ones left, so this PR is limited to those.

Following the decision tree in `design/fastapi-async.md`:

- `active_scans` — no awaits, sync sqlite access, so converted to plain `def` for FastAPI's auto-threadpool.
- `run_llm_scanner` — awaits and blocks, so kept `async def` and offloaded `_spawn_scan_subprocess`, the temp-file cleanup, and `proc.wait` via `anyio.to_thread.run_sync`.
- `delete_scan` — awaits `notify_topics` and blocks, so offloaded `scan_path.exists()`, the active-scans store read, and `send2trash`.

Added two small module-level helpers (`_remove_if_exists`, `_read_active_scans`) since `run_sync` can't take a context manager or a multi-statement block inline.

No control-flow changes: same exceptions, status codes and ordering.

Verified locally: `ruff check src/` and `ruff format --check src/` both clean, and 208 tests covering the view and scans API pass.

Happy to fold this into #584 if you'd rather have one PR, though they're independent changes so I've kept them separate.

### DIFF
--- a/src/inspect_scout/_view/_api_v2_scans.py
+++ b/src/inspect_scout/_view/_api_v2_scans.py
@@ -10,6 +10,7 @@ import tempfile
 import threading
 import time
 import zipfile
+from functools import partial
 from typing import Any, Iterable
 
 import anyio
@@ -161,8 +162,10 @@ def create_scans_router(
         summary="Get active scans",
         description="Returns info on all currently running scans.",
     )
-    async def active_scans() -> ActiveScansResponse:
+    def active_scans() -> ActiveScansResponse:
         """Get info on all active scans from the KV store."""
+        # Sync sqlite access and no awaits: a plain `def` endpoint is run by
+        # FastAPI in a threadpool, keeping the event loop free.
         with active_scans_store() as store:
             return ActiveScansResponse(items=store.read_all())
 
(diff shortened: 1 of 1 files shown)


epatey:
Thanks for the careful survey here. Your analysis of what's already fixed matches main, and the offloading itself follows the design doc correctly.

Unlucky timing though: all three of these endpoints exist only to serve a run-scan feature that's disabled behind a build-time flag (`SCOUT_RUN_SCAN`) that no build sets, so it has never been enabled in a shipped viewer. Rather than polish dead code, I've opened #598 to delete the feature and these endpoints, so I'm closing this PR. The validations and config parts of #363 were already fixed, so that issue is now covered too.

Sorry the issue pointed you at code that's going away, and thanks again for the legwork.

WatchTree-19:
No problem, and thanks for explaining rather than just closing it. Deleting the feature is clearly better than polishing endpoints nobody can reach, and it is useful to know the flag was never set in a shipped build.

Good to hear #363 is covered either way.

I saw #587 landed the tg_collect convention work, so that one is done too. If there is anything on the board you would rather see picked up, I would be glad to take it. Otherwise I will find something.

PR #597, opened 2026-08-31
ci: gate inline lint/type suppressions behind a reviewed ledger

epatey:
Adds the inline-suppression ledger and gate, ported from inspect_ai's Python version (meridianlabs-ai/inspect_ai#5136). Every inline suppression (`# noqa`, `# type: ignore`, `# pyright: ignore`, and the file-wide forms) is recorded in a committed `suppressions.json`. CI fails when code and ledger disagree, so every add, remove, or move of a suppression shows up as a reviewable ledger diff instead of slipping through.

## What's in it

- `scripts/check_suppressions.py`: stdlib-only, tokenize-based scanner plus the gate. Directives in string literals never count; a `type: ignore` counts only when it opens the comment; `# type: ignore [code]` with a space is code-scoped; a bare own-line `# type: ignore` before any code is file-wide. `--update` regenerates the ledger and refuses to grow any rule's repo-wide reason-less count (new suppressions must carry a reason in a trailing `# reason` segment, the only style mypy accepts). `--update --allow-growth` is the loud escape hatch for wholesale code imports.
- `scripts/suppressions_pr_delta.py`: renders the sticky PR comment body.
- `.github/workflows/suppressions.yml`: the gate plus its tests, on `pull_request` with a read-only token.
- `.github/workflows/suppressions-comment.yml`: the sticky delta comment, on `pull_request_target` so fork PRs get it too. It runs only base-branch code and reads the PR's ledger as data via `git show`, never checking out or installing PR code. The security model is documented at the top of the file; please read it before editing that workflow. Because `pull_request_target` runs the base branch's workflows, no comment appears on this PR itself; it starts working for PRs opened after this merges (the comment body it would have posted here was rendered locally, see below).
- `make suppressions-check` / `make suppressions-update`, with the check folded into `make check`.
- An AGENTS.md "Suppression gate" section with the rules and the reason convention.
- `tests/test_check_suppressions.py`: the ported test suite (67 tests).

## Baseline

`suppressions.json` captures the repo as it stands: 274 suppressions, 270 without a reason, across 81 files. No suppressions were removed in this PR; the reason-less ones burn down over time under the ratchet. Largest rules: `type: ignore[arg-type]` (70), `noqa:E712` (28), bare `type: ignore` (22).

## Verified locally

- Gate passes on the baseline; the 67 ported tests pass.
- A fresh reason-less `# type: ignore[assignment]` fails the check (NEW + UNDESCRIBED) and `--update` refuses to record it.
- The same suppression with a trailing `# reason` segment passes `--update` and the delta script renders the sticky-comment body for it.
- Reverting returns the gate to green.
- ruff, ruff format, and mypy are clean on the touched files.

## Coexistence with #594

#594 is the config-level sibling (a ledger for checker relaxations in pyproject.toml). The two gates coexist on purpose: separate ledgers, separate scripts, separate CI jobs, distinct Makefile targets and AGENTS.md sections. Both branch off main, so they overlap in the Makefile `check:` line and nearby AGENTS.md text; whoever merges second resolves that small conflict. The two could be unified later if that ever seems worth it.

## Port adaptations

Scripts live in `scripts/` (scout convention, matching #594) rather than `.github/scripts/`. Scout's stricter ruff rules required `strict=` on a `zip()` (with an early return for the empty-match case, covered by the ported tests) and an explicit `check=False` on a `subprocess.run` in the test helper. The test file gained a typed `Ledger` alias for scout's strict mypy. The AGENTS.md ratchet-exception wording was adjusted since scout is not a fork.

### Agent review

- Author: Claude (Fable 5). Reviewer: /code-review at high effort (fresh context, same model family), one pass: 8 finder angles, 1-vote verification, 10 findings.
- 1 fixed pre-review (a port-introduced `zip(strict=True)` crash on comments with no directive match, caught by the ported tests), 9 dismissed with reasons:
  - Inherited from the reviewed upstream reference, kept for parity and fixable there first: wrong-shape PR ledger crashes the comment job while broken JSON silently becomes `{}` (data problem; the read-only gate still enforces), the comment compares against the base tip rather than the merge-base, the sticky-comment lookup doesn't filter by author, hand-rolled argv parsing, lowercase noqa codes counted although ruff rejects them (overcount, fails safe), and untracked files included in `--update` scans (deliberate per the code comment, so pre-`git add` updates work).
  - Port-parity choices: the gate workflow re-runs the test file py-test also runs (intentional: proves the gate passes in a bare environment with no project install); bare `python3` in the Makefile matches the existing targets' assumption that the dev env is on PATH (repo requires Python >= 3.10).
  - Out of scope: enforcing "never the bare code-less form" via ruff PGH004 / mypy `ignore-without-code` is a checker-config change that belongs with #594's territory and would need the 270 baselined entries handled; possible follow-up.

### DIFF
--- a/.github/workflows/suppressions-comment.yml
+++ b/.github/workflows/suppressions-comment.yml
@@ -1,0 +1,85 @@
+# Posts the sticky suppression-ledger delta comment on PRs, fork PRs
+# included — see scripts/suppressions_pr_delta.py and the
+# "Suppression gate" section in AGENTS.md.
+#
+# SECURITY MODEL — read before editing.
+#
+# This workflow runs on pull_request_target, which executes in the BASE
+# repo's context with a WRITE token even for fork PRs. That is the only
+# reason it exists: the plain pull_request token on a fork PR is read-only
+# and cannot post the comment (enforcement lives in suppressions.yml,
+# where read-only is enough). A pull_request_target workflow that runs
+# PR-controlled code hands that write token to the PR author, so this job
+# keeps three invariants:
+#
+#   1. Only BASE code runs. The checkout below has no `ref`, so it checks
+#      out the base branch: the delta script and the base ledger are the
+#      base repo's own trusted copies. A PR's changes to the script take
+#      effect only after a maintainer merges them.
+#   2. PR content is data, never code. The PR's suppressions.json is read
+#      with `git show` on the fetched head ref — that reads a blob and
+#      executes nothing. The PR head is never checked out.
+#   3. No dependency installs. The delta script is stdlib-only by design;
+#      installing the project (pip/uv) would execute PR-controllable
+#      setup and dependency code.
+#
+# Do NOT "fix" this workflow by checking out the PR head, running any
+# script from it, or adding an install step — any of those breaks the
+# security model.
+name: Suppressions comment
+
+on:
+  pull_request_target:
+    types: [opened, synchronize, reopened]
+    branches:
+      - main
+      - "release/**"
+
+permissions:
+  contents: read
+  pull-requests: write
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  comment:
+    runs-on: ubuntu-latest
+    steps:
+      # No `ref`: on pull_request_target this checks out the BASE branch
+      # (trusted script + base ledger). Never add ref: pull_request.head.
+      - uses: actions/checkout@v4
+
+      - name: Comment ledger delta on PR
+        env:
+          GH_TOKEN: ${{ github.token }}
+          GH_REPO: ${{ github.repository }}
+          PR: ${{ github.event.pull_request.number }}
+        run: |
+          # The PR's ledger is fetched and read as DATA: `git show` reads
+          # a blob and executes nothing. The PR head is never checked out.
+          git fetch --depth=1 origin "pull/$PR/head"
+          git show FETCH_HEAD:suppressions.json > /tmp/head-suppressions.json 2>/dev/null \
+            || echo '{}' > /tmp/head-suppressions.json
+          # The workspace is the base branch, so its ledger is the base side.
+          if [ -f suppressions.json ]; then
+            cp suppressions.json /tmp/base-suppressions.json
+          else
+            echo '{}' > /tmp/base-suppressions.json
+          fi
+          # Base script, system python, stdlib only — no installs (see the
+          # security model above).
+          body=$(python3 scripts/suppressions_pr_delta.py /tmp/base-suppressions.json /tmp/head-suppressions.json)
+          # Sticky comment: match on the marker, not --edit-last, so other
+          # bot comments on the PR are never clobbered.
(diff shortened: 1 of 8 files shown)


epatey:
@auto

i-am-marvin:
**Claude finished @epatey's task in 4m 22s** —— [View job](https://github.com/meridianlabs-ai/inspect_scout/actions/runs/33429981816)

---
### Status check & review pass — all green, no blocking findings

- [x] Gather context (CI status, branch freshness vs `origin/main`)
- [x] Run the gate and its test suite locally
- [x] Review the PR changes
- [x] Report status/findings

The bare `@auto` gave no specific instruction, so I verified the PR end-to-end and did an independent review pass. No code changes were made or needed.

### Verified

- **CI: all 13 checks pass** (py-test/mypy/ruff/build on 3.10 and 3.14, schema/types, js-dist, submodule-on-main, PR title lint, and the new `suppressions` gate itself); the two skips are release-only jobs.
- **Branch is current with `origin/main`** (0 commits behind, clean tree — the merge of main is already in).
- **Locally reproduced the gate**: `python3 scripts/check_suppressions.py` reports the ledger matches (274 suppressions, 270 without a reason, 81 files — exactly what the PR description claims), and the 67 tests pass under the same bare invocation CI uses (`pytest --noconftest -o addopts=""`, no project install).

### Review notes

I read both scripts, both workflows, the Makefile/AGENTS.md changes, and the test suite. The `pull_request_target` security model in `suppressions-comment.yml` holds up as written: only base-branch code runs (no `ref` on checkout), the PR's ledger is read as data via `git show` without ever checking out PR code, no dependency installs, and untrusted ledger strings are HTML/pipe-escaped before landing in the comment table (`_cell` in `scripts/suppressions_pr_delta.py:30`). The `|| echo '{}'` fallback also correctly handles PRs that delete the ledger. The directive scanner's edge-case handling (inert mid-comment mypy directives, legacy type comments, the preamble file-wide `type: ignore`, ratchet-vs-move semantics) is thoroughly covered by the tests.

Two minor observations, neither blocking:

- `.github/workflows/suppressions.yml:33` pins `actions/setup-python@v4`; v5 is current (v4 runs on the deprecated Node 16 runtime). Worth bumping whenever convenient — I can't edit workflow files myself (GitHub App permissions exclude `.github/workflows`). [Fix this →](https://claude.ai/code?q=In%20.github%2Fworkflows%2Fsuppressions.yml%20on%20branch%20feat%2Fsuppression-gate%2C%20bump%20actions%2Fsetup-python%40v4%20to%20%40v5&repo=meridianlabs-ai/inspect_scout)
- `scripts/suppressions_pr_delta.py:104` unpacks `sys.argv[1:3]` and raises a bare `ValueError` when invoked with fewer than two paths — already covered by the PR description's "hand-rolled argv parsing" dismissal, and the workflow always passes both, so fine as-is.

The remaining dismissals in the PR's Agent review section all check out against the code (base-tip vs merge-base comparison, author-unfiltered sticky lookup, untracked-file scanning per the comment at `scripts/check_suppressions.py:355`, etc.).

**Bottom line: the PR does what it says, CI is fully green, and I found nothing that needs to change before merge.**

---
 · [Branch](https://github.com/meridianlabs-ai/inspect_scout/tree/feat/suppression-gate)

PR #603, opened 2026-09-02
fix: attachment contents no longer come back as dangling attachment:// hashes

rasmusfaber:
## Summary

Attachment content stored inside `events_data` pool entries was never resolved, so scanners and the viewer received literal `attachment://<hash>` text where the content should be. The pool section is parsed *after* `attachments`, so a membership test taken while the attachments stream past cannot yet know that a pooled message will need one.

Attachments are now retained whenever the transcript can still turn out to need them, and filtered to the referenced set when it provably cannot.

fix: an erroring scan item no longer reports the previous item's validation result

## The retention trade

Bounding this correctly is the whole difficulty, and the decision is wrong in one direction or the other if you get it slightly off. The rule now: if events are not being collected at all, no event can ever need a pool entry and the referenced-ID filter applies — this is what keeps the common messages-only scan bounded. If events are being collected, the table is retained, because a pool-ref-carrying event may still be ahead in the stream. The check is a one-time snapshot at the first `attachments.*` key, not a running condition.

For a messages-only scan with 200 unrelated attachments, that means one attachment retained rather than all 201.

## Notes for reviewers

This is the first of six PRs carved out of #491, which was closed as too large to review. It stands alone on `main` and depends on none of the others.

An `attachment://` reference only counts when it is the entire string value, matching `inspect_ai`'s own `create_attachment`/`startswith` semantics. A reference appearing inside a longer string is user-authored text and is deliberately left alone.

### Agent review

- Author: Claude Opus 5. Reviewed across several fresh-context passes — a five-dimension analysis (types, bugs, simplification, structure, tests), an independent `codex` pass, and per-round scoped re-reviews. Multiple passes, each in a fresh context.
- Findings on this slice: the retention rule was wrong three separate times before this state (under-retaining, then unbounded, then order-dependent on JSON key order); each was caught by review or by a test in a sibling PR, and each direction now has a regression test that fails if only that direction breaks.
- One fix was reverted on evidence: an earlier version also resolved references embedded inside larger strings, which corrupted user text. `inspect_ai` does not do this, and neither does this PR now.

### DIFF
--- a/src/inspect_scout/_scan.py
+++ b/src/inspect_scout/_scan.py
@@ -81,6 +81,7 @@ from ._scanjob_config import ScanJobConfig
 from ._scanner.loader import config_for_loader
 from ._scanner.result import Error, Result, ResultReport, ResultValidation, as_resultset
 from ._scanner.scanner import Scanner, config_for_scanner
+from ._scanner.types import ScannerInputNames
 from ._scanner.util import get_input_type_and_ids
 from ._scanspec import ScanSpec, Worklist
 from ._transcript.transcripts import ScannerWork, Transcripts, TranscriptsReader
@@ -1066,6 +1067,13 @@ async def _scan_one(
         result: Result | list[Result] | None = None
         final_result: Result | None = None
         error: Error | None = None
+        validation_result = None
+
+        # Reset per iteration and before the `try`: the error path below
+        # reads this, so a raise inside the `try` must not leave it holding
+        # the previous item's ids (this runs in a loop) or unbound.
+        type_and_ids: tuple[ScannerInputNames, list[str]] | None = None
+
         try:
             type_and_ids = get_input_type_and_ids(loader_result)
             if type_and_ids is None:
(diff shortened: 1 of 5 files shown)


epatey (approved):
Approving at 294ef27a300c68aedf3f0d8a9a9b4f7f9e18d6c6.

I reproduced both bugs against main before approving. With only the three source files reverted to main, the pool-ref test returns the literal `attachment://b1c2d3e4f5a678901234567890123456` where the prompt text belongs, and the errored row's `validation_result` reads `true` instead of null. Both pass on this head, and I confirmed the whole-value rule against inspect_ai: `create_attachment` returns the entire value and `mm3_hash` is always exactly 32 hex characters, so the length-45 match is exact.

The one red check, `check-schema-and-types`, is main's own drift rather than anything in this PR. This branch does not touch `openapi.json`, so the job regenerated main's committed file and found a diff; its only hunk is inspect_ai's new `fail_on_refusal` field, which arrived upstream after main last regenerated on 2026-09-13. The fix belongs on main, because regenerating here would also move the ts-mono pointer. Squash auto-merge is armed and will fire once that lands and this branch re-runs green.

PR #604, opened 2026-09-02
perf: bound the segments an llm_scanner holds and scans at once

rasmusfaber:
## Summary

An `llm_scanner` over a long transcript rendered every segment into a list before scanning any of them, then issued one model call per segment with no ceiling. On a transcript that segments into hundreds of chunks, that is hundreds of rendered copies of the conversation resident at once.

Segments are now produced lazily and admitted through a fixed window, so at most `_SEGMENT_WINDOW_CAP` rendered segments exist at a time.

## Why a fixed cap and not a derived one

The window deliberately does *not* try to match the model's connection limit. `inspect_ai` already gates real requests with its own per-model semaphore, so a task admitted beyond that limit simply waits — it costs one rendered string, which is exactly what this cap bounds. An earlier version derived the window from `max_connections`; it read the unmerged config and mirrored only one of `inspect_ai`'s two concurrency strategies, so it under-bounded for most model shapes while claiming the opposite.

This is a behaviour change for any existing multi-segment `llm_scanner` user: fewer segments are in flight at once.

## Notes for reviewers

One of six PRs carved out of #491, which was closed as too large to review. This one stands alone on `main` and depends on none of the others.

`stream_segment_messages` lands here with no caller — its consumer is the viewer LLM-search slice, later in the series. It is a streaming counterpart to `segment_messages` with identical sizing rules, proven by parametrizing the existing suite over both.

### Agent review

- Author: Claude Opus 5, reviewed across several fresh-context passes including a five-dimension analysis and an independent `codex` pass.
- The derived-window removal above came out of that review: two reviewers independently found the derivation read the wrong config object, and a third had given it a clean bill.
- One proposed change was declined on evidence: switching the `ExceptionGroup` unwrap to `inner_exception` looked like a simplification but walks `__context__` and selects from an unordered set, which surfaces an unrelated `RuntimeError` as the user-visible scan error.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,6 +17,7 @@ dependencies = [
     "click>=8.1.3,!=8.2.0,!=8.2.2,!=8.3.0,!=8.3.1",
     "cloudpickle>=3.0.0",
     "duckdb>=1.4.0",
+    "exceptiongroup; python_version < '3.11'",
     "fastapi>=0.128.1",
     "ijson>=3.2.0",
     "importlib-metadata",
--- a/src/inspect_scout/_llm_scanner/_llm_scanner.py
+++ b/src/inspect_scout/_llm_scanner/_llm_scanner.py
@@ -1,7 +1,7 @@
-from functools import partial
-from typing import Any, Awaitable, Callable, Literal, cast, overload
+import sys
+from typing import Any, AsyncIterator, Awaitable, Callable, Literal, cast, overload
 
-from inspect_ai._util._async import tg_collect
+import anyio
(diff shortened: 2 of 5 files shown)


epatey (approved):
Approving at a5a60908. The fixed window of 16 is the right call; it does not need to be configurable.

check-schema-and-types is red from main's own drift, not from anything in this PR: upstream inspect_ai added a generate-config field after main last regenerated openapi.json and the generated types, so main itself would fail that check today. I'm addressing that schema bump in a separate PR, which should give this one a path to green. Auto-merge is armed and will fire once the required checks pass.

PR #605, opened 2026-09-02
perf: write recorded transcript columns as UTF-8 bytes instead of str

rasmusfaber:
## Summary

`_serialize_input` built a `str` for the `input` and `input_data` columns, which pyarrow then re-encodes to UTF-8 on the way into the table. For a large transcript that is a second full copy of the payload — at 1, 2 or 4 bytes per character depending on the widest code point in the string, per PEP 393.

The recorder now hands pyarrow UTF-8 bytes directly.

## Notes for reviewers

One of six PRs carved out of #491, which was closed as too large to review. It stands alone on `main` and depends on none of the others.

The three source files change together by necessity: `result.py` emitting buffers while `buffer.py` still treated them as an unknown scalar would stringify them and write a literal `b'...'` into a public results column.

`bytes | bytearray` is deliberate throughout, and the `bytearray` arm looks dead in this PR because it is — the producer is the spool, which lands later in the series. Narrowing it to `bytes` silently re-encodes a multi-MB transcript as a JSON *string* into the `input` column.

### Agent review

- Author: Claude Opus 5, reviewed across several fresh-context passes including a five-dimension analysis and an independent `codex` pass.
- The buffer passthrough originally applied to every column, which crashed with `ArrowInvalid` on a metadata column holding non-UTF-8 bytes. Found by review, reproduced, and now scoped to the two columns that carry pre-serialized JSON.
- Three separate reviewers recommended narrowing the `bytes | bytearray` union. That recommendation is correct for this PR read in isolation and wrong for the series, for the reason above; it was declined with a probe rather than an argument.

### DIFF
--- a/src/inspect_scout/_recorder/buffer.py
+++ b/src/inspect_scout/_recorder/buffer.py
@@ -532,17 +532,23 @@ def _sanitize_component(name: str) -> str:
     return re.sub(r"[^a-zA-Z0-9_\-=+.,@]", "_", name)
 
 
-def _normalize_scalar(v: Any) -> Any:
+_SERIALIZED_INPUT_COLUMNS = frozenset({"input", "input_data"})
+"""Columns `ResultReport.to_df_columns` fills with pre-serialized UTF-8 JSON."""
+
+
+def _normalize_scalar(key: str, v: Any) -> Any:
     if v is None:
         return None
     if isinstance(v, bool):
         return v
     if isinstance(v, (str, int, float)):
         return v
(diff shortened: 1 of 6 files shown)


dragonstyle (approved):
Verified recorder, scanner and persistence regressions, base/head stored-value equivalence, and the real viewer Arrow decoder. At a 64 MiB payload, median total peak RSS fell 14.4% for pooled Unicode and 5.0% for unpooled Unicode; ASCII total peak was effectively unchanged. Approving.

PR #606, opened 2026-09-02
perf: parse large eval-log samples into a disk spool instead of memory

rasmusfaber:
## Summary

Reading a transcript out of an eval log materializes the whole sample in memory before anything can filter it, so peak usage scales with the largest sample rather than with what a scan actually needs. This adds the machinery to avoid that: three disk spools, a single-pass `ijson` parse that fills them, replay iterators that resolve attachment and pool references one item at a time, and a `TranscriptHandle` that exposes the result.

**Nothing calls it yet.** The first consumer is the viewer LLM-search slice, next in the series.

Four symbols in particular arrive with no production caller in this PR, so they read as
dead code here. Each is consumed by #608, the last in the series:

| symbol | consumed by |
| --- | --- |
| `BlobSpool.has()` (`_transcript/json/spool.py`) | `_transcript/json/passthrough.py` |
| `ByteSpool.read()` (`_transcript/json/spool.py`) | `_transcript/json/passthrough.py` |
| `StreamParseResult.spool_dir` (`_transcript/json/stream_parse.py`) | `_transcript/json/passthrough.py` |
| `SpooledTranscriptHandle.parsed_result()` (`_transcript/handle.py`) | `_scan.py` |

## Review scope

Fourth of six PRs carved out of #491, which was closed as too large to review. It sits on #603, so **until that merges GitHub shows #603's commits here too**. This PR's own contribution is `_transcript/json/{spool,stream_parse,hydrate}.py`, `_transcript/handle.py`, the touched pre-existing `_transcript/json/{reducer,load_filtered,pool}.py`, the `open()` plumbing in `transcripts.py`/`database/`/`eval_log.py`, and their tests.

Own work: 20 files changed, 2607 insertions(+), 82 deletions(-) (18 commits) — https://github.com/METR/inspect_scout/compare/split/attachment-and-scan-attribution-fixes...split/spool-streaming-parse

## Judgement calls worth your pushback

Three decisions taken here, that could reasonably go the other way:

- **64 MiB** as the spool threshold — below it, samples stay on the materialized path.
- **A timeline request forces materialization**, because resolving timeline UUIDs needs the full event set.
- **The JSON-error fallback re-reads the source**, so a malformed sample on S3 costs two ranged GETs.
- **Spools live under the files-cache temp dir** (`tempfile`, honours `TMPDIR`) — the same place remote logs are already downloaded. On hosts with a tmpfs `/tmp` the spool is RAM-backed; point `TMPDIR` at disk. No scout-specific knob is added.
- **Without events, only message-referenced attachments are spooled**, the same rule `load_filtered.py` applies; with events every attachment is kept because pool refs are only known at replay.

## Divergences from the materialized path

None remain that a caller can observe. Nested `ToolEvent.events` are hydrated into `Event`s on both paths (shared `json/hydrate.py`), and a materialized handle drops timelines it was not opened for, matching the spooled path — including on the spooled handle's NaN/Inf fallback, which re-reads through the same loader — all pinned by equivalence tests. The parse still skips the sample's `timelines` section; a timeline request never reaches it, because `open()` routes timeline requests to the materialized path.

### Agent review

- Author: Claude Opus 5, reviewed across several fresh-context passes — a five-dimension analysis (types, bugs, simplification, structure, tests), an independent `codex` pass, and scoped re-reviews after each fix round.
- Review found and fixed: lifecycle gaps in the spooled handle and its spools, a shared pool-sink counter that let overlapping shapes overwrite each other, and strictness in the nested-event hydration that could both raise and fabricate an event.
- A final whole-branch pass on the rebased series (fresh context, Claude Opus 5) found that the spooled handle's JSON-error fallback still returned stored timelines the handle was not opened for — the one caller-observable divergence left — fixed here with a regression test, then re-reviewed in a scoped pass.
- A portability regression was caught here twice: this module deliberately seeks under a per-spool lock rather than using `os.pread`/`os.pwrite`, and an earlier change to `_read_at` broke macOS before being reverted.


epatey (approved):
Approving at d21111295fd26163127834db6cbed2f1cba5ece7.

Nothing in `src/` calls the new `open()` yet, so the only change that reaches today's readers is the `hydrate_nested_tool_events()` call in `_resolve_attachments()`. The equivalence tests pin that against the materialized path, and the default suite passes on this head in CI and locally (687 passed, 2 skipped across `tests/scanner/test_load_filtered.py` and `tests/transcript`).

Not arming auto-merge here, deliberately. This has to land after #603: while #603 is open, a squash of this branch would carry its commits under a `perf:` message. #603 is armed and goes first. Once it lands I will merge main up here and take this one in order. The red `check-schema-and-types` is scout main trailing inspect_ai on `fail_on_refusal`, not anything on this branch, and it clears when main is regenerated.

epatey:
Merged `main` up to `4e94eb06` on your behalf, now that #603 has landed. Flagging what I resolved, since it needed a decision rather than an auto-merge.

#603 landed as a squash (`419a1903`), so main carries it as one commit while this branch carries its eleven originals and edits the same lines further. That conflicted in the two files #603 touched and this PR touches again. Three hunks, all resolved to this branch's side:

- `json/load_filtered.py` — the expanded cross-reference comment on the `_SECTION_*` block from `e2acd56d`, over main's one-line original.
- `json/reducer.py` — the typed `@_coroutine` from `dfce6434`, over main's `@_ijson_coroutine  # type: ignore`. This is the change the ledger drop depends on.
- `json/reducer.py` — `JsonTextWriter` and `spooling_metadata_coroutine` in `__all__`.

Two checks that the resolution is right: after resolving, both files are byte-identical to the pre-merge head `d2111129`, so main contributed nothing to them beyond #603's squashed content this branch already had; and the diff against main is now 20 files, +2605/-79, exactly this PR's own slice. Main's regenerated `openapi.json` and the ts-mono gitlink bump come through from `8e941500` untouched.

Locally on the merged tree: 755 passed, 2 skipped across `tests/scanner/test_load_filtered.py`, `tests/transcript` and `tests/scan`; ruff and mypy clean on the touched package; suppressions ledger matches.

No change to your code beyond the conflict resolution.

PR #607, opened 2026-09-02
perf: viewer LLM search no longer loads the whole transcript into memory

rasmusfaber:
## Summary

An LLM search in the viewer loaded the entire transcript into memory to scan it, so a multi-GB transcript was a multi-GB allocation before the first model call. It now streams the transcript off a handle instead.

Getting there needs a streaming timeline extractor, which works in two passes: pass 1 streams the handle with bulk content replaced by stubs and runs the *unmodified* `timeline_build` classifier over that skeleton; pass 2 re-streams and substitutes the full events back.

## Review scope

Fifth of six PRs carved out of #491, which was closed as too large to review. It sits on #606 and merges #604, so **until those merge GitHub shows their commits here too**. This PR's own contribution is `_transcript/timeline_stream.py`, the handle-acceptance changes in `_llm_scanner.py`, the scanner-signature widening in `_scanner/{_loaders,validate,scanner}.py`, `_view/_api_v2_search.py`, and their tests.

Own work: 14 files changed, 2206 insertions(+), 60 deletions(-) (13 commits). https://github.com/METR/inspect_scout/compare/split/spool-streaming-parse...split/streaming-timeline-and-viewer-search also shows #604 (merged here).

## Notes for reviewers

`TimelineSpan.agent_result` goes unpopulated on the streamed path. The stub blanks `ToolEvent.result` and reduces `ModelEvent.input` to system messages, which is every field all three of `_extract_agent_results`' sources read. It is acceptable because the only Python reader of the field is `_extract_agent_results` itself; the inspect_ai and scout viewers render it, but a streamed span is consumed only for its id and never serialized to a viewer, and these spans are transient — but the equivalence test asserts the two paths differ in *only* that field, so a future reader turns it into a failure.

The endpoint now holds the transcripts view and its DB connection open across the whole LLM call, rather than just an initial read.

### Agent review

- Author: Claude Opus 5, reviewed across several fresh-context passes — a five-dimension analysis, an independent `codex` pass, and scoped re-reviews per fix round.
- Two passes independently found and reproduced that `stream_timeline_messages` did not prune the `scorers` span the way `transcript_messages` does, so an events-scope handle scan sent the grader's rubric — expected answer included — to the scanner model; the fix applies the same `timeline_filter` in pass 1. The code is new in this PR, so nothing on `main` was affected. The tests missed it because they compared streamed output against an *unfiltered* reference rather than against `transcript_messages`; that reference is now the filtered one.
- The equivalence tests assert twice on purpose. `span.model_dump()` compares uuid lists, because `TimelineEvent` serializes each event as its uuid, so it cannot see a payload divergence; the rendered `messages_str` comparison can. Forcing a stub into rendered output fails the second and passes the first.
- Two further fresh-context passes after the rebase onto current `main` rejected a first attempt at a public `@scanner(supports_streaming=...)` argument in this PR (inert here, no sentinel, and an inverted precedence). That argument now lands in #608 together with the gate that reads it. This PR contributes what it builds on: the streaming vouch is an identity-keyed registry (`mark_streaming_support`/`streaming_support_of`) replacing a `__dict__` attribute that `functools.wraps` could copy onto a wrapper, `llm_scanner` records its verdict both ways, and a `MaterializedTranscriptHandle` takes the batch path rather than being streamed pointlessly.


epatey:
Merged main up to `afdbcd1b`, so 603, 604 and 606 are all on main now. Two files conflicted from the squash landings, `_llm_scanner/_llm_scanner.py` and `_transcript/handle.py`, and I resolved both by keeping this branch's side. Neither 604 nor 606 added a content commit after the versions this branch carries, only main merges, so nothing of theirs is dropped.

Checked two ways: both resolved files are byte-identical to the pre-merge head `2af478d3`, and `git diff --stat afdbcd1b f37afd11` is back to this PR's own slice at 14 files, +2206/-60.

One other change rides that commit. `scanner_supports_streaming`'s docstring still said the fallback reads an attribute on the scan function; `fb5341ee` moved that to the identity-keyed registry. Corrected to name the registry. Two lines, no behaviour change.

epatey (approved):
Approving at f37afd115e472085deb36cddfee266b5b2b5f310.

Since this PR was opened its three predecessors landed on main, 603, 604 and 606. I merged main up on this branch and resolved the two conflicts the squash landings created, in `_llm_scanner/_llm_scanner.py` and `_transcript/handle.py`, by keeping the branch side. Neither 604 nor 606 added content after the versions this branch carried, and the diff against main is back to this PR's own 14 files, +2206/-60. I also corrected two lines of `scanner_supports_streaming`'s docstring, which still described its fallback as an attribute on the scan function after `fb5341ee` moved it to the identity-keyed registry. Every check is green on this head.

Landing it by squash merge.

PR #608, opened 2026-09-02
feat: scans stream large transcripts, and scanners can opt into streaming

rasmusfaber:
## Summary

This is where the memory bound actually lands for `scan()`. Scanners that are permitted to stream read through one shared, refcounted transcript handle rather than a materialized `Transcript`. To be precise about what that saves: `main` already reads one union `Transcript` per transcript in flight and shares it across that transcript's scanners, so nothing was being copied per scanner. What the streaming path removes is the object graph itself. For an eligible sample the messages and events are replayed from a disk spool one at a time and the full lists are never built, on the scan side or in the recorder. Permission is per scanner: `llm_scanner` grants it to itself unless `question`/`template_variables` is a callable or a timeline is requested, and any other scanner declares it with the new `@scanner(supports_streaming=True)` (a scan streams only when every scanner sharing the transcript is permitted; `TranscriptHandle` is now exported and `docs/custom_scanner.qmd` gains a Streaming section). Streaming also requires an `.eval`-format source and a sample above 64 MiB; everything else takes `main`'s materialized path unchanged, and the recorder copies the compact pool-encoded form straight out of the spool rather than expanding it into a `Transcript` and re-condensing it.

It also fixes a failure mode users can hit today: a single event that fails validation made `scan_results_df()` raise for the **whole file**, while the scan itself reported `complete` — a successful scan whose results would not open.

## Review scope

Last of six PRs carved out of #491, which was closed as too large to review. It sits on #607 and merges #605, so **until those merge GitHub shows the whole series here**. This PR's own contribution is the seam in `_scan.py`, `_transcript/json/passthrough.py`, `_scanresults.py`, `_scanner/result.py` (`ResultReport.input: ScannerInput -> ReportInput`, `SerializedTranscript`, dual value+error rows), `_scanner/scanner.py` and `_scanner/_loaders.py` (`supports_streaming`), `_transcript/eval_log.py` (`EvalLogTranscriptsReader.open`), `__init__.py` (`TranscriptHandle` export), `docs/custom_scanner.qmd`, the `_concurrency/` teardown, and their tests.

Own work: 20 files changed, 2217 insertions(+), 113 deletions(-) (21 commits). https://github.com/METR/inspect_scout/compare/split/streaming-timeline-and-viewer-search...split/scan-seam-and-spool-record also shows #605 (merged here).

## Limitation of the unexpandable-row guard

The guard keeps the frame readable, but `input_data` is dropped for every row in that file, so a row left condensed keeps its pool references with nothing to resolve them against — and nothing in the frame marks which rows those are. Only a log line records that anything was degraded. The honest fix is a per-row marker column, which changes the public results schema and is therefore a maintainer's call rather than something to slip into a bug-fix slice.

Strictly better than what it replaces: one bad event previously made the entire file unopenable.

## Notes for reviewers

**One conditional tolerance, deliberately:** `tests/scan/test_scan_streaming.py::test_recorded_input_resolves_attachments_like_materialized` currently tolerates exactly one key — `score_explanation` — because `main` added `TranscriptInfo.score_explanation` and the materialized read path (`load_filtered.py`) drops it while the spooled record path carries it. That fix is being made against `main` in its own PR rather than here, so two PRs do not edit the same lines. The tolerance is conditional on the defect (it switches itself off when the fix lands) and every other assertion in the test, including any other divergence, still fails for real; an `xfail` mark was rejected because `raises=AssertionError` would have covered the whole test body.

A transcript that cannot be read back for its record (spool gone, `load()` raising) is recorded as an Error row with the scan's own result kept; it was previously logged and recorded as a clean result over an info-only placeholder. A row carrying both a value and an error is a new shape: `resume()` treats it as unrecorded and redoes the scan, the summary counts it under both results and errors, and `status.complete` is False for the run — intended, all flagged for follow-up.

On a legacy log (inspect_ai < 0.3.73 for `ChatMessage.id`, < 0.3.113 for `Event.uuid`) whose samples exceed 64 MiB, the spooled record path copies bytes without pydantic validation, so `id`/`uuid` are null in the recorded `input` while the scanner saw generated ids; `message_references` on that row cannot be matched against the recorded input. Materialized rows are unaffected.

`@scanner(supports_streaming=True)` is verified, not trusted: the decorator resolves the scan function's own annotations (against its own module, so `from __future__ import annotations` works cross-module) and raises `TypeError` unless the transcript parameter accepts `TranscriptHandle`, or when the function carries `__wrapped__`, because `functools.wraps` copies the wrapped function's annotations onto a wrapper that may only accept `Transcript`. Declaring `False` always wins; leaving it undeclared defers to `llm_scanner`'s own per-instance verdict.

The eligibility rule is the subtlest thing in the series. A shared handle bypasses the per-scanner loader's `filter_transcript`, so `_streaming_eligible` must reject any job whose scanner has a narrower filter than the union — otherwise that scanner sees a sibling's broadened content. Forcing the guard true demonstrates the leak.

`_scanresults.py`'s attachment resolution is a provable no-op on `main` today: `condense_events` returns an `EventsData` with exactly `messages` and `calls`, so `input_data` never carries an `attachments` key.

### Agent review

- Author: Claude Opus 5, reviewed across several fresh-context passes — a five-dimension analysis, an independent `codex` pass, and scoped re-reviews per fix round. The final pass ran on a more capable model than the implementation, given this slice writes to a public results column.
- Review found and fixed: the `on_complete` refcount had no owner of last resort, so a follower stranded by cancellation left a handle unclosed; and the metadata merge had no test at all — inverting its precedence wrote stale index metadata into the public `input` column and passed 1,411 tests.
- The teardown drain runs under a shielded cancel scope. That was stress-tested rather than reasoned about: cancelling the pool while the lead held the handle lock mid-parse, to confirm the shield cannot convert a leaked descriptor into a hung shutdown.
- An 89-agent audit (15 candidate changes × 3 adversarial refuters, plus a completeness critic whose 7 findings got the same treatment) preceded this revision, followed by a subagent-driven fix pass with a fresh implementer and a fresh reviewer per change. It found the record-fallback error swallowing and the materialized-handle streaming cost, and rejected a first version of the public opt-in; each fix has a test that fails without it, and the reviewers caught three defects in the plan itself along the way.
- The final whole-branch review of that pass (fresh context, Claude Opus 5) returned four Important findings, all fixed and confirmed by one scoped re-review: the spooled fallback's timelines leak (fixed in #606), an `xfail` mark whose `raises=AssertionError` had covered the whole record test (replaced by the one-key tolerance above), and two holes in the `supports_streaming=True` validator — it resolved annotations against the factory's globals instead of the scan function's, and `functools.wraps` could smuggle a wrapped function's annotations past it. A scoped re-review (fresh context, Claude Sonnet 5) confirmed all four fixed and caught one wrong remedy in a new error message, corrected before push.


epatey (approved):
Approving at 523e5159.

Two things I did on the branch rather than asking you to. I merged main up twice and resolved the squash conflict that 603, 604, 605, 606 and 607 left behind, keeping the branch side in all six hunks; the four resolved files are byte-identical to the pre-merge head, and the diff against main is exactly this PR's own slice. I also corrected the opening paragraph of the description, with no code change: it said scanners had each been materializing their own copy, and main does not do that. Main already reads one union `Transcript` per transcript in flight and shares it across that transcript's scanners, with the per-scanner loader filtering shallowly. What the streaming path actually removes is the object graph itself, and the description now says so. Worth getting right because the description becomes the commit message.

On the question you left for a maintainer: the unexpandable-row guard ships as is. A per-row marker for degraded rows would change the public results schema, and nobody has hit the failure it would flag. Keeping a readable frame with a log line is the right trade for now.

The mechanism reads well. Eligible jobs share one refcounted handle, a large `.eval` sample replays from a disk spool instead of being built into a `Transcript`, and the recorder emits the column bytes from that same spool. Everything that does not qualify takes the existing path untouched, and the eligibility rule is conservative in the right direction.

PR #617, opened 2026-09-06
fix: allow scanners with postponed annotations

### LINKED ISSUE #615: `from __future__ import annotations` breaks implicit loader selection
Claude's favourite unnecessary import causes an error.

Under PEP 563 the annotation is the *string* `"Transcript"`, so every scanner in a module with the future import fails to register (inspect-scout 0.4.44, Python 3.12):

```python
from __future__ import annotations  # remove this line and it works

from inspect_scout import Result, Scanner, Transcript, scanner

@scanner(messages="all")
def my_scanner() -> Scanner[Transcript]:
    async def scan(transcript: Transcript) -> Result:
        return Result(value=len(transcript.messages))
    return scan

my_scanner()
# RuntimeError: Type annotation must conform to ChatMessage or Event, got Transcript
```

`list[ChatMessageAssistant]` fails the same way.

## Cause
The problem is these lines:
https://github.com/meridianlabs-ai/inspect_scout/blob/90a0060dcce3a852d85960ead638f96ab4fd2cf0/src/inspect_scout/_scanner/_loaders.py#L116-L118

Adding the eval_str argument fixes `inspect.signature(scanner_fn, eval_str=True)`.

YusefSyed:
Fixes #615.

Scanners with postponed input annotations can fail during implicit loader selection when an unrelated return annotation uses a factory-local alias. Resolve only the first input annotation needed for the loader. Preserve the defining namespace for functions, partials, cross-module wrappers and async callable instances; callable instances use their unwrapped `__call__` method's namespace.

The branch is integrated with upstream `d2e2e3b`, preserving the `Transcript | TranscriptHandle` behavior from #608. Public registration/execution tests cover postponed Transcript/message-list inputs, local return aliases, partial functions, cross-module wrappers and the callable-instance case requested in the latest review. The new callable regression fails with the reported `AttributeError` on `0e9eaef` and passes with the fix, while its local return annotation remains unevaluated. No frontend gitlink change or new suppressions.

Validation on `f3edbdb`:

- 83 focused loader/scanner tests passed; the 12 annotation tests also passed after formatting the test declaration.
- Scoped mypy, Ruff check/format, suppression ledger and diff checks passed.
- Full existing-environment suite: 3,547 passed, 107 skipped, 2 xfailed, 5 failed. The five failures are the same three installed `inspect_ai` Anthropic `cache_ttl` incompatibilities and two Rich-rendering failures previously reproduced on clean upstream with explicit baseline source imports.
- Repository-wide `make check` retains the documented unrelated environment/type-checking limitations; focused changed-file type checking passes.
- Hosted Build run 35801633281 passed on this callable-instance follow-up, including Python 3.10/3.14 tests, type checks, lint, package and frontend checks.

### Agent review

OpenAI Codex assisted with implementation, testing and this description. A routine worker prepared the earlier rebase; the parent task reviewed it in its existing context and caught an initially invalid baseline import path, then reran the baseline with explicit source imports. After the maintainer identified the remaining callable-instance case, the parent reproduced it, implemented the method-namespace fix, added a public regression and reviewed the scoped diff in the same context. No fresh independent agent review was run for this follow-up. Earlier review had identified and fixed the partial/wrapped namespace issue. Explicitly double-quoted postponed input references remain outside the original fix, with unchanged baseline behavior.

### DIFF
--- a/src/inspect_scout/_scanner/_loaders.py
+++ b/src/inspect_scout/_scanner/_loaders.py
@@ -1,4 +1,5 @@
 import inspect
+from functools import partial
 from typing import (
     Any,
     AsyncIterator,
@@ -112,10 +113,16 @@ def create_implicit_loader(
     Returns:
         Appropriate loader for the scanner's input type.
     """
-    # Get the first parameter's annotation
+    # Resolve only the annotation that selects the loader. Resolving the full
+    # signature can fail on an unrelated return annotation.
     input_annotation = next(
-        iter(inspect.signature(scanner_fn, eval_str=True).parameters.values())
+        iter(inspect.signature(scanner_fn).parameters.values())
     ).annotation
+    if isinstance(input_annotation, str):
+        annotation_source = inspect.unwrap(scanner_fn)
+        while isinstance(annotation_source, partial):
+            annotation_source = inspect.unwrap(annotation_source.func)
+        input_annotation = eval(input_annotation, annotation_source.__globals__)
     if input_annotation is inspect.Parameter.empty or input_annotation == Transcript:
         return _IdentityLoader(content)
 
--- a/src/inspect_scout/_scanner/_loaders.py
+++ b/src/inspect_scout/_scanner/_loaders.py
@@ -114,7 +114,7 @@ def create_implicit_loader(
     """
     # Get the first parameter's annotation
     input_annotation = next(
-        iter(inspect.signature(scanner_fn).parameters.values())
+        iter(inspect.signature(scanner_fn, eval_str=True).parameters.values())
     ).annotation
     if input_annotation is inspect.Parameter.empty or input_annotation == Transcript:
         return _IdentityLoader(content)
--- a/tests/scanner/test_scanner_type_annotations.py
+++ b/tests/scanner/test_scanner_type_annotations.py
@@ -5,9 +5,49 @@ from __future__ import annotations
 import inspect
 from typing import Optional, get_type_hints
 
+import pytest
+from inspect_ai.model import ChatMessageAssistant
 from inspect_ai.model._chat_message import ChatMessage
+from inspect_scout import Transcript
 from inspect_scout._scanner.result import Result
 from inspect_scout._scanner.scanner import Scanner, scanner
+
+
+@pytest.mark.asyncio
+async def test_scanner_instantiates_with_postponed_transcript_annotation() -> None:
+    @scanner(messages="all")
+    def transcript_scanner() -> Scanner[Transcript]:
+        async def scan(transcript: Transcript) -> Result:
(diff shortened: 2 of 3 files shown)


dragonstyle (changes requested):
Verified the postponed-input fix and reproduced a registration regression.

Before merge:

- [ ] Resolve only the input annotation needed for loader selection. `inspect.signature(scanner_fn, eval_str=True)` also evaluates the return annotation. A scanner factory that imports `Result as LocalResult` locally and returns a scanner annotated `scan(transcript: Transcript) -> "LocalResult"` registers before this change but now raises `NameError: name LocalResult is not defined`. Preserve registration when an unrelated return annotation cannot be resolved, and add a regression test for this case.

YusefSyed:
Codex-assisted follow-up: addressed in 5b554fb.

Only the first input annotation used for loader selection is evaluated now; the return annotation is left untouched. The new regression registers and executes a scanner with a factory-local `LocalResult` return alias. Partial scanners and cross-module wrapped scanners also have regression coverage, preserving their original annotation namespace.

The 39 focused scanner/loader tests, `make check`, and suppression checks pass. The full local suite has 3,325 passes and 107 skips; its two Rich display failures reproduce on the unchanged previous head in the same environment. No new suppressions or numerical/reference changes were introduced.

epatey:
Heads up on where main went while this sat.

Rasmus's stack landed today (603, 604, 606, 607, 605, 608 and 634). One of those, 608, changed the same annotation handling in `_loaders.py` that this PR fixes, so this branch now conflicts with main. We did not spot the overlap with your PR before merging, which is our miss.

We still want this one. We ran your tests against what 608 landed and 2 of the 11 fail, so the fix now on main is the narrower of the two and yours covers cases it does not.

The requested changes from September 9 still stand, and this does not change them. Flagging the conflict so it is not a surprise when you pick this back up: it is in `_loaders.py`.

cc @rasmusfaber

dragonstyle (changes requested):
Verified the current post-integration head. Resolving only the input annotation fixes the unrelated return-alias case, but an async callable-instance scanner with a postponed `Transcript` input still fails registration. Before merge:

- [ ] Resolve that input annotation using the callable method's namespace. `annotation_source.__globals__` in `_loaders.py` raises `AttributeError` on the scanner instance. Add a public registration and execution regression test for this supported callable shape, while leaving unrelated return annotations unevaluated.

YusefSyed:
Hi @dragonstyle, your callable-instance case is addressed in f3edbdbf. The input annotation now resolves through the callable method's namespace, while unrelated return annotations remain unevaluated. The regression registers and executes that scanner shape, and the current hosted checks pass. Could you take another look when you have time?

Codex-assisted follow-up, posted with my approval.

PR #620, opened 2026-09-06
fix: count scanner results with no value instead of silently dropping them

### LINKED ISSUE #618: MetricsAccumulator drops `None` results with no count, so scanner metrics are computed over an unknown subset
## Summary

A scanner that returns `Result(value=None)` is excluded from that scanner's metrics with no record that it was excluded. The mean is computed over whichever transcripts happened to produce a value, and nothing in the metrics, the `ScannerSummary`, or the results data frame says how many were left out or why.

This matters most for LLM scanners that catch a judge refusal and return `None` so the scan completes. Refusals are not missing at random: a safety-tuned judge declines the most disturbing transcripts, so the samples that vanish from the denominator are systematically the ones that would have raised the mean.

## Reproduction

inspect_scout 0.5.0 from PyPI. `MetricsAccumulator` is a plain class, so this runs without a model or a scan:

```python
from inspect_ai.scorer import mean, stderr
from inspect_scout._scanner.metrics import MetricsAccumulator

def run(values):
    acc = MetricsAccumulator("concern", [mean(), stderr()])
    for v in values:
        acc.add_result(v)
    return len(acc._scores), acc.compute_metrics()

print(run([9, None, 1]))        # (2, {'concern': {'mean': 5.0, 'stderr': 4.0}})
print(run([1, 1, 1, None]))     # (3, {'concern': {'mean': 1.0, 'stderr': 0.0}})
print(run([None, None]))        # (0, {'concern': {}})
```

The second case is the important one. If the refused transcript was the one that would have scored 10, the batch reports a mean of 1.0, the most benign possible reading, because the dangerous sample was the one the judge declined.

The third case reports `{}`, which is indistinguishable from a scanner that has no metrics configured.

## Mechanism

Three layers, each of which independently removes the information:

1. `src/inspect_scout/_scanner/metrics.py:24-28`, `MetricsAccumulator.add_result`:
   ```python
   if value is not None:
       self._scores.append(SampleScore(score=Score(value=_numeric_score_value(value))))
   ```
   A `None` never becomes a `SampleScore`, so it is not in the denominator and not counted anywhere.

2. `metrics.py:37-47`, `compute_metrics`, strips any NaN metric value before returning, with the comment "compute_eval_scores produces NaN when there are zero successful results (e.g. all errored). NaN serialises to JSON null which fails Summary Pydantic validation." So even the all-unscored sentinel is removed rather than surfaced.

3. `compute_metrics` reads only `score.metrics` from each `EvalScore` that `inspect_ai.compute_eval_scores` returns. But inspect_ai already does the accounting this issue is asking for: `inspect_ai/_eval/task/results.py:406-415` treats a NaN-valued `Score` as "not scored", counts it, and returns `scored_samples` and `unscored_samples` on the `EvalScore` (lines 481-482). Scout discards both fields at the boundary.

Nothing else records it either:

- `ScannerSummary` (`_recorder/summary.py`) has `scans`, `results` ("scans which returned truthy results") and `errors`. A `None` result is not truthy and is not an error, and a legitimate `0` or `False` is also not truthy, so `scans - results - errors` cannot recover the count.
- In the results data frame, `ResultReport.to_df_columns` (`_scanner/result.py:143-183`) routes a `Result(value=None)` through the `result` branch: `value=None`, `value_type="null"`, `scan_error_type=None`. It is recorded as a successful scan with a null value.
- `Result.value` is typed `JsonValue` with the docstring "Scan value." `docs/_scanner_metrics.md` says metrics "aggregate result values." Neither says that `None` is excluded.

## Which scanners hit this

Not the built-in `llm_scanner`. `generate_retry_refusals` (`_util/refusal.py:31-40`) raises `RefusalError` once `retry_refusals` are exhausted, and nothing under `_llm_scanner/` catches it, so the refusal propagates to the `error` branch, is counted in `ScannerSummary.errors`, and shows up in the frame with `scan_error_type="refusal"`.

The drop is reached by any scanner that catches the refusal and returns `None` so the transcript is preserved. `inspect_petri` does exactly this in both its judge and scanner paths (`_judge/judge.py:82-88`, `_judge/scanner.py:109-115`), and reported the resulting under-count as meridianlabs-ai/inspect_petri#113. The scoping comment on that issue established that the drop is in scout, not petri, and asked whether the count should live upstream. This is that upstream issue.

## Proposed fix

The count already exists in inspect_ai; the change is to stop discarding it.

Option A, confined to `metrics.py`, no schema change:

- In `add_result`, record a `None` as `SampleScore(score=Score(value=float("nan")))` instead of skipping it. inspect_ai then counts it as unscored and excludes it from `mean()` exactly as it does today, so no reported metric value changes.
- In `compute_metrics`, carry `score.scored_samples` and `score.unscored_samples` into the returned dict alongside the metrics, for example as `"scored_samples"` and `"unscored_samples"`. The existing NaN strip on metric values stays, so `Summary` validation is unaffected.

After that, the three cases above report `unscored_samples` of 1, 1 and 2, and the all-refused case reports `{"scored_samples": 0.0, "unscored_samples": 2.0}` instead of `{}`.

Option B: add an `unscored` field to `ScannerSummary`. Cleaner, but it crosses the Pydantic to `openapi.json` to `generated.ts` boundary described in CONTRIBUTING.

Either way, `_scanner_metrics.md` should say what a `None` result means for metrics.

Happy to open the PR for option A with a test covering all three cases, once the direction is confirmed.

Checked against HEAD (pushed 2026-09-04): `metrics.py` there matches the 0.5.0 code above. No open PR touches `metrics.py` and I found no prior issue on this.

crishN144:
## Summary

A scanner that returns `Result(value=None)` (for example after catching a judge refusal, as `inspect_petri` does) was excluded from its metrics with no record that it existed: `MetricsAccumulator` skipped it, `ScannerSummary` counted it as neither a result nor an error, and an all-`None` scanner reported `{}`. Because judge refusals are not missing at random, a batch could read safer than the behaviour actually submitted (meridianlabs-ai/inspect_petri#113).

This adds `unscored` to `ScannerSummary`, counted in `Summary._report` beside `scans` / `results` / `errors`, so every scan lands in exactly one of results, unscored, or errors. The terminal summary gains an `unscored` column when any scanner has such results. Metric values and the metrics dict are unchanged.

Closes #618

## Reproduction (from #618, inspect_scout 0.5.0)

```python
from inspect_ai.scorer import mean
from inspect_scout._scanner.metrics import MetricsAccumulator

acc = MetricsAccumulator("concern", [mean()])
for v in [1, 1, 1, None]:
    acc.add_result(v)
len(acc._scores), acc.compute_metrics()   # (3, {'concern': {'mean': 1.0}}) — the None left no trace
```

End to end on this branch, 216 transcripts with a scanner returning `None` for half:

```
ScannerSummary: scans=216 results=109 unscored=107 errors=0     (109 + 107 + 0 == 216)
metrics: {'half_none': {'mean': 7.0}}                            (unchanged)
```

## Changes

- `_recorder/summary.py` — `ScannerSummary.unscored`; `_report` counts `result.result is not None and result.result.value is None`. Errored scans have no `result` and stay in `errors`; falsy values (`0`, `False`) are scored, not unscored.
- `_display/rich.py` — `unscored` column between `results` and `errors`, shown only when some scanner has unscored results (same pattern as the validation column).
- `_view/openapi.json` — regenerated; one new integer property.
- `docs/_scanner_metrics.md` — one paragraph on what a `None` result means for metrics.
- Tests: `tests/recorder/test_summary.py` (none / all-none / falsy-is-scored / error-is-not-unscored), `tests/display/test_rich.py` (column present only when needed, ordered between results and errors).

## Expected CI failure: `generated.ts`

`check-schema-and-types` will fail on "generated types are in sync with checked-in schema": `apps/scout/src/types/generated.ts` lives in the `ts-mono` submodule and needs `pnpm --filter scout types:generate` there. Per `AGENTS.md` I have not touched the submodule pointer. Happy to open the ts-mono PR and follow `land-ts-mono`, or leave the regeneration to whoever lands this — whichever you prefer. Everything else (ruff, mypy strict, suppressions ledger, tests) passes locally.

## Why not the option A in #618

I built option A first (counts as `scored_samples` / `unscored_samples` keys inside the per-scorer metrics dict). Review found it regresses the summary display: whenever no metric value survives the NaN filter — an all-`None` scanner, or the first throttled tick of any scan whose first batch errored — the counts become the first keys, and both `_display/rich.py` and the viewer's `ActiveScanView` headline the first key, so the summary showed `scored_samples = 0` where it used to show `-`. It also retained ~1.1 KB per `None` result as a NaN `SampleScore`, changed the dict shape for every scanner, and could collide with a user metric of the same name. The structural field has none of those problems and needs no change to `metrics.py` at all.

Two things deliberately left out, both bigger than this PR: a `None` nested inside a dict-valued result is neither counted nor excluded cleanly today (pre-existing), and `unscored` does not distinguish a judge refusal from an unparseable answer, which is the distinction that would make the count actionable. I can open issues for either.

### Agent review

- Reviewer: Claude Fable 5.1 via `/code-review`, 1 pass, 8 angles (line-by-line, removed behaviour, cross-file consumers, reuse, simplification, efficiency, altitude, CLAUDE.md conventions), each angle in a fresh subagent context. The pass ran against the first implementation (option A); its findings are why this PR is option B.
- Findings: 10 kept after one-vote verification, executed where possible.
  - 7 fixed by switching to option B: counts-first headline regression in the terminal and viewer; ~1.1 KB retained per `None` (measured); counts on the wrong layer; silent overwrite of a user metric named `scored_samples`; unresolved `*` key persisted as a scorer in all-`None` dict mode; docs claiming "M of N" when errored scans never reach the accumulator; counts diverging from `scans` on resume.
  - 1 no change needed: prefer `Score.unscored(reason=...)` over a bare NaN — superseded, option B constructs no sentinel.
  - 2 deferred as out of scope (above): nested `None` inside dict results; refusal vs parse-failure reasons.
- Local checks on the final diff: `ruff check` / `ruff format`, `mypy src examples tests` (strict), `python3 scripts/check_suppressions.py` (ledger unchanged), full suite `pytest tests` (3327 passed, 107 skipped), end-to-end scan above.


crishN144:
Rebased on current main, no conflicts. Local checks pass: ruff, `mypy src examples tests`, suppressions ledger, and the full suite (3346 passed, 107 skipped).

The two red checks here look pre-existing on main rather than from this PR:

- `py-mypy` fails with 23 errors in `_observe/providers/{openai,anthropic,google}.py` (`Missing type arguments for generic type "ObjectProxy"` / `Unused "type: ignore"`). That started with wrapt 2.4.1 (released 2026-09-10), which makes `ObjectProxy` generic; CI resolves wrapt unpinned. I reproduced it locally by installing wrapt 2.4.1 and get the identical 23 errors, none in files this PR touches. The nightly run on main shows the same failure.
- `check-schema-and-types` is also red on that same nightly run on main.

Happy to open a separate PR for the wrapt 2.x typing fix if that would help.

crishN144:
Rebased onto main (`1fda384a`). That clears `py-mypy`: it was the unpinned wrapt 2.4.1 `ObjectProxy` issue, fixed on main by #637, and this branch doesn't touch `_observe/providers`.

`check-schema-and-types` is now down to one line, the new `unscored: number` in the submodule's `generated.ts`. Since `submodule-on-main` needs the gitlink reachable from ts-mono main, I can't land that side myself. Happy to open the ts-mono PR and follow `land-ts-mono` so you just merge the two in order; say the word and I'll push it. Otherwise it's a one-line regeneration whenever you land this.

Issue #495, opened 2026-07-06
Request: Better rate-limit behaviour / support adaptive connections


nix-aisi:
Inspect-scout usage is often very bursty and with very high number of input tokens. This can cause problems with rate limits:

1. saturating usage for other team members
2. causing ~30 min retries on scout samples (with default behavior)

I think the best fix here involves:
1. Support adaptive connections (as inspect-ai does) for `scan()`
2. decouple api-concurrency from worker-pool limit to enable this 

Note that currently in my understanding `adaptive_connections` is unreachable for inspect-scout scan, even if you set it in the model-config since it gets overwritten by `max_connections`.

antnewman:
I looked into this and can confirm the diagnosis, with a couple of findings that I think change the shape of the fix — including one that makes the obvious fix actively harmful. Happy to implement whichever direction you prefer.

## The root cause is confirmed

`_scan.py`:

```python
if scanjob._generate_config.max_connections is None:
    scanjob._generate_config.max_connections = scanjob._max_transcripts
```

and inspect_ai's predicate (`util/_concurrency.py`):

```python
return adaptive_connections is not False and max_connections is None and not batch
```

`max_connections is None` is a hard gate, so any non-`None` value — including one scout computes — turns adaptive off. There is no "the user set it" versus "the tool set it" distinction; the field is the only signal. So `adaptive_connections` is unreachable through any route, exactly as you describe.

## This looks like an accident rather than a sanctioned opt-out

The precedence rule exists to honour a *user's* deliberate pre-adaptive setting, and it is silent by design — so scout's stamp is indistinguishable from someone typing `--max-connections`. Two places in inspect_ai suggest the intended contract runs the other way:

`_eval/evalset.py` declines to inject a `max_connections` for precisely this reason:

> injecting would be a no-op for static behavior but would silently override any task-level `adaptive_connections` (since any non-None `max_connections` disables adaptive per the precedence rule in `Model._connection_concurrency`)

And where inspect_ai embeds scout, `_eval/task/scan.py` deliberately omits the mirror-image knobs:

> `limit` / `shuffle` / `max_processes` / `max_transcripts`: sample-selection and concurrency knobs that belong to the eval

`create_sample_semaphore` is the template for the general problem — bound your own workers in your own limiter, read `max_connections` if you need a number, never write it.

## What's lost is more than throughput

[...]

PR #622, opened 2026-09-07
fix(scan): keep adaptive connections when the scan is certainly single process

antnewman:
Refs #495. This is item **A** from the split proposed on that issue: stop stamping `max_connections` when the scan is certainly single-process, so adaptive connections are reachable.

Opening it speculatively rather than waiting for a ruling, because the analysis was already done and a concrete diff is easier to argue with than a proposal. Happy to reshape or close it.

> **Status (09-09):** @dragonstyle's review is correct — this branch drops the fixed limit in cases where adaptive could never activate, giving pools of 10 and 10,000 where `max_transcripts` was intended. Measurement in the thread also shows a composed guard here fixes only part of it, because this code runs before the model is built. Holding for a direction on shape rather than pushing a partial fix.

## Root cause

`_scan.py` resolved `max_connections` from `max_transcripts` whenever the caller had not set one:

```python
if scanjob._generate_config.max_connections is None:
    scanjob._generate_config.max_connections = scanjob._max_transcripts
```

`adaptive_active()` in inspect_ai is `adaptive_connections is not False and max_connections is None and not batch`, so that assignment silently disabled adaptive connections for **every** scout scan. `git grep adaptive_connections` across scout returns two hits, both in generated schema: nothing in scout reads or forwards it.

## Why the fix is conditional rather than just removing the stamp

Adaptive really is unavailable across processes. `_mp_registry.py` warns and degrades to a fixed limit when a semaphore requests it. So simply deleting the stamp would leave multi-process scans on `AdaptiveConcurrency().start`, which is 20, below the 25 they get today. That is a regression, not a fix.

> **Correction (09-09).** This section originally argued that moving the decision to where `single_process` is computed "does not work either", because `init_scan_model_context()` has already built the `Model` from this config. That was wrong, and I retracted it in the thread: `Model._resolve_config` reads `active_generate_config()` on **every** generate call, so a decision made after the strategy is chosen does take effect. Measurement since suggests the later placement is not merely possible but necessary — a guard at config-resolution time cannot see a config carried on a pre-built `Model` passed as `model=`.

Three of the four conditions in that `single_process` expression are already known during config resolution. `certainly_single_process()` answers with those, and both call sites share it so they cannot drift apart. It is deliberately one-directional:

- `True` means single-process for certain, so adaptive is safe to leave enabled.
- `False` means "possibly multi-process", which keeps the fixed limit.

The condition it cannot see is `total_scans == 1`, so a single-transcript scan keeps today's fixed limit.

**Behaviour is unchanged wherever the scan might be multi-process, including the default.**

## Tests

There was no coverage of this config-resolution block at all, so nothing pinned the old behaviour either way. Added `tests/scan/test_max_connections_resolution.py`:

- unit coverage of the predicate, including the Windows branch via a patched `os.name`;
- an end-to-end scan whose scanner reports what `active_generate_config()` actually shows it, asserting `max_connections` arrives as `None` when single-process is certain;
- an explicit `model_config=GenerateConfig(max_connections=7)` is still honoured;
- the no-regression case, that the fixed limit is still applied when multi-process is possible. That one is skipped on Windows, where `os.name == "nt"` short-circuits the predicate, so it runs on CI rather than here.

Verified the behavioural test genuinely fails without the change, by reverting only the condition and keeping the helper defined.

These tests assert the resolved config rather than the resulting connection pool, which is why they did not catch the defect @dragonstyle found. Pool-level assertions are the right shape and will replace them.

One caveat worth stating: `tests/scan/test_active_scan_cleanup.py::test_interrupted_scan_survives_a_late_metrics_write` failed on one run and passed on a repeat of the identical suite with identical code, so I believe it is flaky rather than affected by this change.

## Not in this PR

- **B**, better rate-limit backoff on the multi-process path, which is where it matters most.
- **C**, adaptive across processes, which would supersede B.
- The adjacent ordering defect at `_scan.py:313-321`, where `generate_config.batch` is read before `model_config` is merged at `:334`, so `--batch` cannot influence the transcript default. Separate defect, separate change.

### Agent review

Authored with Claude Code (Opus 5). Two adversarial passes, each in a fresh context on the same model, instructed to refute rather than confirm.

**This PR was opened without a pass, and that is why the review below was needed rather than avoided.** A fresh-context reviewer told to refute would have checked the guard against all three terms of `adaptive_active()`, which is exactly the defect @dragonstyle found. Recording the gap rather than presenting the passes as prior work.

- **Open** — the guard omits the stamp on process-mode certainty alone, ignoring `adaptive_connections is not False` and `not batch`, giving pools of 10 and 10,000 where 25 was intended. **Found by @dragonstyle**, reproduced and confirmed by pass 1.
- **Open** — a composed guard at this site still leaves `batch` / `adaptive_connections` carried on a `Model` passed as `model=` at 10,000 and 10, because the guard runs at `_scan.py:339` and the model is built at `:349`. Found by pass 2; measured against `main` as the baseline. This is the reason the PR is holding for direction.
- **Fixed** — the description asserted that moving the decision later "does not work" because the `Model` is already built. Retracted in the thread and corrected above.
- **Fixed** — stale `_scan.py:303` reference, now `:313-321`.
- **Confirmed sound** — the premise that the stamp must stay wherever the scan might be multi-process. `_concurrency/_mp_registry.py:36-54` does warn and degrade to a fixed limit.
- **Dismissed** — that the reply should push the partial fix rather than ask, on the grounds that a third request for direction reads as stalling. Reasonable, but it was written before pass 2 established that the fix is partial; landing it silently would leave two routes returning the numbers the review was about.

Test caveat: `test_fixed_limit_is_still_applied_when_multi_process_is_possible` is skipped on Windows, which is where I develop, so the multi-process half is covered by CI rather than locally.

### DIFF
--- a/src/inspect_scout/_scan.py
+++ b/src/inspect_scout/_scan.py
@@ -97,6 +97,21 @@ from ._util.log import init_log
 logger = getLogger(__name__)
 
 
+def certainly_single_process(*, limit: int | None, max_processes: int | None) -> bool:
+    """Will this scan certainly run in a single process?
+
+    The scan strategy is not chosen until the transcripts have been enumerated, but
+    three of its four conditions are known as soon as the options are resolved. Only
+    "there is exactly one scan to run" needs the reader.
+
+    This is deliberately one-directional: True means single-process for certain, while
+    False means "possibly multi-process" rather than "definitely multi-process". That
+    is the safe direction for the caller that uses it to decide whether adaptive
+    connections can be left enabled, since adaptive is not supported across processes.
+    """
+    return limit == 1 or max_processes == 1 or os.name == "nt"
+
+
 def scan(
     scanners: Scanners,
     transcripts: Transcripts | None = None,
@@ -321,7 +336,13 @@ async def scan_async(
         if scanjob._generate_config and model_config
         else model_config or scanjob._generate_config or GenerateConfig()
     )
-    if scanjob._generate_config.max_connections is None:
+    if (
+        scanjob._generate_config.max_connections is None
+        and not certainly_single_process(
+            limit=scanjob._limit,
+            max_processes=scanjob._max_processes,
+        )
+    ):
         scanjob._generate_config.max_connections = scanjob._max_transcripts
 
     # initialize runtime context
@@ -659,11 +680,9 @@ async def _scan_async_inner(
                     "yes",
                 )
                 # are we running single process?
-                single_process = (
-                    total_scans == 1
-                    or scan.spec.options.limit == 1
-                    or scan.spec.options.max_processes == 1
-                    or os.name == "nt"
+                single_process = total_scans == 1 or certainly_single_process(
+                    limit=scan.spec.options.limit,
+                    max_processes=scan.spec.options.max_processes,
                 )
 
                 # set strategy accordingly
--- a/tests/scan/test_max_connections_resolution.py
+++ b/tests/scan/test_max_connections_resolution.py
@@ -1,0 +1,128 @@
+"""How `max_connections` is resolved, and when adaptive connections survive it.
+
+Scout used to stamp `max_connections` unconditionally during config resolution. That
+silently disabled adaptive connections, because `adaptive_active()` in inspect_ai
+requires `max_connections` to be None. Adaptive is not supported across processes (the
+cross-process semaphore registry warns and degrades to a fixed limit), so the stamp is
+still correct whenever the scan might be multi-process, and is only skipped when the
+scan is certainly single-process.
+"""
+
+import os
+import tempfile
+from pathlib import Path
+from unittest.mock import patch
+
+import pytest
+from inspect_ai.model import GenerateConfig
+from inspect_ai.model._generate_config import active_generate_config
+from inspect_scout import Result, Scanner, scan, scanner
+from inspect_scout._scan import certainly_single_process
+from inspect_scout._scanresults import scan_results_df
+from inspect_scout._transcript.factory import transcripts_from
+from inspect_scout._transcript.types import Transcript
+
+LOGS_DIR = Path(__file__).parent.parent.parent / "examples" / "scanner" / "logs"
+
+
+@scanner(name="max_connections_probe", messages="all")
+def max_connections_probe() -> Scanner[Transcript]:
+    """Record what the model layer actually sees, from inside the scan."""
+
+    async def scan_transcript(transcript: Transcript) -> Result:
+        observed = active_generate_config().max_connections
+        return Result(value=observed is None, explanation=str(observed))
+
+    return scan_transcript
+
+
+def test_single_process_is_certain_when_only_one_scan_is_permitted() -> None:
+    assert certainly_single_process(limit=1, max_processes=None) is True
+
+
+def test_single_process_is_certain_when_only_one_process_is_permitted() -> None:
+    assert certainly_single_process(limit=None, max_processes=1) is True
+
+
+def test_single_process_is_certain_on_windows() -> None:
+    # The strategy has always forced single process on Windows.
+    with patch("inspect_scout._scan.os.name", "nt"):
+        assert certainly_single_process(limit=None, max_processes=4) is True
+
+
+def test_single_process_is_not_certain_for_an_ordinary_scan() -> None:
+    # False means "possibly multi-process", which is the answer that keeps the
+    # fixed limit in place. It must not be read as "definitely multi-process".
+    with patch("inspect_scout._scan.os.name", "posix"):
+        assert certainly_single_process(limit=None, max_processes=None) is False
+        assert certainly_single_process(limit=10, max_processes=4) is False
+
+
+def test_adaptive_survives_when_the_scan_is_certainly_single_process() -> None:
+    # max_processes=1 is knowable before the transcripts are read, so the stamp is
+    # skipped and max_connections reaches the model layer as None, which is what
+    # adaptive_active() requires.
+    with tempfile.TemporaryDirectory() as tmpdir:
+        status = scan(
+            scanners=[max_connections_probe()],
+            transcripts=transcripts_from(LOGS_DIR),
+            scans=tmpdir,
+            limit=1,
+            max_processes=1,
+            model="mockllm/model",
+        )
+        assert status.complete
+        assert status.location is not None
+
+        results = scan_results_df(status.location, scanner="max_connections_probe")
+        observed = results.scanners["max_connections_probe"]["value"].tolist()
+        assert observed == [True]
+
+
+def test_explicit_max_connections_is_never_overridden() -> None:
+    # The resolution only ever fills in an unset value, so a caller who asks for a
+    # specific cap keeps it whichever process mode the scan turns out to use.
+    with tempfile.TemporaryDirectory() as tmpdir:
+        status = scan(
+            scanners=[max_connections_probe()],
+            transcripts=transcripts_from(LOGS_DIR),
+            scans=tmpdir,
+            limit=1,
+            max_processes=1,
+            model="mockllm/model",
+            model_config=GenerateConfig(max_connections=7),
+        )
+        assert status.complete
+        assert status.location is not None
+
+        results = scan_results_df(status.location, scanner="max_connections_probe")
+        frame = results.scanners["max_connections_probe"]
+        assert frame["value"].tolist() == [False]
+        assert frame["explanation"].tolist() == ["7"]
+
+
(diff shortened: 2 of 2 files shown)


nix-aisi:
I think this PR is an improvement over the status quo but I'd prefer if scout supported `adaptive_connections` in a more clear and discoverable way, e.g. something like:

- Support a documented argument `scan(adaptive_connections=)`. Probably default to off.
- Never set `max_connections` at `_scan.py:336`
- Later in `scan.py` (after we know for sure if we are in a single process or multi process) update the generation config via `set_active_generate_config`:
    - if the user doesn't request adaptive, set max-connections
    - if we are in single process and the user requested adaptive do nothing (it's the default)
    - if we are in a multi-process and the user requested adaptive throw a warning and fall back to non-adaptive.
 
Not 100% this alternative is better but worth getting a maintainer judgment on the right way to proceed.

antnewman:
You're right, and my stated blocker was wrong.

I claimed the decision couldn't move later because `init_scan_model_context()` has already built the `Model` from this config. That's not how it works. `Model._resolve_config` (`_model.py:1907`) reads `active_generate_config()` on **every** generate call and merges it, and `max_connections`/`adaptive_connections` are explicitly among the operational fields merged even for non-active models. So `set_active_generate_config()` after the strategy is chosen does take effect. I should have checked that before asserting it.

I also checked the multi-process half, since your design needs it: `multi_process.py:206` reads `active_generate_config()` inside `the_func`, which runs when the strategy executes rather than at construction. So a late set propagates to the workers as well.

Your approach works end to end, and it's better than mine. It puts the decision where the process mode is actually known, handles all four cases explicitly instead of collapsing them into "certainly single-process or not", and makes the capability discoverable rather than implicit.

One question worth settling before I rewrite it. You suggest `adaptive_connections=` defaults to off. Note that inspect_ai's own default is effectively on: `adaptive_active()` is `adaptive_connections is not False and max_connections is None and not batch`, so with nothing set upstream runs adaptive. Defaulting scout to off is a deliberate divergence from that. It's defensible, since scout's default path is multi-process where adaptive can't work anyway, but it's a policy choice rather than a detail, and it changes what a user who sets nothing gets.

@epatey happy to rewrite this to the shape above, and to take the default either way. Leaving the PR as-is rather than updating it against main until there's a direction, since a rewrite would discard that anyway.

dragonstyle (changes requested):
The single-process path now creates an adaptive pool, and explicit connection limits are preserved. Before merge:

- [ ] Preserve the existing fixed limit when adaptive connections cannot activate, including `adaptive_connections=False` and batch mode. With `max_transcripts=25`, the current change makes those pools 10 and 10,000 respectively instead of 25. Only omit the default when the effective model configuration permits adaptive connections, and add regression tests that inspect the resulting pools for opt-out, batch, and explicit-limit cases.

antnewman:
You're right, and the numbers reproduce exactly. With `max_transcripts=25`, `max_processes=1`:

| resolved config | pool on this branch | intended |
|---|---|---|
| default | 20, adaptive | 20, adaptive |
| `adaptive_connections=False` | **10** | 25 |
| `batch=True` | **10000** | 25 |
| `max_connections=7` | 7 | 7 |

The mistake is that `certainly_single_process()` answers "can this scan run in one process?", when the question that decides the outcome is "will adaptive actually activate?". Single-process is necessary and not sufficient — `adaptive_active()` also requires `adaptive_connections is not False` and `not batch`. I quoted that predicate verbatim in the PR description and then applied one of its three terms.

The obvious fix is to compose it rather than re-derive it:

```python
if scanjob._generate_config.max_connections is None and not (
    adaptive_active(
        scanjob._generate_config.adaptive_connections,
        scanjob._generate_config.max_connections,
        scanjob._generate_config.batch,
    )
    and certainly_single_process(
        limit=scanjob._limit,
        max_processes=scanjob._max_processes,
    )
):
    scanjob._generate_config.max_connections = scanjob._max_transcripts
```

**That only half-works, though, and the half it misses is why I'm asking rather than pushing.**

Your wording is the *effective* model configuration, and this guard cannot see that. It runs at `_scan.py:339`; the model is built at `:349`. So a config carried on a pre-built `Model` passed as `model=` never reaches it. Measured, same conditions:

| route | main | this branch | branch + the fix above |
|---|---|---|---|
| `model_config=GenerateConfig(batch=True)` | 25 | 10000 | **25** |
| `model=get_model(..., config=GenerateConfig(batch=True))` | 25 | 10000 | **10000** |
| `model=get_model(..., config=GenerateConfig(adaptive_connections=False))` | 25 | 10 | **10** |

The same two numbers you named, still regressed against main, still there after the fix. `model_roles` looks structurally similar, though I have not measured it.

I could make the guard peek at `model.config` when a `Model` instance is passed, but that means duplicating `Model._resolve_config`'s merge semantics at the call site and keeping them in step — the same class of mistake as re-deriving `adaptive_active()` by hand, which is what produced this bug.

Which is really an argument for @nix-aisi's shape, and I should be plainer than I was above: I think theirs is the better end state, not a coin flip. Evaluating after the model is resolved is the only placement that gets the effective config for free. One correction to their sketch, though — the three bullets cover the opt-out but not batch, so `adaptive_connections=True` with `batch=True` in a single process would still take the static path and still want the fixed limit.

@epatey @dragonstyle so the question is which you want:

- **A** — the composed guard above, landing now. Fixes the `model_config=` and CLI routes, leaves the `model=` route regressed against main.
- **B** — the decision moved to after model resolution. Fixes all of them, and settles whether `scan(adaptive_connections=)` becomes a documented parameter. Worth noting inspect_ai's own default is effectively on, so defaulting scout to off would be a deliberate divergence rather than a detail.

I would rather do B, and I will do A today if you would sooner have the partial fix landed while B is decided. What I do not want to do is push something that still returns 10000 without saying so.

One test note for whichever lands: the connection-pool registry is process-global and scout only resets it on the multi-process path, so pool assertions need `init_concurrency()` per case — otherwise the second scan silently reuses the first scan's semaphore and the test passes for the wrong reason. That cost me a false green while checking the table above.

Two adjacent things, neither in scope here:

- `_scan.py:313-321` reads `generate_config.batch` before the `model_config` merge at `:334`, so `batch` passed as `model_config=` cannot influence `max_transcripts`.
- Where there is no project- or ScanJob-level `generate_config` to merge with, `scan()` stamps `max_connections` onto the caller's own `GenerateConfig` rather than a copy — the merge path deepcopies, so it is only that branch. A config reused across two `scan()` calls carries the first scan's stamp into the second.

Happy to file both.

antnewman:
@epatey @dragonstyle @nix-aisi a nudge, and a change of position I should state rather than quietly act on.

First, naming: I used **A** and **B** in my last comment for the two shapes of *this* fix, which collides with the A/B/C split on #495 and in this PR's own description, where **B** is rate-limit backoff. Ignore those letters. The two options are "decide at config resolution" and "decide after model resolution".

**Withdrawing an offer.** On 9 September I said I would land the config-resolution fix if you would sooner have the partial fix while the shape was decided. I no longer think that is a good offer, because it is not merely partial. Measured against main, with `max_transcripts=25`:

| route | main | this branch | branch + config-resolution fix |
|---|---|---|---|
| `model_config=GenerateConfig(batch=True)` | 25 | 10000 | **25** |
| `model=get_model(..., config=GenerateConfig(batch=True))` | 25 | 10000 | **10000** |
| `model=get_model(..., config=GenerateConfig(adaptive_connections=False))` | 25 | 10 | **10** |

So it fixes two routes and leaves a third regressed against main. Landing it would trade one regression for a smaller one, not remove one. That is what changed my mind; the review's wording — the *effective* model configuration — is precisely what the config-resolution site cannot see, since it runs before `init_scan_model_context` builds the model.

**What I would rather do** is @nix-aisi's shape: move the decision to after model resolution, keep today's fixed limit on every path where adaptive cannot activate, and assert the resulting connection pool in tests rather than the resolved config. I have checked it is feasible — `Model._resolve_config` reads `active_generate_config()` per generate call, and `multi_process.py` captures the active config inside `the_func`, so a late set reaches spawned workers too.

One thing I would not want to decide unilaterally is their first bullet, a documented `scan(adaptive_connections=)` parameter. That is the discoverability half of their proposal and a public API question, so I would rather you called it — including the default, since inspect_ai's own default is effectively on and defaulting scout to off would be a deliberate divergence.

Also worth flagging: #608 landed this morning and restructured `_scan.py`, so this branch is now behind and I would rebase before touching anything.

No rush on my account, and I am not asking you to take it just because it is open — if you would rather this waited for #575 and #608 to settle, or would rather close it and keep the measurement on #495 for whoever picks it up, either is a fine answer and I will do that instead.

PR #631, opened 2026-09-10
ci: resync the Claude dev-agent stub with agents#93 (@auto cost filter)

ransomr:
Stub sync for meridianlabs-ai/agents#93 (author_association cost filter on the @auto text paths). Refreshes the whole file from agents/examples/claude-stub.yml — it had drifted (missing the bot guard, the auto-label machine-account exception and the optional OPENAI_API_KEY pass-through) and carried no repo-specific deviations.


ransomr:
@review

PR #632, opened 2026-09-10
fix: preserve provider error details in multi-process scans

### LINKED ISSUE #616: API provider errors mangled with multi-process
## What happens

Under `max_processes > 1`, a worker that hits a provider error sends the raw exception object over the multiprocessing queue (`_mp_subprocess.py:145`, and `Exception` is a member of `UpstreamQueueItem` in `_mp_common.py:223`). The parent unpickles it in
`_upstream_collector` (`multi_process.py:235`).

Anthropic/OpenAI SDK errors don't survive that round trip. `APIStatusError.__init__` takes keyword-only `response` and `body`, but the default exception `__reduce__` only carries `args`, so unpickling calls `APIStatusError(message)` and raises:

    TypeError: APIStatusError.__init__() missing 2 required keyword-only arguments: 'response' and 'body'

The unpickle error is raised from inside `queue.get()`, so it propagates as the scan-killing exception. The real provider error (status code, message, request id) is gone — you can't tell whether you hit a 429, a 529 overload, or a genuine bad request.

It then happens a second time in the shutdown drain (`_mp_shutdown.py:62` → `get_nowait()`), so the traceback the user sees is the shutdown-path `TypeError`, not even the collector one.

### Minimal repro

```python
import pickle, httpx, anthropic
err = anthropic.APIStatusError(
    "overloaded",
    response=httpx.Response(529, request=httpx.Request("POST", "https://api.anthropic.com/v1/messages")),
    body=None,
)
pickle.loads(pickle.dumps(err))
# TypeError: APIStatusError.__init__() missing 2 required keyword-only arguments: 'response' and 'body'
```

Note httpx.Response isn't picklable either, so a __reduce__ fix upstream in the SDK wouldn't round-trip the response anyway.

## Suggested fix

Claude suggests (unverified):
> Don't put exception instances on the upstream queue. Send a picklable payload instead — e.g. WorkerError(type_name: str, message: str, traceback: str) — and raise a generic error in the parent carrying that text. Optionally also wrap the get/get_nowait unpickle calls in the collector and the shutdown drain so a failure to deserialize one item can't mask the shutdown path.

## Versions
inspect_scout 0.4.46, inspect_ai 0.3.262, anthropic 1.3.0 (also reproduces on 0.118.0), Python 3.12, Linux.

verlyn13:
Fixes #616.

Fatal exceptions from spawned scan workers can lose their useful diagnostic
when the parent tries to reconstruct the original exception class. Workers
now send a Scout-owned `WorkerError` payload instead, and the collector
raises a parent-owned `WorkerProcessError`.

The payload carries the worker ID, qualified exception type, message,
formatted remote traceback, and available HTTP status and request ID.
Exception and response objects are not transported, and traceback locals
are not captured. Diagnostic text remains unredacted and unbounded.

## Behavior and scope

The collector-level exception identity is now `WorkerProcessError`, not
the original provider exception class. This applies to worker job errors
made fatal by `fail_on_error=True` and to worker `PrerequisiteError`
failures with or without `fail_on_error`. Public `scan()` reports the
failure and returns incomplete status, subject to the sync diagnostic
limitation below. Ordinary job-error recording and the single-process
strategy are unchanged.

The Python 3.10 backport formatter is retained so nested exception-group
contents remain visible rather than showing only the group heading.

The narrowed diff leaves shutdown draining, cleanup-error precedence,
closed-queue tolerance, and the shared exception-selection path unchanged.
The malformed class-metadata guard and direct `TracebackException`
formatting change are excluded, along with their associated coverage.
No dependencies or suppressions are added.

## Known synchronous diagnostic limitation

I filed the standalone reproduction as
[Inspect AI #5399](https://github.com/UKGovernmentBEIS/inspect_ai/issues/5399).
Synchronous scans can still display Inspect AI's incidental
`RuntimeError: no running event loop` instead of the worker diagnostic.
This PR addresses exception transport, not that separate selection problem.

Only the two dedicated sync Anthropic/OpenAI diagnostic cases use
`xfail(strict=False, raises=ProviderDiagnosticMismatch)`, linked to that
issue. The dedicated exception requires missing provider details and
the known rendered loop-check signature. Other missing-detail failures
remain ordinary assertion failures.

Each marked case first enforces its own returned and persisted incomplete
status, constructor-error exclusion, source binding, spawn/PID evidence,
recorded-worker liveness, and registry/SIGINT-handler restoration.
These checks and the scenario timeouts remain mandatory, as do payload
tests and both async provider-diagnostic cases.

## Validation

Validation ran on the narrowed source subsequently committed as
`4448e212ae7004dc00cfc7b6413b4c2dd70d32dd`. Post-commit comparison confirmed
that the committed source and tests match the validation manifests and
that the resulting PR diff is unchanged. The suites were not rerun after
creating the commit.

Local validation used macOS arm64 and the asyncio backend. Except for
the separate compatibility check, Inspect AI was pinned to
`3dfae0dce8fcf674783f2e3fb2b975eb1bef8c86`.

| Check | Result |
| --- | --- |
| Python 3.12.14 `make check` | Passed |
| Strict mypy, Python 3.10.21 and 3.14.7 | Passed; 472 files each |
| Concurrency suite plus display regression, Python 3.10 | 31 passed, 2 XPASS |
| Same subsystem selection, Python 3.12 | 31 passed, 1 XFAIL, 1 XPASS |
| Same subsystem selection, Python 3.14 | 31 passed, 1 XFAIL, 1 XPASS |
| Full default `tests/` selection, Python 3.12 | 3,357 passed, 107 skipped, 1 XFAIL, 1 XPASS |

All listed invocations returned 0. The subsystem selection was
`tests/concurrency` plus `tests/display/test_exception_traceback.py`.
The full run collected 3,466 tests and used 14 workers with
`--dist worksteal`.

An earlier full run had three fixture filename-limit failures and two
Rich display failures. Isolated controls reproduced those failures with
a long fixture root and `TERM=dumb`, respectively. The original run's
terminal settings were not recorded. The successful full run used a
short fixture root and `TERM=xterm-256color`, with `NO_COLOR=1` retained
from the controls. No source or test changes were made between the runs.

A separate fresh Python 3.12 environment pinned to Inspect AI
`ba57a753b23fa85a0fb4fd1f448337a6a0cf7bea` passed the focused compatibility
selection: 23 passed, 2 XFAIL, exit 0. This was a focused runtime check,
not a full-suite or packaging/build validation.

XFAIL/XPASS counts describe individual executions, not failure rates.
Linux, Windows, and Trio were not validated locally. No hosted CI
result is claimed for the narrowed update.

### Agent review

Codex performed the implementation and narrowing and ran the recorded
validation. Its exact serving model was not independently verified.

ChatGPT reviewed the narrowing plan and expected-failure boundary,
then performed one complete review of the pre-format narrowing diff
in this existing conversation, not a fresh context. Review feedback
restricted expected failures to the known diagnostic signature,
preserved same-run mandatory checks, and placed modifying checks
before final runtime validation. The subsequent Ruff change only
reformatted the two issue-link expressions.

Earlier reviews covered broader candidate revisions. They are not
presented as reviews of the final narrowed tree. No independent
fresh-context complete-diff review of that final tree is recorded.

### DIFF
--- a/design/exception_handling.md
+++ b/design/exception_handling.md
@@ -29,7 +29,7 @@ Two functions serve as the **containment boundary** for job exceptions:
 1. **`_parse_function`**: Executes parse jobs (transcript reading and scanner job creation)
 2. **`_scan_function`**: Executes scan jobs (running scanner functions on inputs)
 
-**Design Principle:** Both functions must catch ALL exceptions from their operations. Exceptions are transformed into `Error` objects within `ResultReport` structures. **No exceptions should escape these functions** - they always return normally, possibly with Error results.
+**Design Principle:** Ordinary job exceptions are transformed into `Error` objects within `ResultReport` structures by default. Explicit `fail_on_error=True` and `PrerequisiteError` bypass this containment and interrupt the scan. Cancellation is also allowed to propagate. These exceptions to containment are deliberate; infrastructure failures must not become successful job results.
 
 This ensures:
 - The scan can continue processing other work items
@@ -66,30 +66,37 @@ The scan recorder persists these Error results, making failures visible in scan 
 
 Infrastructure exceptions must propagate to the top level to terminate the scan with a clear error. However, in a multi-process architecture, this propagation requires careful coordination.
 
-**Design Principle:** Infrastructure exceptions use a **single communication path** from worker to parent - the upstream queue. Workers do NOT use exception bubble-up (re-raising) because the parent doesn't monitor worker process exit codes.
+**Design Principle:** Fatal work-task exceptions use the upstream queue to report to the parent. The worker also re-raises locally; its stderr or process exit is not the parent's diagnostic transport. Failures before the worker reaches this work-task boundary retain their existing behavior.
 
 ## Multi-Process Exception Flow
 
 ### Worker Exception Handling
 
-When a worker encounters an infrastructure exception:
+When a worker's work task encounters an infrastructure exception, a prerequisite failure, or an explicit fail-fast job error:
 
 1. **Catch**: Worker's exception handler catches the exception
-2. **Report**: Worker sends the exception object to parent via `upstream_queue.put(exception)`
-3. **Exit Gracefully**: Worker exits cleanly (possibly sending `WorkerComplete` sentinel first, or simply returning)
-4. **No Re-raise**: Worker does NOT re-raise the exception
+2. **Normalize and report**: Worker sends a `WorkerError` containing worker ID, qualified exception type, message, formatted remote traceback, and optional status/request ID
+3. **Re-raise**: The original exception propagates locally; the shutdown monitor is cancelled in `finally`
+4. **Completion**: `WorkerComplete` is sent only on clean completion, never to reinterpret a fatal error as success
 
-The graceful exit ensures the worker doesn't leave the collector hanging. Whether to send `WorkerComplete` before exiting is a design decision - sending it maintains the normal completion contract, but the exception itself signals abnormal termination.
+The error payload contains only deliberate scalar fields. It excludes response/request objects, bodies, headers, traceback locals, and exception object graphs. Existing exception text may contain sensitive information; this is not a general redaction guarantee. Other upstream message types retain their existing structures.
+
+Chains and exception groups are preserved as formatted traceback text when they reach this boundary. Optional metadata comes only from the caught exception's own plain `int`/`str` attributes; it is not borrowed from an arbitrary cause or group member. Broken optional getters or formatters do not replace the original diagnostic. The existing single-process strategy still controls which exception reaches the worker boundary.
 
 ### Parent Exception Handling
 
 When the parent's collector receives an infrastructure exception from the queue:
 
-1. **Match**: Pattern matching identifies the item as an `Exception` instance
-2. **Propagate**: Exception is re-raised in the collector task
+1. **Match**: Pattern matching identifies `WorkerError`
+2. **Propagate**: The collector raises a parent-owned `WorkerProcessError` with the remote diagnostic, without constructing the original provider exception class
 3. **Cancel**: Task group catches the exception, cancelling the producer task
 4. **Shutdown**: Exception propagates to the strategy's exception handler
 5. **Cleanup**: Shutdown machinery terminates remaining workers, drains queues, cleans up resources
 
 This ensures orderly teardown even when one worker encounters an infrastructure failure.
 
+The parent unwraps only single-member task groups, preserving multi-member groups and avoiding incidental exception context. A failed active collector read remains fatal. The public scan path normally records an incomplete status and displays the fatal error; callers should not depend on catching the original provider class across a process boundary. Single-process handling and default job-error recording remain unchanged.
+
+Python 3.10 uses the existing `exceptiongroup` backport to render task-group failures as text: Rich does not expand those backported groups. This preserves all nested diagnostics at the final display boundary without changing exception propagation. Python 3.11 and later retain Rich's native group rendering.
+
+Both shutdown drains log unexpected read failures through a module logger, stop reading the affected queue, and continue bounded termination and queue closure. A consumed item's decoding failure does not prove that its pipe is healthy. Shutdown returns its first read error after teardown; the parent raises it only after otherwise normal completion. An existing primary error, cancellation, or swallowed `KeyboardInterrupt` takes precedence. Other unexpected cleanup exceptions are also secondary when a primary failure or interrupt exists. This does not add supervision for arbitrary process crashes or a timeout around an OS pipe read that never returns.
--- a/design/exception_handling.md
+++ b/design/exception_handling.md
@@ -29,7 +29,7 @@ Two functions serve as the **containment boundary** for job exceptions:
 1. **`_parse_function`**: Executes parse jobs (transcript reading and scanner job creation)
 2. **`_scan_function`**: Executes scan jobs (running scanner functions on inputs)
 
-**Design Principle:** Ordinary job exceptions are transformed into `Error` objects within `ResultReport` structures by default. Explicit `fail_on_error=True` and `PrerequisiteError` bypass this containment and interrupt the scan. Cancellation is also allowed to propagate. These exceptions to containment are deliberate; infrastructure failures must not become successful job results.
+**Design Principle:** Ordinary job exceptions become `Error` objects within `ResultReport` by default. `fail_on_error=True`, `PrerequisiteError`, and cancellation bypass this containment and interrupt the scan.
 
 This ensures:
 - The scan can continue processing other work items
@@ -66,7 +66,7 @@ The scan recorder persists these Error results, making failures visible in scan 
 
 Infrastructure exceptions must propagate to the top level to terminate the scan with a clear error. However, in a multi-process architecture, this propagation requires careful coordination.
 
-**Design Principle:** Fatal work-task exceptions use the upstream queue to report to the parent. The worker also re-raises locally; its stderr or process exit is not the parent's diagnostic transport. Failures before the worker reaches this work-task boundary retain their existing behavior.
+**Design Principle:** Fatal work-task exceptions are reported to the parent through the upstream queue. Workers also re-raise locally; the parent does not use that re-raise to receive the diagnostic.
 
 ## Multi-Process Exception Flow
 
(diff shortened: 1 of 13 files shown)


epatey (changes requested):
Thanks for this, and for the careful write-up. Saying what you verified and what you did not made the intake much easier.

The transport fix is the part we want: worker errors travel as the `WorkerError` dataclass, the collector raises `WorkerProcessError`, and the tests and design-doc edits cover that path. Issue #616 is real and this closes it.

Could you narrow the PR to that? Please hold these for separate work:

- the shutdown-drain rework in both drains (`drain_one` and the new return value of `shutdown_subprocesses`)
- the cleanup-error precedence in `multi_process.py` (the `completed` flag and raising the first read failure after a normal completion)
- dropping the final drain's tolerance of a closed-queue `ValueError`
- replacing `inner_exception` with local unwrapping and `raise ex from None`

This is a scope call, not a finding against the code. Our rule for error-path machinery is that each piece names the failure it handles, and we agree that handling it is required before building it. Once every worker item is one of our own picklable dataclasses, an undecodable item on the upstream queue stops being an expected case, so the drain and cleanup changes need their own stated requirement. The `inner_exception` replacement rests on a claim about the shared helper that we would rather examine at its other three call sites as well. Each of these is welcome as its own issue or PR with the failure mode written down.

One thing to add to the description when you narrow: under `fail_on_error`, a fatal worker error now arrives as `WorkerProcessError` rather than as the original exception type (a plain `ValueError` from scanner code, for example). That is fine, but callers who catch a specific type around `scan()` should be able to see it in the PR body.

CI runs will be approved once the narrowed head is pushed.

verlyn13:
Thanks, that makes sense. I'm narrowing to the transport fix with its tests and
design-doc updates, and leaving the four held items out of this PR.

Testing that narrower scope left two questions and two scope notes.

Reverting the `inner_exception` change exposes an intermittent failure where sync
scans report `RuntimeError: no running event loop` instead of the provider error.
With main's production code unchanged at `9a1f9881`, I ran four spawned sync
trials on each of Python 3.12 and 3.14 using an ordinary picklable `ValueError`.
I observed the wrong diagnostic in three trials on 3.12 and one on 3.14. One
async control on each runtime passed (macOS spawn, `inspect_ai` at `3dfae0dc`,
whose runner and helper files match main at `8ebe620d`). That main-branch
confirmation did not cover 3.10.

I also reproduced that selection in a fresh Python 3.12 environment with
`inspect_ai` and its required dependencies, without Scout. The shared helper
selected the incidental event-loop context instead of the intended exception.
The interaction
described in [Inspect AI #2694](https://github.com/UKGovernmentBEIS/inspect_ai/pull/2694)
is present again. I now have a short, executed Scout-free repro through
`run_coroutine` that I can file against `inspect_ai` unless you'd prefer it
elsewhere. I'm treating it as a separate upstream issue rather than a reason
to keep the replacement here.

For this PR, would you prefer isolating the provider-detail assertions for the two
sync Anthropic/OpenAI cases in dedicated tests with `xfail(strict=False)` linked
to that issue, or omitting those checks from this PR pending an upstream fix?
The second option defers that diagnostic coverage, not the PR. Payload,
incomplete-status and spawn/PID checks stay unmarked and enforced either way.
So do the checks that the missing-keyword-only-arguments `TypeError` no longer
appears, workers exit, and registry/SIGINT state is restored. The external
timeout and async diagnostic checks also remain enforced.
The injected collector-read-failure test will go out with
the held coverage, not receive an xfail. Nothing is marked yet.

The other is the 3.10 exception-group formatter in `_display/util.py`. In the
macOS Python 3.10 run, with the four held changes removed, removing it too caused
ten retained tests to fail. These included async provider cases whose captured
parent output showed the task-group heading without the worker diagnostic.
Would you prefer to keep that formatter here or handle it separately?

I'm also removing the malformed class-metadata guard and the direct
`TracebackException` formatter. With only those two changes removed from the
current branch, and the four held changes and 3.10 formatter still in place,
the 22 core transport tests passed in one run on each of 3.10, 3.12 and 3.14.
That result does not include the dedicated malformed-metadata or broken-notes
coverage. Those failures remain separate work, including the broken-notes
formatting failure under Logfire.

The current body's claim that prerequisite failures retain their existing
behavior is too broad, and I'll replace that sentence. The body will say that
in multi-process scans, fatal worker errors under `fail_on_error` are reported
as `WorkerProcessError` rather than the original type, and that
`PrerequisiteError` failures are too, with or without `fail_on_error`.
It will also state the unresolved sync-selection limitation described above.

The removal experiments above are not validation of the final narrowed patch.
Once those two questions are settled I'll prepare the narrow diff, rerun the
project checks and tests, and report the results and runtime coverage when I
update the branch.

epatey:
Three answers, and thanks for testing each option before asking.

Keep the two sync provider assertions, marked `xfail(strict=False)`, with the marker linked to the inspect_ai issue. That only works if the issue exists, so please file it first. File it upstream on UKGovernmentBEIS/inspect_ai. The bug is in their code and your repro does not involve Scout, so it belongs there rather than in our fork.

Keep the 3.10 traceback formatter in this PR. Python 3.10 is still supported and ten tests depend on it, so removing it is not part of a narrowing pass. Say in the PR body what failure it prevents. We plan to drop 3.10 in November, and the formatter can go then.

The malformed class-metadata guard stays out, as before.

Your planned description rewrite sounds right. Push the narrowed diff when it is ready and I will approve the CI runs.

verlyn13:
Pushed the narrowed update in `4448e212` and rewrote the description.

The standalone repro is filed as UKGovernmentBEIS/inspect_ai#5399.
The two sync diagnostic cases are marked as requested, the Python 3.10
formatter is retained, and the held changes are out.

Local validation results are in the description. The workflows are
showing `action_required`, as expected.

epatey:
The narrowed head is in and it matches what I asked for on 11 September. I merged main into the branch (`0dc37429`), because our ruleset requires the head to be current with its base. The merge was clean, and the diff against main still holds only this PR's eleven files.

Against my earlier requests:

- The four held items are out. `_mp_shutdown.py` is untouched, the cleanup-error precedence is gone, and `inner_exception` is back at `multi_process.py:370`.
- The description records the behavior change. Under `fail_on_error` a fatal worker error now arrives as `WorkerProcessError`, and so does a worker `PrerequisiteError`.
- The two sync provider cases carry `xfail(strict=False)` pointing at UKGovernmentBEIS/inspect_ai#5399, which is filed and open.
- The Python 3.10 group formatter stays, and the description says what it prevents.
- The malformed class-metadata guard is out.

Our CI has still not run on this branch. Every `pull_request` workflow at this head is `action_required`, so the only check on the commit is the `comment` job from the `pull_request_target` suppressions workflow. None of the thirteen required contexts exist yet.

Next is a full review. Three places for that reviewer to start:

1. Approve the fork runs and read them. No head of this PR has been through our CI. The validation in the description is local, on macOS with asyncio, and it says the suites were not rerun after the narrowing commit.
2. `tests/concurrency/test_mp_worker_error_e2e.py` starts a separate Python process per case, fourteen of them, each with a sixty-second timeout. `py-test` is capped at twenty minutes and the slowest job on main takes about five. Check what this adds on 3.10 and 3.14.
3. `Exception` is no longer a member of `UpstreamQueueItem`, and the collector's match statement has no default case. `_mp_subprocess.py:153` was the only producer that sent one and it now sends `WorkerError`, so nothing is unhandled today. A later item type that misses a case would be dropped in silence instead of raised.

epatey:
Thank you for narrowing this. The head matches what I asked for on 11 September, and the transport fix does what the description says it does. I reviewed it at `0dc37429` and pushed three commits to your branch.

I proved the headline claim rather than taking it on reading. I ran your scenario fixture against main's source and against this branch's, on one machine, by pointing `PYTHONPATH` at a worktree of the merge base:

| source | api | provider details in the parent display |
| --- | --- | --- |
| main | async | all five missing, ending in `TypeError: APIStatusError.__init__() missing 2 required keyword-only arguments` |
| main | sync | all five missing, ending in `RuntimeError: no running event loop` |
| this branch | async | all five present, with `HTTP status: 529` and `Request ID: req_scout_616` |
| this branch | sync | all five missing, ending in `RuntimeError: no running event loop` |

Main's sync run shows the same event-loop masking as this branch's. That is the evidence I wanted that inspect_ai issue 5399 pre-dates this change instead of arriving with it. I also traced the mechanism: `inner_exception` returns element 0 of `_flatten_exception`, and that list's tail order comes from iterating a set of exception objects at `inspect_ai/util/_anyio.py:45-48`, so which exception surfaces is not stable between runs. That matches what you filed.

What I changed, and why. `_has_issue_5399_signature` compacted the display with `"".join(diagnostic.split())`, which drops whitespace but leaves Rich's panel borders in place. A wrapped frame header therefore reads `in││││run_coroutine`, so the literal `"inrun_coroutine"` never matched. The two sync cases raised `AssertionError` instead of `ProviderDiagnosticMismatch`, and `xfail(raises=...)` does not catch that, so they failed rather than expected-failed. Where Rich wraps that header depends on the console width and on how long the absolute checkout path is. On the runner's short path the cases expected-fail, and in a longer worktree they fail: 7 runs out of 8 here. At `COLUMNS=50` the two mandatory async cases failed the same way, on a diagnostic that was complete and correct.

One `_compact()` helper now strips whitespace and box-drawing glyphs from both sides of the provider-detail comparison and the signature check. After it, `COLUMNS=50` gives 12 passed and 2 xfailed, and the signature check still returns false for a lost diagnostic that is not the 5399 rendering.

Everything else is as you wrote it. My middle commit also normalized your `APIStatusError.__init__()` guard and the phrase checks in three other tests, and my third commit takes that back out. Those did not need it: I ran your guard against main's buggy output at every width down to 30 columns and it fires correctly every time, and the other three only break at 30 columns, which is not a width anyone runs. Your assertions there are untouched.

Where this stands. Our CI ran on this PR for the first time today, and it went green at `9c2671ea`. All thirteen required contexts reported, twelve of them green and `no-git-deps` skipped by its own release-branch condition, with `py-test` at about six minutes on 3.10 and on 3.14.

The head is red now, and none of it is yours. inspect_ai `7e88e7fa` landed upstream at 13:17 today and added a required `cache_ttl` argument to `AnthropicAPI.resolve_chat_input`. Our `tests/llm_scanner/test_anthropic_cache_control.py` calls it without that argument, and our build installs inspect_ai from main, so `py-test` and `py-mypy` now fail on main and on every open PR here. `py-test (3.14)` reported 3 failed and 3543 passed, and all three are in that one file. That belongs on main, not on your branch, so there is nothing for you to do about it.

Nothing else needs changing. The PR is waiting on my sign-off and on main going green again.

epatey (approved):
Approving. The head satisfies everything I asked for on 11 September, and the transport fix does what the description says it does. The detail is in my earlier comment today.

Your own coverage went green on our CI at `9c2671ea`. The current head is red for a reason that is not yours. inspect_ai added a required `cache_ttl` argument to `AnthropicAPI.resolve_chat_input` this afternoon, and our `tests/llm_scanner/test_anthropic_cache_control.py` calls it without one. That file is byte-identical to main and it fails on every open PR in this repo right now.

I have armed squash auto-merge. It will land once main is fixed and these checks have been rerun. Nothing further needed from you, and thanks for narrowing it as carefully as you did.

PR #633, opened 2026-09-11
fix: preserve transcript content across JSON field order

rasmusfaber:
## Summary

Recovered or reordered evaluation logs can place `events` before transcript content. A scanner that requests messages but excludes events could then return an empty or incomplete transcript, dropping later messages and sample target, scores, and metadata. When that order makes an early exit unsafe, this change parses to EOF while still omitting excluded events.

When optional top-level fields such as scores are absent, the reader conservatively parses to EOF because it cannot establish that the field will not occur later. Incremental reducers still do not materialize excluded events or their pools.

<details>
<summary>Minimal reproduction</summary>

Run the following against the base revision and then this branch:

```bash
.venv/bin/python - <<'PY'
import asyncio
import io
import json

from inspect_scout import TranscriptInfo
from inspect_scout._transcript.json.load_filtered import load_filtered_transcript


async def main() -> None:
    sample = {
        "events": [],
        "messages": [{"role": "user", "content": "Hello"}],
        "target": "expected",
        "scores": {"accuracy": {"value": "C"}},
        "metadata": {"marker": "present"},
    }
    result = await load_filtered_transcript(
        io.BytesIO(json.dumps(sample).encode()),
        TranscriptInfo(transcript_id="repro"),
        "all",
        None,
    )
    print(json.dumps({"messages": len(result.messages), "metadata": result.metadata}))


asyncio.run(main())
PY
```

The sample orders `events: []` before one user message (`Hello`), target (`expected`), scores (`accuracy: C`), and metadata (`marker: present`), and calls `load_filtered_transcript(..., "all", None)`.

On `origin/main` (b61f5c067), it prints:

```json
{"messages": 0, "metadata": {}}
```

On this branch, it prints one message and metadata containing `sample_metadata: {"marker": "present"}`, `target: "expected"`, and `scores: {"accuracy": {"value": "C"}}`.

</details>

## Manual validation

An affected recovered `.eval` sample contains 31,596 inline messages; its `events` key is at index 4 and `messages` is at index 24. The base reader returned 0 messages. In a fresh process, the corrected public reader returned 31,596 messages and retained 0 events in 46.25 seconds with 490.4 MiB peak RSS. These are observations for this sample, not benchmark guarantees.

### Agent review

- Code author: GPT-5.6 Luna via Codex; subsequent test cleanup by the parent Codex agent.
- Reviewers: GPT-6 Astra via Codex subagents, using a different model from the original implementation author. Four passes ran in fresh contexts (spec compliance, code quality, whole-change review, and cleanup review); two follow-up passes reused the initial reviewer contexts.
- Findings: 1 important finding, fixed. Continuing to parse could hydrate stored timelines against excluded events and raise `ValidationError` for valid messages-only, unscored samples. Timeline dispatch now runs only when events are requested; a deterministic real `InfoEvent`/`Timeline` regression and requested-events control pass. No findings were dismissed. Follow-up and whole-change reviews found no remaining issues; the latter independently confirmed the portable reproduction against base and current source. Cleanup review found no issues, confirmed production logic and established tests were unchanged, and verified that the retained regressions fail against the relevant faulty versions.


epatey:
Pushed 76b8fc15 to your branch, docs only, under the PR's maintainer-edit permission. The load_filtered_transcript docstring now says stored timelines come back only when events are requested, and the timeline test has a comment noting that the missing scores key is a synthetic trigger (inspect_ai always writes scores, as {} when unscored). No code changes.

epatey (approved):
Approving at 76b8fc15.

Verified the fix against inspect_ai's recovery writer (it puts events before messages, metadata and scores, which is what broke the early exit), ran the new tests against main's reader and the repro both ways, and checked who pays the read-to-end cost: only recovery-written or key-missing samples; normally written logs and the parquet path keep the early exit. The last commit is mine, docs only: the docstring now states that stored timelines are returned only when events are requested, and the timeline test notes its missing scores key is a synthetic trigger. Nothing else open. Thanks Rasmus.

PR #634, opened 2026-09-11
feat: scanners can decline sample metadata so metadata-heavy logs stay within memory

rasmusfaber:
## This PR contains:
- [x] New features
- [x] Docs
- [ ] Bug fixes
- [ ] Code refactor

### What is the current behavior?

`TranscriptContent` has `messages`, `events` and `timeline` and no way to decline sample metadata. On the materialized read path the sample's full metadata, target and scores are always parsed from the log body into objects and merged over the index summary, whatever the scanner asked for.

Some historic evals of our misused the metadata field to store large data. One production eval set carries 8.7 GB of sample metadata against 131 MB of messages across 15 transcripts. A scanner that only wants to scan messages nevertheless have to read the full metadata field. (The new streaming path spools metadata and is fine).

### What is the new behavior?

A scanner can declare `@scanner(messages="all", metadata=False)` (or
`TranscriptContent(metadata=False)`, or `@loader(metadata=False)`). Both parsers
then never build the `metadata`, `target` or `scores` sections. With nothing to
overlay, every merge site returns the index's `LazyJSONDict` by identity, so the
scanner sees the log's summary values: metadata thinned to about 1 KB per key,
target as a string, scores as `score_*` values only. Existing scanners are
unchanged — the default reads.

The three keys are declined together on purpose. Skipping only `sample_metadata`
leaves an override that copies the `LazyJSONDict` into a plain dict and hands the
scanner `sample_metadata` as a raw JSON string (verified while designing this).
Declining all three is the only shape that keeps the index dict intact.

Measured on the source case, one
`EvalLogTranscriptsView.read(info, TranscriptContent(messages="all"))`, fresh
process per variant, `ru_maxrss` sampled after the read, spool on disk:

| | peak RSS | `metadata` as JSON | `scores` present | messages |
|---|---|---|---|---|
| default | 3,060 MB | 2,353.6 MB | yes | 9,309 |
| `metadata=False` | 358 MB | 0.0 MB | no | 9,309 |

Across all 15 samples of that 1.6 GB log, `ru_maxrss` summed, minus a
236 MB/process interpreter floor: 13.52 GiB default vs 0.79 GiB declined (17x).

### Does this PR introduce a breaking change?

No. The field defaults to reading, and it is a last-position defaulted dataclass
field, so positional `TranscriptContent(...)` constructions still work.

### Other information:

**Review scope.** Seventh PR on the #491 stack, on top of #608 — until that
merges, GitHub shows the whole series here. This PR's own contribution is
`_transcript/types.py`, `_transcript/util.py`, `_scanner/{scanner,loader}.py`,
the coroutine guards in `_transcript/json/{load_filtered,stream_parse}.py`, the
four reader call sites in `_transcript/eval_log.py` and
`_transcript/database/parquet/transcripts.py`, docs, and tests
(9 commits; `git diff 608ccd716`).

**Deliberately not done here**, one line each:
- ijson still tokenizes every metadata byte; there is no byte-level skip, and the
  early exit fires after `metadata` in field order. No work was done to reduce
  tokenizing, so most of the read cost stays. It is a modest time fix even so,
  because the object construction it skips is not free: 12.0 s -> 9.5 s on the
  sample above and 82 s -> 65 s across all 15 (1.27x), read-only, 3 repeats.
- Parquet DBs: no effect — the `sample_metadata` column *is* the stored value;
  the cost there is at `connect()`, not `read()`. The reader forwards the flag
  for uniformity only.
- `as_scorer` and inspect_ai's in-eval dispatch build metadata themselves and
  never consult `TranscriptContent`; the knob is read-path only.
- `.json`-format logs are materialized by inspect_ai before we see them; the flag
  keeps the objects out of `Transcript.metadata` but saves no parse memory there.
- `llm_scanner` keeps the default because custom templates may read
  `{{ metadata.* }}` and cannot be inspected statically.
- `_streaming_eligible` does not compare `metadata`: a decliner grouped with a
  wanter is over-delivered, not mis-delivered, and comparing would force
  materialization for exactly these workloads. Declining takes effect when every
  scanner in the scan declines.
- The pre-existing `LazyJSONDict.copy()` demotion on *non*-declined reads
  (`eval_metadata` arrives as the string `'{}'`) is untouched.
- The materialized record path (`result.py` `_serialize_input`) serializes any
  lazy key the scanner never accessed as a JSON string — pre-existing for
  `eval_metadata`/`task_args` on every scan; under decline that now includes
  `sample_metadata` in the `input` column. Fixing it means parsing the full
  parquet `sample_metadata` column per result row, so it is left for its own
  change. The streaming passthrough *is* fixed here (it splices the index
  dict's JSON verbatim), so streamed records carry proper objects.

### Agent review
- Author: Claude Fable 5.1 (Claude Code)
  via subagent-driven development: a fresh implementer per task, a fresh
  reviewer per task, one scoped re-review, and a whole-branch review at the end.
- Reviewers, all in fresh contexts that had not seen the authoring session:
  six per-task reviews (Claude Sonnet 5 ×5, Claude Opus 5 ×1) and one scoped
  re-review (Claude Haiku 4.5) — different models from the author; one
  whole-branch review (Claude Fable 5.1, same model family as the author, fresh
  context) that ran its own probes rather than reading the diff; one scoped
  re-review of the fix wave (Claude Sonnet 5).
- Findings: 15 — 5 fixed, 10 dismissed.
  - Fixed (per-task): an import line over 88 chars failed `ruff` I001 (Task 1).
  - Fixed (whole-branch, Important): the streamed record path emitted the
    declined index dict's *unparsed* JSON strings into the `input` column, so
    `sample_metadata` arrived as a string. A spec-level oversight — the design
    doc had reasoned "the passthrough copies nothing → fine". Fixed by splicing
    `LazyJSONDict.to_json_string()` (verbatim, no parse) when nothing was
    overlaid, with a test that fails against the old code.
  - Fixed (whole-branch, Minor ×3): a doc line overclaimed "costs nothing extra
    to scan" (ijson still tokenizes the bytes; now "is not held in memory"); two
    docstrings updated for the declined case and the custom-loader case.
  - Dismissed (10 Minors from per-task reviews, triaged by the whole-branch
    reviewer): union docstring silent on the new semantics (field docstring and
    code comment carry it); no union test for all-`True`/empty (True branch is
    exercised; empty is unreachable); one coroutine variable lacks an explicit
    `| None` annotation its neighbours have (never reassigned; mypy infers it);
    a six-line test spy duplicated rather than a fixture (that test was later
    removed, see below); a spy parameter named `dir` shadows the builtin
    (mirrors the spied signature); `.json`-log and parquet call sites covered
    only by rows on empty-metadata fixtures (parquet is a no-op; `.json` routes
    into the tested parser); `@loader` docstring Args order ≠ signature order
    (cosmetic); the new kwarg sits at different positions in `@scanner` vs
    `@loader` (keyword-only); `@loader`'s error paths untested (pre-existing
    gap); a subagent report miscounted its own edits (not code).
- Test-pruning pass after the reviews, since the diff was 1.65 lines of test
  per line of source: two independent fresh-context analyses (one applying the
  pruning rubric, one instructed to argue for deleting everything and name the
  cheapest surviving test that catches the same routine break), adjudicated by
  the author, then every "another test covers this" claim re-proved by
  mutation on the final tree (10 mutations, 10 caught). Removed: a test that
  asserted the kwarg's default against itself; a stream-parse test whose every
  reachable assertion the passthrough-equivalence and end-to-end tests already
  fail on; a 20-line loader test folded into one assertion on an existing
  loader test; a `CONTENTS` row that ran declined on three fixtures with empty
  metadata (vacuous) and duplicated the dedicated test on the fourth; two union
  rows and two "merged" rows that fail for the same reason as a surviving row;
  two trailing asserts already implied by a dict equality. Ratio now 1.21.


epatey (approved):
Approving at 5d560a1f.

I merged main up on this branch after 603 through 608 landed. Main carries those
six as squash commits while the branch already had their content through the
predecessor branch merges, so every conflicting hunk was that duplication. I
kept the branch side in each one and checked the result three ways: the merged
tree is byte-identical to the pre-merge head, the diff from that head to the
merge commit is empty, and the diff against main is exactly this PR's own 21
files at +337/-57. Full test suite green locally, 3,523 passed and 107 skipped.

On the change itself, three things are fine by me as they stand. The flag is
opt-in and the default read path is untouched. Declining only takes effect when
every scanner in the scan declines, which is the right trade for one shared
read per transcript. And the recorded `input` column writing `sample_metadata`
as a JSON string under decline joins four keys that already serialize that way
on every scan, so fixing it inside this change would undo its purpose.

Squash auto-merge armed. 608 is on main, so this is clear to land.

PR #635, opened 2026-09-11
fix: scorer explanations are no longer dropped when reading or recording transcripts

rasmusfaber:
## Summary

`TranscriptInfo.score_explanation` is copied field by field in four places that omit it, and a stale cache hides it in a fifth, so the grader's explanation is present on the transcript index and silently absent everywhere downstream:

- **Every content-bearing read.** `load_filtered_transcript()`, shared by the eval-log reader, the parquet reader's full read, and parquet ingestion, rebuilds the transcript through a `RawTranscript` mirror without the field.
- **The parquet reader's no-content path.** `read(info, TranscriptContent())` returns `score_explanation=None` for a transcript whose `select()` info has the value.
- **Scan results.** The recorder writes 20 `transcript_*` columns and no `transcript_score_explanation`, so results can never be queried for it. The field's own docstring names that as its purpose: telling a genuinely failed task from a grading misfire.
- **llm_scanner templates.** `{{ score_explanation }}` raises `UndefinedError` per transcript, and `{% if score_explanation %}` silently takes the else branch.
- **Warm index caches.** The per-log index frames are cached in a KVStore invalidated only by `_CACHE_VERSION`, which was not bumped when the column was added, so an already-indexed log keeps reporting `score_explanation=None` from `select()` until its cache entry evicts. The version is bumped so existing caches rebuild.

Each copy site gets the line it is missing and a test that fails on `main` for that site alone; the cache version is bumped (its invalidation mechanism already has a test). With the materialized read fixed, the streaming comparison's `score_explanation` tolerance in `tests/scan/test_scan_streaming.py` is dead code, so it goes too. The three user-facing tables that list every other field (transcript fields, results data frame, llm_scanner template variables) gain a row.

The new results column also exposed a compaction bug: `scanner_table()` took its output schema from pyarrow's dataset inference, which reads the first fragment only. A scan started before a column existed and resumed after it dropped that column from every row whenever a pre-upgrade buffer file happened to be read first. The per-batch cast already fills missing columns with nulls; only discovery was wrong. It now takes the union of field names across all inputs, first-seen type winning, with a deterministic test that orders an old-style buffer file ahead of a compacted file carrying the column.

## Repro

On `main`, insert a transcript with an explanation into a parquet DB and read it back:

```python
db = ParquetTranscriptsDB(location)
await db.connect()
await db.insert([Transcript(transcript_id="t1", source_type="test", source_id="s",
                            source_uri="test://s", score="C",
                            score_explanation="graded C because the tests failed",
                            messages=[ChatMessageUser(content="hi")], events=[])])
await db.disconnect()

read_db = ParquetTranscriptsDB(location, read_only=True)
await read_db.connect()
info = [i async for i in read_db.select()][0]
info.score_explanation                                                     # 'graded C because ...'
(await read_db.read(info, TranscriptContent())).score_explanation          # None
(await read_db.read(info, TranscriptContent(messages="all"))).score_explanation  # None
```

Both reads return the value with this PR. The parquet read tests in `tests/transcript/test_parquet_db.py` and the eval-log read test in `tests/transcript/test_eval_log_read.py` now assert equality on every `TranscriptInfo` field between `select()` and `read()`, so the next field added to the model fails them too.

## Considered and left out

- The seven `Transcript(...)` sites in the source adapters (`_claude_code`, `_langsmith`, `_logfire`, `_phoenix`, `_weave`) also enumerate 20 of 21 fields, but none has an explanation to carry: their `score` is a pass-through of a user-set metadata key, and none reads platform feedback. Adding the field there is a no-op today.
- Converting these mirrors to `Transcript(**info.model_dump(), ...)`, as `transcript_from_eval_sample` and `observe()` already do, would make the class of bug impossible. It is a refactor with its own blast radius (`RawTranscript` renames `transcript_id`; the recorder prefixes and bridges through metadata) and belongs in its own PR.
- The parse-error placeholder in `_scan.py` was a fifth copy site when this PR opened; the streaming series fixed it structurally with a `**info.model_dump()` helper, so that hunk is dropped on rebase. Its test is kept: it passes on `main` today, and nothing else covers that the placeholder carries every info field.
- Follow-ups noticed while auditing, unrelated to this field: `Columns` has no typed `total_tokens` property (works via `__getattr__`); `transcript_info_from_eval_sample` omits `limit` and `total_tokens` although both are derivable from the sample; `observe_update()` cannot set `score_explanation`; the viewer's results column picker hard-codes its column list, so the new column is not selectable in the UI until a ts-mono change lists it.

### Agent review

Prepared by Claude Fable 5.1 (Claude Code). Each claimed site was first reproduced on a clean `origin/main` checkout by fresh-context agents before any code was written.

- Claude reviewers: three fresh-context passes over `git diff origin/main..HEAD`, same model as the author, different lenses (bug-finder on correctness and downstream consumers of the new column, test-reviewer, conventions/docs). Each finding was then judged by two further fresh-context verifiers, one reading, one executing. 6 findings: 2 fixed (the eval-log read test injected a synthetic explanation although the fixture's scorer recorded a real one, so it now compares every info field of `read()` against `select()`; a new one-test file duplicated the existing parquet read tests, so it was folded into `test_parquet_db.py`), 1 subsumed by that fix (file location nit), 3 dismissed (a request for a field-complete template-variable test, broader than this fix; `exclude=` vs `include=` in the placeholder test, equivalent today and both verified to fail on `main`; a PR-description wording suggestion, not a branch change).
- Codex (`gpt-6-astra`, xhigh reasoning, fresh context, different model from the author), two passes: the built-in `codex review --base origin/main` found no actionable regressions; a pass given the intent and out-of-scope list raised 3 findings. 1 fixed: the compaction schema issue described above, which it reproduced with a pre-upgrade buffer file. 2 dismissed: calling `_reports_for_parse_error` directly rather than through a full scan (this file's neighbours already test `_scan` internals the same way, and the end-to-end path would depend on the recorder); the results-table row saying `score` where the column is `transcript_score` (the adjacent `transcript_success` row uses the same wording).
- Codex re-review after the compaction fix, both passes again: no findings. It verified first-seen-type behaviour, null promotion and the `value`/`transcript_score` string casts, the `extra_inputs` dedup path, that discovery adds one footer read per input while preserving batch streaming, and that the new test fails at its final assertion against the pre-fix function and passes on the branch.
- Checks run in a CI-matching venv (Python 3.14, inspect_ai `main`): ruff, mypy, suppressions ledger, and the full default test suite (3528 passed, 0 failed; re-run after rebasing onto the merged streaming series).


dragonstyle (approved):
Verified scorer explanations through transcript reads, recording, templates, and parse-error reports, including legacy Parquet files and mixed-schema compaction. The regression controls fail on the prior behavior and pass with these changes. Approving.

Issue #459, opened 2026-06-09
Running cost estimate for Scout scans


jjallaire:
Scout scans can run a large number of model calls and end up costing a meaningful amount of money. Users have asked for a **running cost estimate** while a scan is in progress (and a final total), so they can keep an eye on spend.

## Cost data source

Inspect AI already has the machinery for this — a model cost/info database plus `compute_model_cost()` and `ModelUsage.total_cost` (UKGovernmentBEIS/inspect_ai#3193, UKGovernmentBEIS/inspect_ai#4015). The catch: **the cost fields in Inspect's model info database are not populated by default**, so out of the box there's no pricing to compute against. Scout needs a pricing source.

In the meantime, community members maintain **URL-accessible model cost databases** we could link to / ingest (referenced in UKGovernmentBEIS/inspect_ai#1152):

- **LiteLLM** `model_prices_and_context_window.json` — https://github.com/BerriAI/litellm/blob/main/model_prices_and_context_window.json (raw: `https://raw.githubusercontent.com/BerriAI/litellm/main/model_prices_and_context_window.json`)
- **models.dev** — https://models.dev/

## Scope / approach

- Surface a running cost estimate in the Scout UI/CLI during a scan, plus a final total.
- Wire a pricing source: either let users point at a pricing config (as Inspect's `--model-pricing-config` does), and/or pull from one of the community URL-accessible DBs above (with caching) so cost works without manual setup.
- Use Inspect's `compute_model_cost()` / `ModelUsage.total_cost` rather than reinventing cost math; aggregate across a scan's model calls.
- Handle models with no pricing entry gracefully (estimate where possible, clearly mark unknowns).
- Decide whether estimates are best-effort/advisory vs. used for any limit enforcement.

## Relevant resources

- UKGovernmentBEIS/inspect_ai#3193 — cost tracking & limits (`--model-pricing-config`, `ModelUsage.total_cost`)
- UKGovernmentBEIS/inspect_ai#4015 — public `compute_model_cost()` + `TimelineSpan.total_cost()`
- UKGovernmentBEIS/inspect_ai#1152 — cost limits discussion + community DB links
- UKGovernmentBEIS/inspect_ai#980 — viewer cost estimates (related)

PR #636, opened 2026-09-12
fix: restore nightly compatibility with Inspect main

brandly:
## Summary

Restore the nightly build against Inspect main while preserving installation with the released Inspect dependency graph.

- Regenerate Scout's OpenAPI schema for Inspect ReviewEvent support.
- Advance the viewer submodule from `4822e2e4` to `57efa420` and rebuild the bundled frontend. This includes the separately merged ReviewEvent implementation (meridianlabs-ai/ts-mono#637), event selection/export (#459), security fixes (#621 and #622), and intervening dependency/tooling updates. ReviewEvent itself reuses approval-event styling; the broader viewer refresh also changes selection and dropdown styles.
- Keep `wrapt>=1.16.0` and introduce a small typed compatibility boundary that inherits the installed, unsubscripted `ObjectProxy` at runtime.
- Preserve event types through standard `Iterable`/`AsyncIterable` contracts and Anthropic message-stream protocols. The contract intentionally covers operations used by Scout; it does not claim arbitrary forwarded attributes or the structured-message result are fully typed.
- Remove obsolete wrapt suppressions and redundant generic bases; read Anthropic snapshot properties once while preserving the missing-attribute fallback.

## Compatibility and regression coverage

Released Inspect 0.3.263 resolves wrapt 1.17.3 through aioboto3/aiobotocore (`wrapt<2`). Inspect main resolves wrapt 2.x. Requiring wrapt 2 would conflict with the released dependency graph.

The new CI matrix tests Python 3.10 and 3.14 with both released Inspect/wrapt 1.17.3 and Inspect main/wrapt 2.x. Each job validates the resolved dependencies, runs the provider/SDK type contracts, and executes the stream tests in separate processes with the C and pure-Python implementations, asserting the selected backend.

The focused suite now has 68 cases (44 existing plus 24 added), including real OpenAI/Anthropic SDK streams over mocked HTTP, sync/async event-type inference, Anthropic manager completion and partial-error handling, connection closure, snapshot absence and single-read behavior, proxy attribute isolation, and function binding/introspection. Tests temporarily remove existing process-wide observe hooks and restore them afterward so suite order cannot introduce nested capture proxies.

Two independently reproduced pre-existing defects remain separate from this compatibility change: raw OpenAI and Anthropic (`stream=True`) context-manager use can bypass capture (and asynchronous entry fails under wrapt 1.x), and Anthropic errors before `message_start` can be masked by the SDK snapshot assertion. The new tests do not claim those paths are fixed.

## Validation

- `make check`: all 469 mypy targets and the lint/format/suppression gates passed; no new suppressions.
- Focused 68-case suite passed locally with wrapt 1.17.3 and 2.4.1, using both C and pure-Python backends. Local environments: Python 3.12/released Inspect and Python 3.14/Inspect main.
- 101 observe tests passed together in one process, including earlier tests that install SDK hooks.
- Full Python suite: 3,363 passed, 107 skipped, with the two previously reproduced macOS Rich console-capture failures deselected. Those two tests fail when included; no change to their code is part of this PR.
- `pnpm check` and `pnpm test` passed on the committed viewer revision.
- CI passed on `1a5e76cd`, including all four new compatibility jobs, the full Python suites on 3.10/3.14, schema/types, viewer tests, and distribution validation: https://github.com/meridianlabs-ai/inspect_scout/actions/runs/34714886609

Fixes the failures observed in nightly run 34682755167.

### Agent review

- Codex (GPT-6) performed one independent review before this follow-up implementation; the authoring conversation was not supplied. It inspected all three original commits, relevant SDK/wrapt source, dependency requirements, and viewer integration, and ran runtime/type probes against both wrapt majors.
- Six findings: three addressed (snapshot double-read, missing permanent compatibility coverage, and disclosure of the broader viewer scope); one accepted limitation (private proxy typing deliberately exposes a narrower contract); two reproduced pre-existing stream defects retained for separate fixes as described above. No merge-blocking runtime regression was found in the original compatibility layer.
- Both of Claude's inline cleanup suggestions were addressed: duplicate iteration protocols were replaced with the standard iterable interfaces, and redundant capture-class `Generic` bases were removed. Runtime generic subscription and SDK type inference are covered by the new tests.
- Follow-up verification included the full checks, SDK-backed regressions, and suite-order isolation checks above; no additional independent-agent review pass is claimed.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -119,6 +119,62 @@ jobs:
         if: ${{ !cancelled() }}
         run: uv cache prune --ci
 
+  wrapt-compatibility:
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.10", "3.14"]
+        dependencies: [released, main]
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+      - uses: astral-sh/setup-uv@v9.0.0
+        with:
+          version: "0.9.4"
+      - name: Install the supported dependency graph
+        env:
+          DEPENDENCIES: ${{ matrix.dependencies }}
+        run: |
+          uv venv --python "$(which python)"
+          if [ "$DEPENDENCIES" = released ]; then
+            # Keep the release pin aligned with Scout's inspect-ai dependency floor.
+            uv pip install '.[dev]' 'inspect-ai==0.3.263' 'wrapt==1.17.3'
+          else
+            uv pip install '.[dev]' 'inspect-ai @ git+https://github.com/UKGovernmentBEIS/inspect_ai.git' 'wrapt>=2,<3'
+          fi
+          uv pip check
+          .venv/bin/python - <<'PY'
+          import os
+          from importlib.metadata import version
+
+          inspect_version = version("inspect-ai")
+          wrapt_version = version("wrapt")
+          print(f"inspect-ai={inspect_version}, wrapt={wrapt_version}")
+          if os.environ["DEPENDENCIES"] == "released":
+              assert inspect_version == "0.3.263"
+              assert wrapt_version == "1.17.3"
+          else:
+              assert wrapt_version.split(".")[0] == "2"
+          PY
+      - name: Check provider and SDK type contracts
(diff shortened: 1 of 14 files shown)


brandly:
@review

dragonstyle:
Superseded by #637 (same schema regeneration plus a minimal wrapt/mypy fix). The tests from this PR are being ported onto main in a follow-up, with @brandly as commit author.

PR #646, opened 2026-09-16
test: pass cache_ttl to resolve_chat_input in the Anthropic cache-control test

epatey:
`py-test` and `py-mypy` are red on main. `AnthropicAPI.resolve_chat_input` in
inspect_ai takes a fourth required argument, `cache_ttl`, added by inspect_ai
pr 5432 (commit `7e88e7fa`, 2026-09-16 13:17Z).
`tests/llm_scanner/test_anthropic_cache_control.py` calls that method at four
sites, and `build.yaml` installs inspect_ai from git main.

The four sites pass `cache_ttl=None`. The provider omits the `ttl` key when it
gets `None`, so each marker stays `{"type": "ephemeral"}` and no assertion in
the file changes. Upstream's own tests for the same method pass `cache_ttl=None`
the same way, at `tests/model/providers/test_anthropic.py:923` and
`test_anthropic_mid_conversation_system.py:28`.

`None` is also what scout's own runs resolve to. Upstream escalates to `"1h"`
only for an active sample, and scout never establishes a sample context, so its
production path takes the same omit-the-`ttl`-key branch the test now pins.

The missing default is not an upstream defect. `resolve_chat_input` lives in
`inspect_ai.model._providers`, and this test reaches into it on purpose to make
an end-to-end claim about request shape. Upstream resolves the TTL in
`generate()` and passes the result down, so a default here would let a caller
skip that resolution silently.

## Verification

Run against inspect_ai main, installed from git the way `build.yaml` does:

- `mypy src examples tests` is clean over 491 files. Before the change it
  reported the 4 errors.
- `pytest` gives 3523 passed, 107 skipped. Before the change, 3 failed.
- `ruff check`, `ruff format --check` and `scripts/check_suppressions.py` are
  clean.

A Build dispatch on main before the fix (run 35104368839) failed `py-test` and
`py-mypy` and nothing else. `py-ruff`, `py-build`, `check-schema-and-types`,
`submodule-on-main`, `viewer-tests` and `js-dist-validation` all passed.

## What stays red on the release PR

#642, the Release Please release PR, validates against the shipped inspect_ai
from PyPI instead of git main. PyPI's latest is 0.3.263, which has no
`cache_ttl` argument, so these four calls fail its `py-mypy`.
`release-pin-deps.yml` names that as the signal to cut an inspect_ai release
first. Its schedule also bumps the `inspect-ai>=` floor in `pyproject.toml` to
the latest shipped version, so the floor needs no hand edit here.

That job is red already, for a separate reason. inspect_ai 0.3.263 depends on
`aioboto3`, which pulls `aiobotocore 2.25.1` and caps `wrapt` below 2, so the
resolver picks `wrapt 1.17.3`. Its stubs predate the generic `ObjectProxy` that
`src/inspect_scout/_observe/providers/provider.py:16` types against, which gives
25 errors under `_observe`. inspect_ai main depends on `aiobotocore 3.9.1` and
allows `wrapt 2.4.1`. It stays as it is: different cause, Release Please owns
that branch, and both parts clear when inspect_ai ships a release carrying
commit `7e88e7fa` and the `aiobotocore` switch.

### Agent review
- Reviewer: Claude Opus 5 as a fresh-context subagent (different context, same
  model as the author), 1 pass.
- Findings: 3. Two are folded into this description, the release-PR lane and the
  note that `None` matches scout's production path. One dismissed: the reviewer
  suggested factoring the four call sites into a local helper, which the
  reviewer itself marked optional, and it is refactoring past what this fix
  needs.
- No second-model pass. `codex login status` reports logged in, so Codex was
  available, but the diff is test-only and the CI run against inspect_ai main is
  the second opinion, per the team's rule for test-only fixes.


epatey:
The release lane came out of this worse than the description says, and whoever next opens inspect_scout pr 642 - chore(main): release 0.5.3 should know that before they read it.

The description says the four calls fail that PR's `py-mypy`, and calls the job red already. `py-test` was green there and is now red too. Under the shipped inspect_ai 0.3.263 there is no `cache_ttl` parameter at all, so the keyword argument raises `TypeError` at runtime, not only under mypy. I ran both revisions of the file against 0.3.263 in a throwaway venv built the way `build.yaml` builds that lane. The base file gives 3 passed. This head gives 3 failed, with `TypeError: AnthropicAPI.resolve_chat_input() got an unexpected keyword argument 'cache_ttl'`. The Build run on #642 before this merge (34963940908) has `py-test (3.10)` and `py-test (3.14)` passing and only `py-mypy (3.10)` failing. The run after it (35107036736) fails all four.

None of that changes the decision. Main was red and is green now. Build 35104368839 on `27716d5d` failed `py-test` and `py-mypy`, and Build 35106977767 on the merge commit `886976e5` passed everything. 0.5.3 was already unreleasable on the 25 errors that `wrapt` 1.17.3 causes across `_observe` and its tests, and the header comment in `release-pin-deps.yml` says failing loudly is what that lane is for when scout needs an unreleased inspect_ai. Both causes clear when inspect_ai ships a release carrying `7e88e7fa`.

I reviewed the change itself and found nothing to fix. `cache_ttl=None` reproduces the request shape the tests were written against. `cache_control_param` in inspect_ai's `anthropic.py:2446` omits the `ttl` key when its argument is `None`, and each assertion compares for exact equality with `{"type": "ephemeral"}`, so a stray `ttl` would fail. I checked the alternative too. `"5m"` produces `{"type": "ephemeral", "ttl": "5m"}` and breaks all three tests.

Issue #650, opened 2026-09-22
feat: option to exclude the role label from rendered messages


marcellomaugeri:
### Issue
`messages_as_str(format="json")` repeats the role inside content.
Example:
```json
{"role": "user", "content": "USER:\nHello\n"}
```

### Proposed approach
Add an
```python
exclude_role_label: bool = False
```
option to `MessageFormatOptions`, which omits the prefix in `message_as_str`. 

**Note:** since `MessageFormatOptions` is part of `openapi.json`, this would also need the `generated.ts` regeneration in `ts-mono`.

PR #651, opened 2026-09-22
feat: option to exclude the role label from rendered messages

marcellomaugeri:
Fix issue #650


marcellomaugeri:
The matching type regeneration is in meridianlabs-ai/ts-mono#684 (`generated.ts` for the new `exclude_role_label` field). Please merge that one first: once it lands, I'll bump this PR's `ts-mono` gitlink to the merged SHA, which also clears `check-schema-and-types`.

